PR #13, opened 2019-03-21
Refactor code base to no longer rely on a global connection

### LINKED ISSUE #11: get rid of get_global_connection
- Link the connection object to the TaskManager,
- Make the TaskManager extensible (inheritance or, preferably, composition) so that someone could implement their own method of connecting to the DB (maybe from Django), with a default implementation that would take no argument and read from env vars, like currently
- Standardize whether functions receive the task manager, the connection or the cursor as parameter. Find the right one and use it consistently where applicable.
- Use consistent connection options and cursor classes.

sdispater:
This PR removes the reliance on a global connection and to make things more consistent most of the time functions and methods expect a connection as an argument.

I also changed the `worker()` function to a `Worker` class to make things easier (especially to gracefully stop workers by adding signal handlers which I added at the same time in this PR to make testing possible). If you see a better way to do it without relying on a class, let me know.

Lastly, but this is beyond the scope of this PR, I wonder if each worker should create its own connection, by getting one from a connection pool for instance, rather than using the one provided by the task manager.

Closes #11


tests/conftest.py
```diff
@@ -52,11 +52,8 @@ def db(setup_db):
     params = setup_db.get_dsn_parameters()
     params["dbname"] = "cabbage_test"
 
-    postgres.reset_global_connection()
-    with closing(postgres.get_global_connection(**params)) as connection:
-        yield connection
-
-    postgres.reset_global_connection()
+    with closing(get_connection(**params)) as conn:
```
ewjoachim: If it's just about calling pyscopg.connect, let's do it directly, no ?

tests/integration/test_task_worker.py
```diff
@@ -0,0 +1,38 @@
...
+    return manager
+
+
+@pytest.fixture()
+def worker(task_manager):
+    return Worker(task_manager, "sum_queue")
+
+
+def test_run(worker, results):
+    worker.run()
```
ewjoachim: Wow, I like the way you're leaving 100% of the test setup to fixtures. Does it scale properly when you have more tests ?

tests/integration/test_task_worker.py
```diff
@@ -0,0 +1,38 @@
...
+def task_manager(connection, results):
+    manager = TaskManager(connection)
+
+    @manager.task(queue="sum_queue")
+    def sum_task(_, a, b):
+        results["sums"].append(a + b)
+
+    @manager.task(queue="sum_queue")
+    def stop_task(_):
+        raise KeyboardInterrupt()
```
ewjoachim: I may be off but I thought unhandled sigint produced KeyboardInterrup, and we're handling sigint, so in real life, no KeyboardInterrupt will be raised, and this code will only be used if a task manually raises keyboard interrupt, no ?

cabbage/task_worker.py
```diff
@@ -11,72 +10,83 @@
...
-
-def one_loop(task_manager: tasks.TaskManager, queue: str, curs: Any, timeout) -> None:
-    process_tasks(task_manager=task_manager, queue=queue, curs=curs)
-    logger.debug("waiting")
-    select.select([curs.connection], [], [], timeout)
+        try:
+            while True:
+                self._install_signal_handlers()
+                self.process_tasks()
+                self._uninstall_signal_handlers()
```
ewjoachim: try finally ?

cabbage/task_worker.py
```diff
@@ -11,72 +10,83 @@
...
+        def stop(_signum, _frame):
+            self._stop_requested = True
+            logger.info("Stop requested, waiting for task to finish")
+
+        signal.signal(signal.SIGINT, stop)
+        signal.signal(signal.SIGTERM, stop)
+
+    def _uninstall_signal_handlers(self) -> None:
+        signal.signal(signal.SIGINT, signal.SIG_DFL)
+        signal.signal(signal.SIGTERM, signal.SIG_DFL)
```
ewjoachim: I'd rather signal specific logic be implemented in a context manager, and we make sure we reinstall previously installed handlers when leaving.

For example, I bet we've destroyed pytests signal handling.

ewjoachim: `signal.getsignal` can give you the info on what's currently there.

ewjoachim:
All tests pass, 100% coverage, linting ok review with @sdispater. Let's do this :)

Issue #3, opened 2019-02-27
Testing tools


ewjoachim:
For Cabbage to be used in a real codebase, we'll need a few tools that will make testability easier:

- A pluggable in-memory list of created tasks
- A way to launch these task synchronously
- A way to launch a single task

I'm thinking something along the lines of:
```python
from cabbage.testing import TestTaskManager
tm = TestTaskManager()

something = []

@tm.task(queue="yay")
def t(task_run, a):
    something.append(a)

t.defer(a=1)

# A pluggable in-memory list of created tasks
assert tm.scheduled == [{"name": "t", "args": {"a": 1}]

# A way to launch these task synchronously
tm.run_scheduled()

assert something == [1]

# A way to launch a single task
t.run(a=2)

assert something == [1, 2]

```

(For now, no idea how we'd substitute the TaskManager with the TestTaskManager in a real case, but we can think of something)

ewjoachim:
> A way to launch a single task

That's being done in #19 :heavy_check_mark:

PR #15, opened 2019-04-05
Refactor the code to make task functions simple functions

sdispater:
This PR makes the `manager.task()` method a true decorator by not altering the decorated function  (except the addition of a `defer()` method to it to replicate the previous behavior).

What this allows is the ability to start decorating existing functions without modifying their signature (no `task_run` anymore) and leave them unit testable without relying on the internals of `cabbage`.

This code came up when trying to address #3 and I think if we go the route of this PR this makes the implementation of test specific tools in `cabbage` mostly unnecessary.

To sum it up, we can now have something like this:

```python

manager = TaskManager()


def add(a, b):
    return a + b


add(1, 2)  # 3
```

### DIFF
--- a/cabbage/task_worker.py
+++ b/cabbage/task_worker.py
@@ -28,7 +28,7 @@ class Worker:
                 break
 
             logger.debug("Waiting")
-            select.select(rlist=[connection], wlist=[], xlist=[], timeout=timeout)
+            select.select([connection], [], [], timeout)
 
     def process_tasks(self) -> None:
         for task_row in self._task_manager.get_tasks(self._queue):  # pragma: no branch
@@ -58,13 +58,12 @@ class Worker:
 
         pk = task_row.id
 
-        task_run = tasks.TaskRun(task=task, id=pk, lock=task_row.targeted_object)
         kwargs = task_row.args
 
         description = f"{task.queue}.{task.name}.{pk}({kwargs})"
         logger.info(f"Start - {description}")
         try:
-            task_run.run(**kwargs)
+            task.func(**kwargs)
         except Exception as e:
             logger.exception(f"Error - {description}")
             raise exceptions.TaskError() from e
--- a/cabbage/tasks.py
+++ b/cabbage/tasks.py
@@ -9,19 +9,17 @@ logger = logging.getLogger(__name__)
 
 class Task:
     def __init__(
-        self, *, manager: "TaskManager", queue: str, name: Optional[str] = None
+        self,
+        func: Callable,
+        *,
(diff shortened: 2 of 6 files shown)


cabbage/task_worker.py
```diff
@@ -28,7 +28,7 @@ def run(self, timeout: int = SOCKET_TIMEOUT) -> None:
                 break
 
             logger.debug("Waiting")
-            select.select(rlist=[connection], wlist=[], xlist=[], timeout=timeout)
+            select.select([connection], [], [], timeout)
```
sdispater: For information, I had to change this because this causes an error (not present in the tests) `TypeError: select() takes no keyword arguments`

ewjoachim: Oo ok, sorry about that :/

cabbage/tasks.py
```diff
@@ -9,19 +9,17 @@
...
-        self.func: Optional[Callable] = None
-        self.name = name
-
-    def __call__(self, func: Callable) -> "Task":
-        self.func = func
-        if not self.name:
-            self.name = func.__name__
-        self.manager.register(self)
-        return self
+        self.func: Optional[Callable] = func
```
ewjoachim: It's not optional anymore, then, is it ?

cabbage/task_worker.py
sdispater: This weren't triggered by the tests so there was no way to know

cabbage/tasks.py
sdispater: That's true. I will update it

cabbage/task_worker.py
ewjoachim: Well, it tells something about our tests, doesn't it ? Can we update the acceptance test to do something ? Either defering a task from another task or spinning a thread that waits for a second and then creates a task should do the trick.

sdispater: I updated the acceptance tests to cover this (by using threads). It fails as expected with the call to `select()` we previously did.

cabbage/tasks.py
sdispater: Updated :-)

ewjoachim:
@mgu @k4nar I'd like a second opinion

ewjoachim:
Ok, let's merge it then, and we'll see later how we want to access current task info.

Maybe we will add a parameter to the task decorator to decide if we want to pass this or not... In a later PR :D

ewjoachim:
Hoping it's ok with everyone, I'll merge now, because I'd like to have a go at some tickets today.

PR #19, opened 2019-04-14
Use Task class instead of modifying function

ewjoachim:
I don't plan to merge this except if we have a consensus.

I admit this is just to make mypy happy.

### DIFF
--- a/tests/unit/test_tasks.py
+++ b/tests/unit/test_tasks.py
@@ -64,7 +64,8 @@ def test_task_manager_task_explicit(manager, mocker):
     assert "foo" == wrapped()
     assert "b" == manager.tasks["b"].name
     assert "a" == manager.tasks["b"].queue
-    assert manager.tasks["b"].func is wrapped
+    assert manager.tasks["b"] is wrapped
+    assert manager.tasks["b"].func is wrapped.__wrapped__
 
 
 def test_task_manager_task_implicit(manager, mocker):
@@ -77,7 +78,8 @@ def test_task_manager_task_implicit(manager, mocker):
     assert "foo" == wrapped()
     assert "wrapped" == manager.tasks["wrapped"].name
     assert "default" == manager.tasks["wrapped"].queue
-    assert manager.tasks["wrapped"].func is wrapped
+    assert manager.tasks["wrapped"] is wrapped
+    assert manager.tasks["wrapped"].func is wrapped.__wrapped__
 
 
 def test_task_manager_register(manager, mocker):
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 __pycache__
 htmlcov
 .tox
+.pytest_cache
--- a/cabbage/task_worker.py
+++ b/cabbage/task_worker.py
@@ -88,7 +88,7 @@ class Worker:
             "Starting task", extra={"action": "task_start", "task": log_context}
         )
         try:
-            task.func(**kwargs)
+            task(**kwargs)
         except Exception as e:
             end_time = log_context["end_time"] = time.time()
             log_context["duration"] = end_time - start_time
--- a/cabbage/tasks.py
+++ b/cabbage/tasks.py
@@ -21,6 +21,9 @@ class Task:
         self.func: Callable = func
         self.name = name or self.func.__name__
 
+    def __call__(self, **kwargs: types.JSONValue) -> None:
+        return self.func(**kwargs)
+
     def defer(self, lock: str = None, **kwargs: types.JSONValue) -> None:
         lock = lock or f"{uuid.uuid4()}"
         task_id = postgres.launch_task(
@@ -63,16 +66,11 @@ class TaskManager:
         Can be used as a decorator or a simple method.
         """
 
-        def _wrap(func: Callable) -> Callable:
+        def _wrap(func: Callable) -> Task:
             task = Task(func, manager=self, queue=queue, name=name)
             self.register(task)
 
-            # Properly typing a callable when adding an attribute to it, especially
-            # a callable attribute, is a hard problem
-            # https://github.com/python/mypy/issues/3882
-            func.defer = task.defer  # type: ignore
-
-            return func
+            return task
 
         if _func is None:
             return _wrap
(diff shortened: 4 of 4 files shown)


sdispater:
I don't think we can do that without breaking the API contract of a decorator. A decorator is supposed to return the initial function not something else. If we do that we remove the possibility to add further decorators to the function (for instance, I might want my function to also be a `click` command).

And, ideally, a decorator should not change the signature of the function (which was the case before #15). Adding a decorator should be as transparent and painless as possible.

ewjoachim:
I'm going to express some disagreement here, but it does not mean I want to overrule your opinion. I would like to know ore about it, and work toward a better solution. By all means, let's talk about this.

> A decorator is supposed to return the initial function not something else

It seems to me that the very facto that [functools.wraps](https://docs.python.org/3.7/library/functools.html#functools.wraps) exists and is advised is saying that you're expected to return a different function than the one your recieved ?

There's only some very limited things a decorator could do if it has to return the same function. Maybe register it somewhere, or maybe attaching an attribute to it, but a decorator cannot change the behaviour of a function with this rule. Is there an example of a decorator in the standard library that returns the initial function ?

In the standard library, @contextmanager does something very similar to what this PR advocates, it replaces the function that is returned with an instance of a function implementing __enter__ and __exit__.

Regarding the change of signature, I'd kinda agree that it's better if we don't but if the decorator makes it extra explicit that it's going to change the signature, it's just a matter of communication. E.g. `@pytest.mark.parametrize` does that and we're fine with it. So does Celery when you use `bind=True` in the decorator, it adds "self" as first argument.

sdispater:
@ewjoachim Yes, you're right regarding the fact that it should not necessarily return the initial function but only a function (basically, you do not want to change the type of the decorated object). It was a bad choice of words on my part.

> Regarding the change of signature, I'd kinda agree that it's better if we don't but if the decorator makes it extra explicit that it's going to change the signature, it's just a matter of communication. E.g. @pytest.mark.parametrize does that and we're fine with it. So does Celery when you use bind=True in the decorator, it adds "self" as first argument.

That's why I said "ideally" and at least by default it should not do so. However, I agree that there are cases when you want that information and, as such, you should explicitly require it.

What I meant is that we can't directly return the `Task` object in the decorator but we can return a wrapper function.

> I'm going to express some disagreement here

In the end there is not much of a disagreement here :-)

mgu:
> What I meant is that we can't directly return the Task object in the decorator but we can return a wrapper function.

why ? it's a callable :)

sdispater:
One more thing though:

> So does Celery when you use bind=True in the decorator, it adds "self" as first argument.

As a developer I would not use it since you are leaking implementation details into your own code and ties you to a third party library which I like to avoid. This is especially true if you choose to use a decorator-based task/queue library since you would expect to be able to start using it on existing functions (or replacing a previous library). This is less true for a class-based library since you have to commit to the interface of the base class anyway.

I know this comes from the discussion on #15 and that we would like to have some sort of information about the current task. I think there are better ways to do so than modifying the function's signature but if this is the direction we want to go in, by using an explicit parameter, that's also fine with me. In the end, it's just a slight difference in vision of API design (and you know I tend to have a strong opinion on API design :-)) but I don't think it's incompatible and we can work something out.

sdispater:
@mgu Because you lose the type of the decorated object. A callable is not the same thing as function. A function decorator should expect to receive a function. For instance, I might want in my decorator to retrieve the name of the function by using `__name__` but if the `task` decorator returns a `Task` object it will raise an error.

ewjoachim:
> you know I tend to have a strong opinion on API design

And I know it's a complicated craft, and among the most valuable things you bring to the team, so that's why I want to find a solution that you find cool too :)

> I think there are better ways to do so than modifying the function's signature

Do you have any in mind ?

> For instance, I might want in my decorator to retrieve the name of the function by using __name__ but if the task decorator returns a Task object it will raise an error.

This is the role of functools's updatewrapper (or wraps)

```python
>>> from functools import update_wrapper
>>> 
>>> class A:
...     def __init__(self, func):
...         self.func = func
...     def __call__(self):
...         return self.func()
... 
>>> def decorator(func):
...     return update_wrapper(A(func), func)
... 
>>> @decorator
... def yay():
...     """
...     ho
...     """
...     print("hey")
... 
>>> yay()
hey
>>> print(yay.__name__)
yay
>>> print(yay.__doc__)

    ho
    
```

If we wrap the function properly so that it keeps all the __ attributes the original had, and it's indistinguishable from the original function except it's an instance of "Task" instead of an instance of "function", would you be ok ?

sdispater:
> If we wrap the function properly so that it keeps all the __ attributes the original had, and it's indistinguishable from the original function except it's an instance of "Task" instead of an instance of "function", would you be ok ?

My only concern is about the public API, so if the behavior of the decorated function is virtually the same regardless of what we actually return in the decorator that's good for me :-)

And to be honest, I am not completely happy with my previous solution (I'd prefer not to tamper with the function by adding methods to it at runtime) so I'm glad we're heading towards a common one.

ewjoachim:
\o/

Now, let's add `update_wrapper` to the call and fix the tests, then.

Thank you *a lot* @sdispater and @mgu !

sdispater:
I just updated the code and the tests. All is green!

ewjoachim:
Ooooh, very nice ! thank you :)

sdispater:
Well, I validated the initial idea (not the implementation which was a common one anyway) and another review would be nice.

ewjoachim:
@k4nar that was done :) LGTY ?

tests/unit/test_tasks.py
```diff
@@ -110,3 +110,23 @@ def test_task_manager_default_connection(mocker):
...
+    @manager.task()
+    def wrapped(a):
+        """bar"""
+        return "foo"
+
+    # Check that even when wrapped within a task, the original function
+    # keeps its name, doc, arg list and is still callable
+    assert wrapped.__name__ == "wrapped"
+    assert wrapped.__doc__ == "bar"
+    assert wrapped(a=1) == "foo"
```
ewjoachim: Ping @k4nar

ewjoachim:
(à la relecture, le test que j'ai ajouté existait déjà précédemment, j'ai enlevé le doublon)

PR #20, opened 2019-04-21
Fix coverage

### LINKED ISSUE #18: Fix coverage
Coverage is reported under 100%, but when launched locally, coverage is 100% Something is wrong in codecov.

ewjoachim:
Closes #18 

As far as I can tell, the problem is that we're computing coverage
independently in the unit, integration and acceptance tests.

Different solutions include:
- Always putting the --cov-append flag, but it might lead to erroneous
reports
- Switching matrix to individual runs. It could be interesting to
identify erroneous builds quicker but at the price of setupping the VM
more times (and it seems it's actually the longest part of the CI)
- Making it so that --cov-append is only passed in the context of the
CI, which is what we're doing here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
   - tox --notest -e $TOX_ENV
 
 script:
-  - tox -e $TOX_ENV
+  - tox -e $TOX_ENV -- --cov-append
 
 after_success:
 - codecov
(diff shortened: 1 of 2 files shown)


ewjoachim:
The main interest of switching to the more granular matrix could be to add coverage tags and then be able to follow independently the unit, integration and acceptance coverage.

YAGNI.

PR #22, opened 2019-04-21
Add the notion of "JobStore"

### LINKED ISSUE #3: Testing tools

ewjoachim:
Fixes #3 

This PR solves 2 related problems in one:

- Postgres was still highly coupled with several parts
- Because of that, it was harder to test (especially for cabbage users)

In order to implement #3, I added the notion of "JobStore", with 2 implemented stores:
- PostgresJobStore
- InMemoryJobStore (in a module named "testing")

This allows us to remove a lot of mocks in tests and all, which feels nice.

Only thing left to decide: how to test the PostgresJobStore that is a thin wrapper around equivalent functions in the module ?
- Unit-test it (and mock the function ?)
- Integration-test it (tests are going to look a lot like the tests of the functions it uses)
- Acceptance-test it (that's already done) (it's going to be the first lines that are neither covered by integration nor unit tests)
- Rewrite it. For example, put the code in the methods directly instead of the function. Not a fan, but we could discuss it.

### DIFF
--- a/cabbage/postgres.py
+++ b/cabbage/postgres.py
@@ -111,8 +111,8 @@ init_pg_extensions()
 
 
 class PostgresJobStore(store.JobStore):
-    def __init__(self, connection_params: Dict[str, Any] = None):
-        self.connection = get_connection(**(connection_params or {}))
+    def __init__(self, connection: Optional[psycopg2._psycopg.connection] = None):
+        self.connection = connection or get_connection()
 
     def register_queue(self, queue: str) -> Optional[int]:
         return register_queue(connection=self.connection, queue=queue)
--- a/cabbage/postgres.py
+++ b/cabbage/postgres.py
@@ -1,5 +1,5 @@
 import select
-from typing import Any, Dict, Iterator, Optional
+from typing import Iterator, Optional
 
 import psycopg2
 from psycopg2 import extras, sql
--- a/cabbage/postgres.py
+++ b/cabbage/postgres.py
@@ -1,17 +1,11 @@
-from typing import Iterator, NamedTuple, Optional
+import select
+from typing import Any, Dict, Iterator, Optional
 
 import psycopg2
 from psycopg2 import extras, sql
 from psycopg2.extras import RealDictCursor
 
-from cabbage import exceptions, types
-
-
-class TaskRow(NamedTuple):
-    id: int
-    task_type: str
-    args: types.JSONDict
-    targeted_object: str
+from cabbage import exceptions, jobs, store, types
 
 
 def get_connection(**kwargs) -> psycopg2._psycopg.connection:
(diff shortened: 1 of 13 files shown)


cabbage/testing.py
```diff
@@ -0,0 +1,43 @@
...
+            jobs.Job(id=id, task_name=name, lock=lock, kwargs=kwargs, queue=queue)
+        )
+        return id
+
+    def get_tasks(self, queue: str) -> Iterator[jobs.Job]:
+        for job in list(self.jobs[queue]):
+            yield job
+
+    def finish_task(self, job: jobs.Job, status: jobs.Status) -> None:
+        print(self.jobs[job.queue])
```
ewjoachim: woops

cabbage/postgres.py
```diff
@@ -103,4 +103,37 @@ def listen_queue(connection: psycopg2._psycopg.connection, queue: str) -> None:
...
 
 
+def wait_for_jobs(connection: psycopg2._psycopg.connection, timeout: int):
+    select.select([connection], [], [], timeout)
+
+
 init_pg_extensions()
+
+
+class PostgresJobStore(store.JobStore):
```
ewjoachim: Test the class instead of the functions

tests/unit/test_tasks.py
```diff
@@ -2,12 +2,13 @@
...
 import pytest
 
-from cabbage import tasks
+from cabbage import jobs, postgres, tasks, testing
 
 
 @pytest.fixture
-def manager():
-    return tasks.TaskManager(object())
+def manager(mocker):
```
ewjoachim: task_manager

tests/unit/test_task_worker.py
```diff
@@ -1,11 +1,29 @@
 import pytest
 
-from cabbage import exceptions, postgres, task_worker, tasks
+from cabbage import exceptions, task_worker, tasks, testing, jobs
 
 
 @pytest.fixture
-def manager():
-    return tasks.TaskManager(connection=object())
+def manager(mocker):
```
ewjoachim: task_manager

cabbage/tasks.py
```diff
@@ -39,14 +40,31 @@ def defer(self, lock: str = None, **kwargs: types.JSONValue) -> None:
 
 
 class TaskManager:
-    def __init__(self, job_store: store.JobStore = None) -> None:
+    def __init__(self, job_store: store.JobStore = None):
```
ewjoachim: job_store_path

PR #26, opened 2019-05-03
Add support for job scheduling

sdispater:
This PR adds two new configuration option for jobs: `schedule_at` and `schedule_in`.

* `schedule_at` must be a timezone-aware datetime representing the moment the job must be executed.
* `schedule_in` must be a dictionnary representing the relative time the job must be executed at

```python
import pendulum

manager = TaskManager()


def add(a, b):
    return a + b

add.configure(schedule_at=pendulum.now().add(minutes=2)).defer(a=1, b=2)
add.configure(schedule_in={"minutes": 2}).defer(a=1, b=2)
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ packages = find:
 install_requires =
     psycopg2
     attrs
+    pendulum
 
 [options.extras_require]
 dev =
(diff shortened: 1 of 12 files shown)


setup.cfg
```diff
@@ -23,6 +23,7 @@ packages = find:
 install_requires =
     psycopg2
     attrs
+    pendulum
```
ewjoachim: add `pendulum.*` at https://github.com/peopledoc/cabbage/pull/26/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R61 to please mypy.

(and maybe add type annotations in a compatible way as decribed in https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages?)

sdispater: I updated the `setup.cfg` file.

And regarding type annotations, there is a PR (https://github.com/sdispater/pendulum/pull/320) that I need to merge at some point.

PR #33, opened 2019-07-05
Log task result when job is done (#21)

### LINKED ISSUE #21: Log the result of the task
It could be interesting to accept that task can return anything, and the return value will be added to the logs.

This way, the logs would naturally be enriched without having to add log to the tasks. Of course, a task could always return None if there's nothing smart to say.

I can imagine this being a nice help for ops more often than not.

pmourlanne:
Closes #21 

Not a whole lot to see :o

The only decision is *where* to put the result. I chose to put in its own key, it could be in the `job` log_context. I don't mind either way :)

### DIFF
--- a/cabbage/worker.py
+++ b/cabbage/worker.py
@@ -139,8 +139,9 @@ class Worker:
 
         logger.info("Starting job", extra={"action": "start_job", "job": log_context})
         try:
-            task(**job.task_kwargs)
+            task_result = task(**job.task_kwargs)
         except Exception as e:
+            task_result = None
             log_title = "Job error"
             log_action = "job_error"
             log_level = logging.ERROR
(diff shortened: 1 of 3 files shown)


cabbage/worker.py
```diff
@@ -139,8 +139,10 @@ def run_job(self, job: jobs.Job) -> None:
 
         logger.info("Starting job", extra={"action": "start_job", "job": log_context})
         try:
-            task(**job.task_kwargs)
+            # We don't know if task return something and we don't really care
+            task_result = task(**job.task_kwargs)  # type: ignore
```
ewjoachim: In this case, task should return `Any` ?

pmourlanne: Yes, you're right :+1: 

(First time with mypy :>)

ewjoachim: :+1:

PR #34, opened 2019-07-05
Added a order by when fetching jobs (#5)

### LINKED ISSUE #5: Add an order by clause in get_task
If tasks are to be executed in order (especially when they share the same lock) then we need to always select the first one, pk wise.

pmourlanne:
Closes #5 

I don't really see how to properly test this? The current tests, like `test_get_jobs` already cover this?
Your ideas are welcome :>

### DIFF
--- a/init.sql
+++ b/init.sql
@@ -71,6 +71,7 @@ BEGIN
 			  AND cabbage_job_locks.object IS NULL
 			  AND status = 'todo'
 			  AND (scheduled_at IS NULL OR scheduled_at <= now())
+            ORDER BY id ASC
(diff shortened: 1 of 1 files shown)


pmourlanne:
I added the `ORDER BY` in `cabbage_fetch_job`.
I'm still unsure on testing this :o It doesn't break anything so that's good.

ewjoachim:
(Suggestion, not obligation, please merge when you feel like it’s ready)

ewjoachim:
Hmm, you know what ? i’ll merge now and if you want to do a test, you’re welcome in another PR :)

PR #35, opened 2019-07-12
Add a function to get stalled jobs

### LINKED ISSUE #7: Get stalled tasks
Get a list of `Job()`s that are:
- in a given queue or not
- with a given name or not
- in the `doing` state for more than `X` secondes

pmourlanne:
Closes #7

### DIFF
--- a/tests/integration/test_postgres.py
+++ b/tests/integration/test_postgres.py
@@ -181,6 +181,45 @@ def test_get_jobs(job_store):
     ]
 
 
+def test_get_stalled_jobs(connection, get_all, job_store):
+    job_store.register_queue("queue_a")
+    job_store.launch_job(
+        jobs.Job(
+            id=0,
+            queue="queue_a",
+            task_name="task_1",
+            lock="lock_1",
+            task_kwargs={"a": "b"},
+            job_store=job_store,
+        )
+    )
+    job_id = list(get_all("cabbage_jobs", "id"))[0]["id"]
+
+    # No started job
+    assert list(job_store.get_stalled_jobs(nb_seconds=3600)) == []
+
+    # We start a job and fake its `started_at`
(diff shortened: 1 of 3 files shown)


pmourlanne:
Je n'arrive pas du tout a ecrire ma fonction SQL :O Je me rassure en me disant que c'est la fatigue du vendredi ...

ewjoachim:
This is not a time/lock sensitive operation so I’m not sure we need a function for this. I’d say you can code the query in the python file directly

pmourlanne:
All right, I don't know if this makes it easier for me but I'll look into it :)
Thanks :+1:

pmourlanne:
Turns out it *is* much easier when I'm writing "classic" SQL :> Thank you :)
PR should be ready for review, I'm removing the draft status.

tests/integration/test_postgres.py
```diff
@@ -180,6 +181,38 @@ def test_get_jobs(job_store):
...
+            job_store=job_store,
+        )
+    )
+
+    # No started job
+    assert list(job_store.get_stalled_jobs(nb_seconds=3600)) == []
+
+    # We start a job and wait a little
+    job = next(job_store.get_jobs("queue_a"))
+    time.sleep(1)
```
pmourlanne: Not particularly fond of this :O Since we are using psql `NOW` I don't see an easy way to cheat our way out of this :\

ewjoachim: `UPDATE cabbage_task SET started_at='2000-01-01T00:00:00+00:00'::timestamp with timezone WHERE id = %(id)s` ?

pmourlanne: J'ai fait quelque chose comme ca, merci :+1:

PR #43, opened 2019-08-04
Stop quicker (don't wait for timeout)

### LINKED ISSUE #41: Interrupt the "select" call on signal handle
The idea is to register a pipe, then in the select call in postgres.py, listen to the pipe too, and in the signal handler, write in this pipe. This way, we'd leave immediately without waiting for the timeout.

I know we can't do everything in a signal handler, but if we can write to a pipe, then it may help us to stop faster (and make the "timeout" argument really much less important).

https://stackoverflow.com/a/4661284

ewjoachim:
Closes #41

### DIFF
--- a/cabbage/postgres.py
+++ b/cabbage/postgres.py
@@ -1,5 +1,6 @@
+import os
 import select
-from typing import Iterable, Iterator, Optional
+from typing import Iterable, Iterator, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


ewjoachim:
This PR is integrated in #51. Maybe it's not useful to merge it (it introduces conflicts that are already solved over there.

PR #47, opened 2019-08-07
Plugin system

ewjoachim:
@mgu 
I didn't take the time to fix all the bugs, but... The more I see it, the less I like it. I'd like your opinion !

(of course, some independant commits here may be salvaged)


ewjoachim:
#50 does it better

PR #49, opened 2019-08-09
Implement retry strategies for failed job attempts

pmourlanne:
Closes https://github.com/peopledoc/cabbage/issues/6

PR #50, opened 2019-08-09
Instantiate job store

ewjoachim:
Based on #49

This PR:
- Finally makes up cabbages's mind regarding App instanciation: it takes an instance of JobStore
- Adds __version__ and other metadata at the top level
- Add building of the docs in the CI, we realized it was missing

### DIFF
--- a/cabbage/retry.py
+++ b/cabbage/retry.py
@@ -1,0 +1,43 @@
+import datetime
+from typing import Optional, Union
+
+import attr
+import pendulum
+
+from cabbage import exceptions
+
+
+@attr.dataclass(kw_only=True)
+class RetryStrategy:
+    max_attempts: Optional[int] = None
+    wait: int = 0
+
+    def get_retry_exception(self, attempts: int) -> Optional[exceptions.JobRetry]:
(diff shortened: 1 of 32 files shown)


cabbage/retry.py
```diff
@@ -0,0 +1,82 @@
...
+    Parameters
+    ----------
+    max_attempts:
+        The maximum number of attempts the job should be retried
+    wait:
+        The number of seconds to wait between attempts
+    """
+
+    max_attempts: Optional[int] = None
+    wait: int = 0
```
mgu: is it a safe default ?

ewjoachim: If you just ask to "retry" without specifying to wait and without specifying how many times, I'd say yes. Who know if 1s is more appropriate than 1 day. Who knows if 3 times is more appropriate than 300 times ?

ewjoachim: Worse case scenario, we'll... Try. again and again and again.

ewjoachim: Tenacity does exactly the same, and they basically know about retry decorators ! https://tenacity.readthedocs.io/en/latest/#basic-retry :)

PR #51, opened 2019-08-09
Reorganize timeout

ewjoachim:
Now that the job store is independently instantiated, it makes more sense that the socket timeout is an implementation detail of the store rather than of the whole app. This simplifies a lot of things. Also, reintergrate #43 that was beginning to drift seriously.

### DIFF
--- a/cabbage/postgres.py
+++ b/cabbage/postgres.py
@@ -1,6 +1,7 @@
 import datetime
+import os
 import select
-from typing import Any, Iterable, Iterator, Optional
+from typing import Any, Iterable, Iterator, Optional, Tuple
 
 import psycopg2
 from psycopg2 import extras, sql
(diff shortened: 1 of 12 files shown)


cabbage/postgres.py
```diff
@@ -162,14 +166,27 @@ class PostgresJobStore(store.BaseJobStore):
     connection to a Postgres database.
     """
 
-    def __init__(self, *args: Any, **kwargs: Any):
+    def __init__(
+        self, socket_timeout: float = SOCKET_TIMEOUT, *args: Any, **kwargs: Any
```
ewjoachim:
```
self, *, socket_timeout: float = SOCKET_TIMEOUT, **kwargs: Any
```

ewjoachim:
@mgu I refactored the tests you spoke about and added the one you said was missing. Also, @SdgJlbl I added comments on that test.

PR #62, opened 2019-08-25
Create task cli

ewjoachim:
Based on #61

Add a `procrastinate defer <task name> <json args>`.

I believe that with that, we don't need to provide a complicated cron-like feature: people can now directly use cron (or systemd equivalent) to run tasks, without having to write custom python scripts.

We keep doing what we do best, not overcomplicate the codebase, while providing the best interface to ease integration.


ewjoachim:
😮 Coverage went down and codecov didn't warn 😭

PR #64, opened 2019-08-27
linear and exponentiall backoff

ewjoachim:


docs/howto/retry.rst
```diff
@@ -23,28 +28,41 @@ Available strategies are:
...
+- ``wait=5`` to wait 5 seconds between each retry
+- ``linear_wait=5`` to wait 5 seconds then 10 then 15 and so on
+- ``exponential_wait=5`` to wait 5 seconds then 25 then 125 and so on
+
+Implementing your own strategy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 - If you want to go for a fully fledged custom retry strategy, you can implement your
-  own retry strategy::
+  own retry strategy (though we recommand always keeping a max_retry)::
```
pmourlanne:
```suggestion
  own retry strategy (though we recommend always keeping a max_retry)::
```

procrastinate/retry.py
```diff
@@ -46,24 +46,39 @@ def get_schedule_in(self, attempts: int) -> Optional[int]:
...
+        total_wait = wait + lineal_wait * attempts + exponential_wait ** attempts
 
     Parameters
     ----------
     max_attempts:
         The maximum number of attempts the job should be retried
     wait:
-        The number of seconds to wait between attempts
+        Number of seconds to wait between attempts will follow a constant series
+        (e.g. if 3, then successive runs will wait 3, 3, 3, 3, 3 seconds)
```
pmourlanne: Les documentations de ces trois arguments ne sont pas claires du tout :o
Je propose ca, c'est un debut de mieux je trouve :
```
        The number of seconds used to compute the constant time to wait between retries
        (e.g. if 3, then successive runs will wait 3, 3, 3, 3, 3 seconds)
```
A generaliser sur les `linear_wait` et `exponential_wait`

ewjoachim: (don't forget we're in an open source repo :us: :gb:)

`The number of seconds used to compute the linear/exponential time to wait between retries` don't strike me as a really better solution, but I admit writing those lines was hard. I still think we could find a better solution.

pmourlanne: :new_zealand: 
The current sentence `Number of seconds to wait between attempts will follow a constant series` doesn't make sense to me :o We can ask an actual native for help :>

ewjoachim: :canada: 
Ah I said "series" instead of "sequence", sorry. Would that have helped, tabarnak ?

pmourlanne: New iteration?
```
Use this if you want to use a constant backoff. Give a number of seconds as argument, it will be used to compute the backoff: if 3, then successive runs will wait 3, 3, 3, 3, 3 seconds.
```

ewjoachim: Can I let you add the appropriate commits on top of this PR ? Thanks :)

pmourlanne: Done :+1:

ewjoachim: Thanks ❤️

PR #68, opened 2019-08-30
Add a maintenance task to clean old jobs

### LINKED ISSUE #65: pre-register a maintenance task that will remove old tasks in a configurable way
It's up to the user to run the task periodically, but we could provide a task to ease this. Something along the way of:

```
procrastinate defer procrastinate.builtin_tasks.remove_old_jobs max_hours=72 remove_error=false
```

It's kinda explicit what it does: delete all finished jobs and optionally also error jobs. When removing error jobs, it produces a log that contains all its informations, so that log exploitation could maybe recover them if need be.

I'm not sure whether it's more relevant to use days or hours but hours feels like it's easier for both use cases.

Note: I've used a syntax for cli defer that has not been implemented yet but I'm considering.

pmourlanne:
Closes #65

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -7,6 +7,16 @@ App
 .. autoclass:: procrastinate.App
     :members: task, run_worker
 
+App.builtin_tasks
+"""""""""""""""""
+Builtin tasks are registered when the App object is initialized, and
+are accessible through :py:attr:`App.builtin_tasks`:
+
+.. code-block:: python
+
+    app.builtin_tasks["remove_old_jobs"].defer(max_hours=72)
+
+
 Job stores
 ----------
 
@@ -19,6 +29,16 @@ Tasks
 -----
 .. autoclass:: procrastinate.tasks.Task
     :members: defer, configure
+
+
+Builtin tasks
+-------------
+
+Procrastinate has builtin tasks that are all available from the CLI:
+
+.. _remove_old_jobs_reference:
+
+.. autofunction:: procrastinate.builtin_tasks.remove_old_jobs(max_hours, queue=None, remove_error=False)
 
 
 Retry strategies
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -2,7 +2,7 @@ import functools
 import logging
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Optional, Set
 
-from procrastinate import migration
+from procrastinate import builtin_tasks, migration
 from procrastinate import retry as retry_module
 from procrastinate import store, utils
 
@@ -53,8 +53,11 @@ class App:
         """
         self.job_store = job_store
         self.tasks: Dict[str, "tasks.Task"] = {}
+        self.builtin_tasks: Dict[str, "tasks.Task"] = {}
         self.queues: Set[str] = set()
         self.import_paths = import_paths or []
+
+        self._register_builtin_tasks()
 
     def task(
         self,
@@ -127,6 +130,9 @@ class App:
                 extra={"action": "register_queue", "queue": task.queue},
             )
             self.queues.add(task.queue)
+
+    def _register_builtin_tasks(self) -> None:
+        builtin_tasks.register_builtin_tasks(self)
 
     def _worker(self, queues: Optional[Iterable[str]] = None) -> "worker.Worker":
         from procrastinate import worker
--- a/procrastinate/builtin_tasks.py
+++ b/procrastinate/builtin_tasks.py
@@ -1,0 +1,29 @@
+from typing import Optional
+
+from procrastinate import store
+
+
+def remove_old_jobs(
+    job_store: store.BaseJobStore,
+    max_hours: int,
+    queue: Optional[str] = None,
+    remove_error: Optional[bool] = False,
+) -> None:
+    """
+    This task cleans your database by removing old jobs. Note that jobs and linked
+    events will be irreversibly removed from the database when running this task.
+
+    Parameters
+    ----------
+    max_hours :
+        Only jobs which were finished more than ``max_hours`` ago will be deleted.
+    queue :
+        The name of the queue in which jobs will be deleted. If not specified, the
+        task will delete jobs from all queues.
+    remove_error :
+        By default only successful jobs will be removed. When this parameter is True
+        failed jobs will also be deleted.
+    """
+    job_store.delete_old_jobs(
+        nb_hours=max_hours, queue=queue, include_error=remove_error
+    )
--- a/procrastinate/builtin_tasks.py
+++ b/procrastinate/builtin_tasks.py
@@ -1,3 +1,4 @@
+import functools
 from typing import Optional
 
 from procrastinate import store
@@ -26,4 +27,13 @@ def remove_old_jobs(
     """
     job_store.delete_old_jobs(
         nb_hours=max_hours, queue=queue, include_error=remove_error
+    )
+
+
+# Register your builtin tasks here
+def register_builtin_tasks(app) -> None:
+    app.builtin_tasks["remove_old_jobs"] = app.task(
+        functools.partial(remove_old_jobs, app.job_store),
+        queue="builtin",
+        name="procrastinate.builtin_tasks.remove_old_jobs",
     )
--- a/procrastinate/postgres.py
+++ b/procrastinate/postgres.py
@@ -31,6 +31,21 @@ WHERE status = 'doing'
   AND (%(task_name)s IS NULL OR task_name = %(task_name)s)
 """
 
+delete_old_jobs_sql = """
+DELETE FROM procrastinate_jobs
+WHERE status IN %(statuses)s
+  AND (%(queue)s IS NULL OR queue_name = %(queue)s)
+  AND (id IN (
+      SELECT DISTINCT ON (job_id) job_id
+          FROM procrastinate_events
+      WHERE at < NOW() - (%(nb_hours)s || 'HOUR')::INTERVAL
+      ORDER BY job_id, at DESC
+      LIMIT 1
+    )
+  )
+"""
+
+
 finish_job_sql = """
 SELECT procrastinate_finish_job(%(job_id)s, %(status)s, %(scheduled_at)s);
 """
@@ -117,6 +132,20 @@ def get_stalled_jobs(
                     "scheduled_at": row["scheduled_at"],
                     "queue": row["queue_name"],
                 }
+
+
+def delete_old_jobs(
+    connection: psycopg2._psycopg.connection,
+    nb_hours: int,
+    statuses: Iterable[str],
+    queue: str = None,
+) -> None:
+    with connection:
+        with connection.cursor(cursor_factory=RealDictCursor) as cursor:
+            cursor.execute(
+                delete_old_jobs_sql,
+                {"nb_hours": nb_hours, "queue": queue, "statuses": tuple(statuses)},
+            )
 
 
 def finish_job(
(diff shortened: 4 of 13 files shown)


pmourlanne:
I'm not satisfied with how the builtin tasks are loaded (you have to load them by hand one by one in `App`).

I'm also unsure as to how access the job store from the `remove_old_jobs` task. Maybe I went down the wrong road :o

procrastinate/app.py
```diff
@@ -128,6 +130,9 @@ def _register(self, task: "tasks.Task") -> None:
             )
             self.queues.add(task.queue)
 
+    def _register_builtin_tasks(self) -> None:
+        self._register(self.task(builtin_tasks.remove_old_jobs, queue="builtin"))
```
ewjoachim: I'd rather call a function defines in the builtin_tasks module here (with the app parameter).

Also, I'd rather call `app.task` than register.

I'm guessing the problem is that you don't yet know the app instance when you define the builtin tasks, but we could have:

```python
def my_builtin():
	...

BUILTIN_TASKS_FUNCS = [
	my_builtin,
]
def register_builtins(app: ...) -> None:
	for func in BUILTIN_TASKS_FUNCS:
		app.task(func)
```

Finally, a test would check that all the functions defined in this module (except a few ones) are in BUILTIN_TASKS_FUNCS, so that the risk of forgetting one would be low ?

ewjoachim: Even better:
```python
app.task(functools.partial(func, app=app))
```

Maybe with

```python
@pass_app
def my_builtin():
	...
```

This way, your job can access the app, and thus the job store, and thus make a query.

procrastinate/postgres.py
```diff
@@ -31,6 +31,14 @@
   AND (%(task_name)s IS NULL OR task_name = %(task_name)s)
 """
 
+select_old_jobs_sql = """
+SELECT job.id, task_name, lock, args, scheduled_at, queue_name
+    FROM procrastinate_jobs job
+WHERE status IN %(statuses)s
+  AND started_at < NOW() - (%(nb_hours)s || 'HOUR')::INTERVAL
+  AND (%(queue)s IS NULL OR queue_name = %(queue)s)
+"""
```
ewjoachim: I think it may be the moment where we could introduce a more abstract way to make queries:

```
class BaseJobStore:
	def perform_sql(self, query:str, **arguments) -> Iterable[Dict[str, Any]]:
		...
	def select_sql(self, query:str, **arugments) -> Iterable[Dict[str, Any]]:
		...
```

ewjoachim: Hmm, let's do this in a separate PR.

procrastinate/app.py
pmourlanne: The issue is indeed that you can't access the app instance when defining the builtin task :o

> Also, I'd rather call app.task than register.

Isn't it necessary to `register` the task to call it by CLI later on?

I tried `functools.partial` and it fricks up the name resolution later on. The root issue being that a `partial` doesn't have a path. Maybe there is a better solution than what I did, or we could try to power through.

procrastinate/postgres.py
ewjoachim: I've done it in #71 , now it's a race :D

procrastinate/app.py
ewjoachim: >> Also, I'd rather call app.task than register.
> Isn't it necessary to register the task to call it by CLI later on? (This is not my main issue, I'm trying to understand this particular remark :>)

Oh, I had not noticed that you're actually calling `self.task` (which is `app.task`). This method instantiates a task *and* registers it, so no need to call `_register` yourself, just call self.task as you did.

> I tried functools.partial and it fricks up the name resolution later on. The root issue being that a partial doesn't have a path.

This is not a problem if you define the task name explicitely (`name=...`). And you're free to have this name be the builtin path you'd expect. As long as a task is registered, its name doesn't have to be it's path. The dynamic import is only done for unregistered tasks.

pmourlanne: Ok for `self.task` :+1: 

You can give a `name` if you want, but the code will still check for the `full_path`: https://github.com/peopledoc/procrastinate/blob/master/procrastinate/tasks.py#L39 And it crashes since a partial does not have a path :o
(It crashes further down if you comment this block)

Not saying it's a deal breaker or anything, but it's less simple than you make it out to be :>

ewjoachim: yeah, sorry about that.

I think the crashing is the problem. One should be able to register a task that has no path. I'd say we can remove the warning. And fix whatever crashed further down ?

pmourlanne: Seems like a good call yes :) I'll look into it on Friday

procrastinate/app.py
```diff
@@ -128,6 +131,13 @@ def _register(self, task: "tasks.Task") -> None:
             )
             self.queues.add(task.queue)
 
+    def _register_builtin_tasks(self) -> None:
+        self.task(
+            functools.partial(builtin_tasks.remove_old_jobs, self.job_store),
+            queue="builtin",
+            name="procrastinate.builtin_tasks.remove_old_jobs",
+        )
```
ewjoachim: Can we have this in the builtins module and the _register_builtin_tasks() call the fonction in the builtins ? That should make it easier to not forget to add tasks here.

pmourlanne: I'm not sure I understood what you meant :o I did something, tell me if it's ok with you :>

ewjoachim: Yep :+1:

procrastinate/builtin_tasks.py
```diff
@@ -11,4 +11,7 @@ def remove_old_jobs(
     queue: Optional[str] = None,
     remove_errors: Optional[bool] = False,
 ) -> None:
-    pass
+    jobs = job_store.get_old_jobs(
+        nb_hours=max_hours, queue=queue, include_error=remove_errors
+    )
+    job_store.delete_jobs(job_ids=[job.id for job in jobs])  # type: ignore
```
ewjoachim: There's a race condition... Why not `DELETE WHERE` ?

pmourlanne:
PR ready for review :rocket: 
The individual commits are a bit ugly, I can try to rebase it all if you prefer, tell me :)

procrastinate/app.py
```diff
@@ -128,6 +130,9 @@ def _register(self, task: "tasks.Task") -> None:
             )
             self.queues.add(task.queue)
 
+    def _register_builtin_tasks(self) -> None:
+        builtin_tasks._register_builtin_tasks(self)
```
ewjoachim: Le _ dans le nom de la méthode, c'est utile, mais celui dans la fonction... Je sais pas... Ca veut pas dire non, mais, je me demande.

procrastinate/postgres.py
ewjoachim: Hm, actually I've done a part of that in #71 but not the whole.

procrastinate/app.py
pmourlanne: :gb:

Which one :> ? The `builtin_tasks._register_builtin_tasks` function or the `App._register_builtin_tasks` method?
For the method I think it makes sense. For the function I don't really have an opinion, whatever you prefer :)

ewjoachim: (hey, for the record, I translated to :gb: less than a minute after the original comment :open_mouth: )
I'm talking about the function, not the method.

I'm ok leaving the underscore because people might look at the module when searching for a builtin task name, but... I wouldn't oppose a commit removing it either :D

ewjoachim:
@pmourlanne We're approaching the end. How about the documentation now ? I think the best place for this would be a "how-to" for cleaning up old tasks, and a section in the reference documentation that lists available builtin tasks (but maybe not as auto-doc ?)

procrastinate/app.py
pmourlanne: I removed the leading `_`

pmourlanne:
Docu-what? No entiendo :es: 

Ok for the "how-to" :+1:  What did you have in mind for the reference section? Who would it be aimed at?

ewjoachim:
For the reference doc, it's about the fact we now have a new "programmatic" way of interacting with the project (mainly launching builtin task). The name and nature of the parameters should be documented like a part of the programmatic api.

pmourlanne:
@ewjoachim I added some documentation and exposed the builtin tasks through `app.builtin_tasks["remove_old_jobs"]` as we talked about.

procrastinate/builtin_tasks.py
```diff
@@ -17,7 +17,7 @@ def remove_old_jobs(
 
 # Register your builtin tasks here
 def register_builtin_tasks(app) -> None:
-    app.task(
+    app.builtin_tasks["procrastinate.builtin_tasks.remove_old_jobs"] = app.task(
```
ewjoachim: do we really need to reiterate `procrastinate.builtin_tasks` ?

pmourlanne: I changed my mind the commit after :)

docs/howto/clean_old_jobs.rst
```diff
@@ -0,0 +1,25 @@
...
+From the CLI
+^^^^^^^^^^^^
+
+.. code-block:: console
+
+    $ procrastinate defer procrastinate.builtin_tasks.remove_old_jobs max_hours=72 remove_error=true
+
+The task takes 2 paremeters related to which jobs to delete:
+
+- ``max_hours``: Jobs which were created more than ``max_hours`` ago will be deleted
```
ewjoachim: Hmm... Why are we filtering on the creation date and not the completion date ? It means a 3 days old job that runs just now might be deleted right after running... I'm not sure it's what we want...

docs/reference.rst
```diff
@@ -7,6 +7,10 @@ App
 .. autoclass:: procrastinate.App
     :members: task, run_worker
 
+App.builtin_tasks
+"""""""""""""""""
+TODO
```
ewjoachim: TODO ?

docs/reference.rst
```diff
@@ -21,6 +25,18 @@ Tasks
...
+Builtin tasks
+-------------
+
+Procrastinate has builtin tasks that are all available from the CLI.
+
+Clean old jobs
+^^^^^^^^^^^^^^
+
+This task lets you delete old jobs and linked events from the database. For more
+information, see :ref:`the how-to section <clean_old_jobs>`.
```
ewjoachim: This is the reference doc, we should document the signature of the task, no ?

ewjoachim: You may want to do this the other way around: a quick presentation in the how-to section, pointing here, and the details of the signature here.

Note: you can still make it look like other functions, either with [this](https://stackoverflow.com/a/30584552) trick or by using a plain [`py:function`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html?highlight=method#directive-py:function)

docs/howto/clean_old_jobs.rst
pmourlanne: I can change this :+1:

pmourlanne:
The code now deletes old jobs based on their "finish" time and not their "start" time.
I also rebased the PR into 5 commits, it should be easier to read :)

procrastinate/postgres.py
```diff
@@ -31,6 +31,21 @@
   AND (%(task_name)s IS NULL OR task_name = %(task_name)s)
 """
 
+delete_old_jobs_sql = """
+DELETE FROM procrastinate_jobs
+WHERE status IN %(statuses)s
+  AND (%(queue)s IS NULL OR queue_name = %(queue)s)
+  AND (id IN (
+      SELECT DISTINCT ON (job_id) job_id
```
ewjoachim: This could be written with a JOIN, no ?

procrastinate/builtin_tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Parameters
+    ----------
+    max_hours :
+        Only jobs which were finished more than ``max_hours`` ago will be deleted.
+    queue :
+        The name of the queue in which jobs will be deleted. If not specified, the
+        task will delete jobs from all queues.
+    remove_error :
+        By default only succeeded jobs will be removed. When this parameter is True
```
ewjoachim: successful jobs

procrastinate/postgres.py
pmourlanne: I don't know how to do it, I can look into it :)

pmourlanne:
The issue to improve the SQL: https://github.com/peopledoc/procrastinate/issues/79

PR #70, opened 2019-08-31
Add a "procrastinate_events" table logging job attempts

### LINKED ISSUE #48: Add a "procrastinate_events" table logging job attempts
Linked to https://github.com/peopledoc/cabbage/issues/6

When a job is re-tried, its `started_at`, `scheduled_at` (and one day `ended_at`?) are overwritten. We should make sure not to lose this data, for example by storing it in a "cabbage_events" table.
This table could be populated when a job is created / started / ended. As we see it, it should contain the following columns:
- A foreign key to a job
- A type of event: `schedule`, `started`, `ended` (more?)
- A datetime with timezone field, containing what is today in `cabbage_job.scheduled_at` / `cabbage_job.started_at`
- The current `job.attempts` (make sure it makes sense for `scheduled`, `started` and `ended` events)

`started_at` and `ended_at` (if it exists) may be removed from the job table. `scheduled_at` is still relevant on the job table. Although it may be renamed into "next_attempt_scheduled_at" (for example).

ping @mgu @ewjoachim

ewjoachim:
Fixes #48

### DIFF
--- a/procrastinate/structure.sql
+++ b/procrastinate/structure.sql
@@ -1,29 +1,29 @@
 CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
 
-CREATE TYPE public.procrastinate_job_status AS ENUM (
+CREATE TYPE procrastinate_job_status AS ENUM (
     'todo',
     'doing',
     'done',
     'error'
 );
 
-CREATE TABLE public.procrastinate_jobs (
+CREATE TABLE procrastinate_jobs (
     id integer NOT NULL,
     queue_name character varying(128) NOT NULL,
     task_name character varying(128) NOT NULL,
     lock text,
     args jsonb DEFAULT '{}' NOT NULL,
-    status public.procrastinate_job_status DEFAULT 'todo'::public.procrastinate_job_status NOT NULL,
+    status procrastinate_job_status DEFAULT 'todo'::procrastinate_job_status NOT NULL,
     scheduled_at timestamp with time zone NULL,
     started_at timestamp with time zone NULL,
     attempts integer DEFAULT 0 NOT NULL
 );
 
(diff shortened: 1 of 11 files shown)


procrastinate/structure.sql
```diff
@@ -79,22 +96,82 @@ BEGIN
...
+    AS $$
+BEGIN
+    WITH t AS (
+        SELECT CASE
+            WHEN OLD.status = 'todo'::procrastinate_job_status
+                AND NEW.status = 'doing'::procrastinate_job_status
+                THEN 'started'::procrastinate_job_event_type
+            WHEN OLD.status = 'doing'::procrastinate_job_status
+                AND NEW.status = 'todo'::procrastinate_job_status
+                THEN 'retried'::procrastinate_job_event_type
```
pmourlanne: Maybe `retried` should be `to_be_retried`?

ewjoachim:
> Do we want to add a documentation explaining what each event means, when it appears etc.? It would help at least internally :o

Definitely !

ewjoachim:
@PierreDucroquet I'd like your opinion regarding:
- 46a82e130e4acfe5b0a18574723c94457695f6e5 where I use a bigserial instead of an integer NOT NULL with a sequence installed manually
- 0cefc8f2ece596f93cdc0d9b27271d3e61f880b4 where I define a PRIMARY key from the table instead of adding a constraint afterwards.

I'm guessing that you wrote it that way for a reason, so before changing it, I want to make sure it's for the better :)

ewjoachim:
We have a :+1: from Marc (can't ping him because he didn't contribute to the repo)

procrastinate/structure.sql
ewjoachim: `deferred_retry` ?

pmourlanne: `deferred_for_retry` sounds :ok_hand:

PR #71, opened 2019-09-03
Sql in dedicated file

ewjoachim:
It started from putting the SQL in a dedicated file
Continued with a much needed refacto of get_jobs > get_job
Which prompted a refacto of the worker
All of which had me rewrite tests for the testing module, and uncover a few bugs :o


procrastinate/retry.py
```diff
@@ -23,6 +23,7 @@ def get_retry_exception(self, attempts: int) -> Optional[exceptions.JobRetry]:
             return None
 
         schedule_at = pendulum.now("UTC").add(seconds=schedule_in)
+        print("retry", schedule_in, schedule_at)
```
pmourlanne: :eyes:

ewjoachim: whoopsie.

ewjoachim:
> A bit hard to read (not your fault :>) but it looks like a nice improvement 👍

Any advice on how to make it better next time ?

pmourlanne:
Diffs of refactoring / renaming commits tend to look bad. I don't think you can do more than you already did :o
The clear separation of commits was very appreciated :)

PR #73, opened 2019-09-06
Add Codeowners and PR template

ewjoachim:
In this PR, we name @mgu @pmourlanne and I as codeowners, and @CorBott as documentation codeowner. I've written the file so that no one ever is a bottleneck.

Also, I'm adding a PR template which should help us remember to always always (always) add some documentation. (and maybe get some feedback about the contribution process)


ewjoachim:
I don't plan to request the repo to require code owner validation, but if you think we should, we can discuss this.

PR #77, opened 2019-09-09
Spelling

ewjoachim:
Adding a spell check validation on the CI because:
- I write bad
- It's annoying to see these typos afterwards
- Machines are sometimes better than humans for spotting those
- It's usually less judgmental to take criticism on your spelling from a machine than from a human.

### Successful PR Checklist:
- [x] Tests < (well new CI checks anyway)
- [x] Documentation

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,5 +2,5 @@ Cf. #<ticket number>
 
 ### Successful PR Checklist:
 - [ ] Tests
-- [ ] Documentation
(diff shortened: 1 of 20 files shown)


ewjoachim:
Ok, I'm not sure about this.
I like the idea of it, but sphinxcontrib-spelling is built on top of PyEnchant (unmaintained), itself built on top of enchant (C lib, installed system-wide).

In order to make it work on the CI, I have to install enchant manually AND apparently, the default dictionary is different from the one on my laptop, which makes PRs that pass locally and fail in the CI. Red flags galore.

Can you folks identify an alternative ? A sane path forward ?

ewjoachim:
Ok, I think the best to do with this PR is to integrate it as an operational tool that won't break the CI and won't prevent people from contributing if it doesn't work, but will let us check easily from times to times.

ewjoachim:
Ready for review !

Issue #79, opened 2019-09-10
Improve SQL to get old jobs


pmourlanne:
See https://github.com/peopledoc/procrastinate/pull/68#discussion_r322609635

PR #80, opened 2019-09-13
Use JOIN for `delete_old_jobs_sql`

pmourlanne:
Closes https://github.com/peopledoc/procrastinate/issues/79

Little Pierre discovers SQL


pmourlanne:
I just rebased, since the SQL was moved around since I started working on this :)

ewjoachim:
I was just about to tackle this one, you're right on time :)

PR #91, opened 2019-09-20
Make it possible to launch tasks not defined in the project

### LINKED ISSUE #86: Defer a job without knowing it
Add a `app.defer()` method that allows deferring a job even if you don't have the Task object (for example if the task deferring and the task running are done in 2 different codebases)

Signature would be something like:

```python
def defer(self, task_name:str, [all the arguments from task.defer]):
```

ewjoachim:
Closes #86 

Ping @alexdashkov 

### Successful PR Checklist:
- [x] Tests
- [x] Documentation

### DIFF
--- a/docs/howto/defer.rst
+++ b/docs/howto/defer.rst
@@ -1,0 +1,78 @@
+Defer a job
+-----------
+
+(See :ref:`glossary <glossary_defer>`)
+
+There are several ways to do this.
+
+In the following examples, the task will be::
+
+    @app.task(queue="some_queue")
+    def my_task(a: int, b:int):
+        pass
+
+Task name is ``my_module.my_task``.
+
+The direct way
+^^^^^^^^^^^^^^
+::
+
+    my_task.defer(a=1, b=2)
+
+
+With parameters
+^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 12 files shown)


docs/howto/defer.rst
```diff
@@ -0,0 +1,78 @@
...
+::
+
+    pattern = my_task.configure(task_kwagrs={"a": 1})
+
+    pattern.defer(b=2)
+    pattern.defer(b=3)
+    pattern.defer(b=4)
+
+
+Defer a job if you can't access the task
```
ewjoachim: @alexdashkov This is the part that will matter to you.

tests/unit/test_cli.py
```diff
@@ -100,14 +100,26 @@ def test_get_schedule_in(input, output):
...
 
 
-def test_get_task_error(app):
+def test_configure_job_unknown(app):
+    job = cli.configure_job(app, "foobar", {}, unknown=True).job
+    assert job.task_name == "foobar"
+    assert job.queue == "default"
+
+
+def test_test_configure_job_error(app):
```
ewjoachim: test_test :D

Issue #85, opened 2019-09-16
Add a very simple cron mechanism


ewjoachim:
I'm being told the project would be easier to use if we provided a simple cron-like process. It might be interesting to look.

@EliotBerriot, you mentionned you were maybe interested to try a contribution ?

EliotBerriot:
Yep, I can definitely have a look!

My initial idea, to keep this simple, was to have a separate, optionnal process, similar to `celery beat` scheduler.This process would be responsible to insert jobs in the database whenever the configured crontabs are matched. These jobs would be processed as usual, by regular workers.

For the cron-logic itself, if we're open to using dependencies, https://github.com/dbader/schedule looks like a pretty solid choice, meaning we'd only have to implement the configuration and job insertion logic.

There would'nt be any catchup mecanism: if the scheduler is down, upon restart, all missed tasked would simply be ignored.

If that sounds good to you, I can try to submit a PR this week :)

ewjoachim:
👍 for me, @mgu a second opinion ?

ewjoachim:
Does schedule 

I could see a 2nd decorator on tasks, like
```python
@app.schedule.every().day
@app.task(...)
def bla():
	pass
```
(if you don't call the `.do(job)` in the decorator, you allow procrastinate to do it itself with the job of its choice, i.e. a specialized `defer` function.)

(OR a global configuration, but then not sure what it should look like. Do you think people may want to control Cron tasks with configuration rather than code ? I honnestly don't know. All I know is that Celery does both.)

I'd say you probably need to raise if a cron task has parameters. Or require it accepts a single `CronContext` argument or something.

pmourlanne:
We don't *have* to insert the jobs when they need to be run. We could take advantage of the `job.scheduled_at` column :o

This is just food for thought, your solution looks good :)

mgu:
I don't like the idea of implementing a cron-like (or celery beat). I don't understand this need as we could already use cron/systemd.timers or something else that already exists.

EliotBerriot:
>  I don't really understand this need as we could already use cron/systemd.timers or something else that already exists.

@mgu maybe I can elaborate on my use case then :)

For Funkwhale, we have potentially dozens or hundreds of separate installations of the software. Some of these are deployed by people with little to no sysadmin knowledge. 

In that context, it's really, really easier to ask them to setup a service with one command, `celery beat` than setting up a dozen cron jobs or services, especially since the list of scheduled job can change with each version ;)

PR #92, opened 2019-09-21
WIP #85: simple cron mechanism

EliotBerriot:
Cf. #85 

So, I got a proof-of-concept working, and it was simpler than expected. I've used the excellent https://github.com/agronholm/apscheduler library, which seemed well adapted to our use case.

In a nutshell:

1. If you want to have scheduled jobs, you add a `schedule` parameter when creating the app. This is a list containing dictionaries, each dictionary describes a job, arguments to pass to the job, and periodicity (using cron-like syntax, cf https://apscheduler.readthedocs.io/en/latest/modules/triggers/cron.html#module-apscheduler.triggers.cro
2. Then, you run `procrastinate scheduler` to start the scheduler process
3. This process will `defer` the corresponding jobs when needed

As you can see, all of this is rather simple and fit in ~100 lines of code. Of course, it's still lacking tests, documentation and everything, but I wanted to check an implementation of this kind would work for you ;)

Let me know what you think about it!

### Successful PR Checklist:
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -31,6 +31,7 @@ class App:
         *,
         job_store: store.BaseJobStore,
         import_paths: Optional[Iterable[str]] = None,
+        schedule: Optional[Iterable[dict]] = None,
     ):
         """
         Parameters
@@ -50,12 +51,16 @@ class App:
             A :py:func:`App.task` that has a custom "name" parameter, that is not
             imported and whose module path is not in this list will
             fail to run.
+        schedule:
+            List of dictionaries describing jobs that must be run periodically.
+            These jobs are inserted by the scheduler process.
         """
         self.job_store = job_store
         self.tasks: Dict[str, "tasks.Task"] = {}
         self.builtin_tasks: Dict[str, "tasks.Task"] = {}
         self.queues: Set[str] = set()
         self.import_paths = import_paths or []
+        self.schedule = schedule
 
         self._register_builtin_tasks()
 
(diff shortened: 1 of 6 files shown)


mgu:
I was thinking about this feature for a long time and I'm wondering how we could implement distributed cron jobs with high availability (what happens if we run `procrastinate scheduler` more than once ?)

mgu:
And I'm really sorry we didn't answer earlier.
It may take time to understand what feature we want to implement or not :)

EliotBerriot:
> I was thinking about this feature for a long time and I'm wondering how we could implement distributed cron jobs with high availability (what happens if we run procrastinate scheduler more than once ?)

Hm that's a good question. Right now, I assume you'd end up with duplicate jobs (or triplicate if you run it 3 times, etc.).

However, maybe this could work?

1. We add a nullable, `schedule_key` column, with a `UNIQUE` constraint.
2. The scheduler process:
  - Insert jobs as usual, but populate the `schedule_key` column with a hash of the : job name, corresponding crontab and args
  - Catch and ignore integrity errors on the `schedule_key` unique constraint

This way, you can run as many schedulers at you want with the same crontab, and you're guaranteed to not have duplicate jobs entry thanks to the UNIQUE constraint in postgres.

> And I'm really sorry we didn't answer earlier.

It's ok, don't worry :)

ewjoachim:
> populate the schedule_key column with a hash of the : job name, corresponding crontab and args

We also need the official launch time, otherwise on a 5 minute repeating job, the first instance will run but not the following ones.

So...
- We will need a "cron job id" that can be the job name, args and crontab
- `(1)` We need our "cron worker" to be able to compute the most recent launch time for every cron job
- And check the ones that don't already exist in the DB
- And create them (I don't know if it's best to rely on a UNIQUE column check exception or to somehow lock the table for inserts so we know we don't have a race condition)
- Then compute the interval until the next cron job will need to run (+ a bit of noise to avoid all the workers to get up at the same millisecond)
- Then sleep this amount of time
- `goto (1)`

ewjoachim:
(there's the `INSERT ... ON CONFLICT DO NOTHING` that could help us)

PR #96, opened 2019-10-04
Async 9 new approach

### LINKED ISSUE #9: Support async mode (aiopg ?)
In this ticket, we'll make the first part: async defering tasks

The rest will be done in #105 and #106

### LINKED ISSUE #105: If tasks are coroutines, await them.
This is the second step in asynchronous compatibility.

- #9 is for defering jobs asynchronously
- This ticket is for accepting a task to be a coroutine
- #106 is for executing asynchronous tasks in parallel

ewjoachim:
Closes #105 

Ping @sophie-ulti @pmourlanne @mgu 
Wow, I would never have guessed it was so direct.
This PR makes the procrastinate core async. Not async friendly, fully async. We add a top level sync API that calls `asyncio.run` underneath.
The PR is quite short (some moves and removal, but very little new code). (EDIT: lol, this was True when the tests were failing, but fixing them made the PR bigger). It's missing the tests (not that we would need to write a lot of tests, but we've broken the existing ones almost beyond repair). The demo works.
See todolist below for things to do before we merge.

### Successful PR Checklist:
- [x] Tests
  - [x] Repair existing tests
  - [x] The lock feature seams broken though... invesstigate (the acceptance lock test is flaky) (edit: #103 )
  - [x] Add tests related to async task execution
  - [x] Review the acceptance test, it can now be launched identically with sync and async (they're now on feature par)
- [x] Documentation
  - Not much to document because the external API didn't change, but it would be nice to say a word or two in the "Discussions" section on why procrastinate has a sync API that is the default, and it's implemented as an async project.
  - Also, we need to document async task execution
- [x] It seems that when we ctrl+c, it waits until timeout to quit. reading the code, I don't understand why, but let's investigate
- [x] Mypy doesn't like our metaprogramming trick `@async.add_sync_api` and frankly, I understand it makes its job harder. Let's help it by implementing a mypy plugin. We can take direct inspiration from https://github.com/strawberry-graphql/strawberry/blob/4d11e2d26cf3f8f1453c95106606acede85125ac/strawberry/ext/mypy_plugin.py and https://github.com/strawberry-graphql/strawberry/tree/4d11e2d26cf3f8f1453c95106606acede85125ac#type-checking (doc: https://mypy.readthedocs.io/en/latest/extending_mypy.html )
   > EDIT: handled: we'll do it in another ticket. For now, `#typing: ignore` will do.
- [x] Chances are, if we start executing jobs asynchronously (and thus in parallel), our very naive implementation of the current log context in worker.py is now completely off. Not 100% sure though.
  > EDIT: Thinking about it, launching tasks with async/await does not mean launching them in parallel. As long as we don't introduce a task pool of sorts, the tasks will still be ran sequentially.
- [x] Had a good time contributing?  Hell yeah !

Note for a subsequent PR:
We'll need something else in order for async task execution to really mean something: for now, our tasks can be asynchronous, but there's still only one task executed at a time. We need a way to launch a pool of workers in the same event loop.
We need to check precisely where the right place is to place our pool. I don't know yet if we want 1 DB connection per process or per worker, and the potential impact on transactions. We'll probably need to setup a connection pool etc.


ewjoachim:
We'll need something else in order for async task execution to really mean something: for now, our tasks can be asynchronous, but there's still only one task executed at a time. We need a way to launch a pool of workers in the same event loop.
We need to check precisely where the right place is to place our pool. I don't know yet if we want 1 DB connection per process or per worker, and the potential impact on transactions.

k4nar:
This looks amazing :heart: .

tests/acceptance/app.py
```diff
@@ -28,3 +29,11 @@ def two_fails():
     if nb_tries < 2:
         nb_tries += 1
         raise Exception("This should fail")
+    print("Yay")
```
ewjoachim: Add a comment because this looks like a forgotten print XD

procrastinate/aiopg_connector.py
```diff
@@ -52,14 +85,41 @@ def __init__(self, *, socket_timeout: float = store.SOCKET_TIMEOUT, **kwargs: An
...
+        try:
+            await asyncio.wait_for(
+                connection.notifies.get(), timeout=self.socket_timeout
+            )
+        except asyncio.futures.TimeoutError:
+            pass
+
+    def stop(self):
+        if self._connection:
+            asyncio.get_running_loop().call_soon_threadsafe(
```
ewjoachim: Ah there's no such thing as ``asyncio.get_running_loop`` in python 3.6 :D

ewjoachim: Maybe `get_event_loop` will be enough. Let's check.

ewjoachim:
Ready for review ! I'd like to thank (and ping) @sophie-ulti and @pmourlanne for your extremely helpful contributions to this PR !

@CorBott Use this link to review the PR and see only the documentation changes :)
https://github.com/peopledoc/procrastinate/pull/96/files?file-filters%5B%5D=.rst

@k4nar feel free to have a look, I have only added commits since your comment, not modified, so this should be short.

@mgu I know you love short scoped PRs, well, I hope you'll forgive me one day :) (honnestly, I think it's as short and scoped as can be, which is not very short, but still quite scoped. So many lines where the only change is async/await...)

Leaving this PR to your wise review. I'd like to merge this quickly if possible.

procrastinate/aiopg_connector.py
```diff
@@ -1,26 +1,70 @@
...
+from procrastinate import store
 
 
-def get_connection(**kwargs) -> Awaitable[aiopg.Connection]:
-    return aiopg.connect(**kwargs)
+def wrap_json(arguments: Dict[str, Any]):
+    return {
+        key: Json(value) if isinstance(value, dict) else value
+        for key, value in arguments.items()
+    }
```
k4nar: Why not registering an adapter for dict instead?

procrastinate/aiopg_connector.py
```diff
@@ -1,26 +1,70 @@
...
+
+
+def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
+    return aiopg.connect(dsn=dsn, **kwargs)
+
+
+async def execute_query(
+    connection: aiopg.Connection, query: str, **arguments: Any
+) -> None:
+    # Strangely, aiopg can work with psycopg2's cursor class.
```
k4nar: aiopg is just a layer on top of psycopg2, this isn't that weird ;)

procrastinate/worker.py
```diff
@@ -131,6 +132,9 @@ def run_job(self, job: jobs.Job) -> None:
         logger.info("Starting job", extra={"action": "start_job", "job": log_context})
         try:
             task_result = task(**job.task_kwargs)
+            if asyncio.iscoroutinefunction(task.func):
```
k4nar: Why not `asyncio.iscoroutine(task_result)` instead? That way we can handle regular functions returning coroutines.

ewjoachim: Realized the same thing when writing the doc, and I though it would be a nice follow up ticket in the future, but now that you've said it out loud, I guess I have no choice :D

procrastinate/aiopg_connector.py
ewjoachim: Yeah, but it's quite the abstraction leak... Having an `import psycopg2` in a project that doesn't define psycopg2 as a dependency...

procrastinate/aiopg_connector.py
ewjoachim: Because I had kinda forgotten that psycopg2's adapters were... A thing?

Thankfully, I think there's a reference implementation lying around somewhere :sweat_smile: (like in the codebase of my main project that I shamelessly stole from yours ;) )

ewjoachim: I suggest we let this be done in #99

procrastinate/worker.py
ewjoachim: Done

README.rst
```diff
@@ -74,6 +74,25 @@ Similarly, from the command line:
...
+
+    # Launch a job asynchronously
+    await sum.defer_async(a=3, b=5)
+
+    # Somewhere in your program, run a worker asynchronously
+    worker = procrastinate.Worker(
+        app=app,
+        queues=["sums"]
+    )
+    await worker.run_async()
```
mgu:
```suggestion
    await worker.run()
```

procrastinate/app.py
```diff
@@ -12,6 +12,7 @@
 logger = logging.getLogger(__name__)
 
 
+@utils.add_sync_api
```
mgu: :heart:

README.rst
ewjoachim: Hm ? Why so ?

PR #104, opened 2019-10-27
Fixes #103: race condition in locks

### LINKED ISSUE #103: Locks have a race condition
Ok, we have, I think, our first real bug.

## TL;DR

If I launch `procrastinate_fetch_job` quickly twice, I can get 2 jobs that share the same `lock`.

## Proof

```console
$ psql -c 'SELECT * FROM procrastinate_jobs;'
 id | queue_name |           task_name            | lock |   args    | status | scheduled_at | started_at | attempts
----+------------+--------------------------------+------+-----------+--------+--------------+------------+----------
  1 | sleep      | procrastinate_demo.tasks.sleep | yay  | {"i": 20} | todo   |              |            |        0
  2 | sleep      | procrastinate_demo.tasks.sleep | yay  | {"i": 20} | todo   |              |            |        0
(2 rows)

$ psql -c 'SELECT procrastinate_fetch_job(NULL);' & psql -c 'SELECT procrastinate_fetch_job(NULL);'&
[1] 22792
[2] 22793
                                       procrastinate_fetch_job
-----------------------------------------------------------------------------------------------------
 (1,sleep,procrastinate_demo.tasks.sleep,yay,"{""i"": 20}",doing,,"2019-10-27 20:14:12.230384+00",0)
(1 row)

[2]  + 22793 done       psql -c 'SELECT procrastinate_fetch_job(NULL);'
                                       procrastinate_fetch_job
-----------------------------------------------------------------------------------------------------
 (2,sleep,procrastinate_demo.tasks.sleep,yay,"{""i"": 20}",doing,,"2019-10-27 20:14:12.233998+00",0)
(1 row)

[1]  + 22792 done       psql -c 'SELECT procrastinate_fetch_job(NULL);'
```
The second call should have returned `(,,,,,,,,)`.

For the sake of it, if I put the 2 calls in the same transaction, I DO have the expected result:

```console
$ psql -c 'SELECT * FROM procrastinate_jobs;'
 id | queue_name |           task_name            | lock |   args    | status | scheduled_at | started_at | attempts
----+------------+--------------------------------+------+-----------+--------+--------------+------------+----------
  1 | sleep      | procrastinate_demo.tasks.sleep | yay  | {"i": 20} | todo   |              |            |        0
  2 | sleep      | procrastinate_demo.tasks.sleep | yay  | {"i": 20} | todo   |              |            |        0
(2 rows)

$ psql -c 'SELECT procrastinate_fetch_job(NULL), procrastinate_fetch_job(NULL);'
                                      procrastinate_fetch_job                                       | procrastinate_fetch_job
----------------------------------------------------------------------------------------------------+-------------------------
 (1,sleep,procrastinate_demo.tasks.sleep,yay,"{""i"": 20}",doing,,"2019-10-27 20:17:46.28228+00",0) | (,,,,,,,,)
(1 row)
```

## Leads

- When we call [`FOR UPDATE OF procrastinate_jobs`](https://github.com/peopledoc/procrastinate/blob/master/procrastinate/sql/structure.sql#L58), should we somehow also lock `procrastinate_job_locks`? But there's no row to lock yet...
- Should there be a `UNIQUE` constraint on `procrastinate_job_locks.object`? I guess it would help but if we add just that, we'll get a crash, and still not the expected behaviour. Or should we also add a `ON CONFLICT DO NOTHING` statement? The procedure would probably need quite a refactor, because in this case it needs to **not** do `UPDATE procrastinate_jobs SET status = 'doing' **and** return nothing.
- Is it because the `procrastinate_job_locks` is `UNLOGGED`?

ewjoachim:
Cf. #103 

### Successful PR Checklist:
- [ ] Tests < Well the test that allowed us to see the problem was in #96. I've validated locally that the problem disappears, and when #96 is merged, we'll know regressions will be catched.
- [ ] Documentation < not sure what would be interesting to document. The whole bug was explained in #103, and given the suggested fix worked, I'm not sure what more to say 


marco44:
I think your troubles stem from trying to get a lock out of a transaction, hence the lock table?
That's a bit awkward, and dangerous (as you may end up with records in this table that you'll need to garbage collect or whatever).
PostgreSQL has a tool for this, that's called advisory locks. It has many advantages on a lock table:
- no storage, so blazing fast
- can be managed with or without transaction
- disappears automatically when a session is closed
Maybe we should discuss this, I feel you may have better tools to do what you're trying to achieve

ewjoachim:
Do you have an idea why @PierreDucroquet would not have used advisory locks in the 1st place? I'll be glad to discuss it with you, but I'm guessing it will be even more productive if you have an opportunity to discuss with him first.

Also, would advisory locks be considered for jobs that can be set for, say 1h ? Or are they more considered for short-lived locks ?

PierreDucroquet:
I have no idea why I did not use advisory lock in my first attempt.

ewjoachim:
(I'm a little worried by the implications of:

> disappears automatically when a session is closed

in our usecase. Let's discuss this !)

ewjoachim:
After a discussion, it appears we were on the right track. This deserves documentation.

Also, remove "UNLOGGED" on locks table.

procrastinate/sql/structure.sql
```diff
@@ -37,7 +37,7 @@ CREATE TABLE procrastinate_events (
 );
 
 CREATE UNLOGGED TABLE procrastinate_job_locks (
-    object text NOT NULL
+    object text NOT NULL UNIQUE
```
marco44: Other name is PK. The DBA CI doesn't accept this (fixed on the peopledoc/sql-ddl-migration#3251) as it means this table is declared without PK

ewjoachim: I'm definitely fine with PK :smile:

ewjoachim:
@mgu @pmourlanne Now with dow, ready for review.

docs/discussions.rst
```diff
@@ -74,6 +74,53 @@ some projects that really stand out, to name a few:
...
 .. _dramatiq-pg: https://pypi.org/project/dramatiq-pg/
 
+.. _discussion-locks:
+
+About locks
+-----------
+
+Let's say we have a task that writes a character at the end of a file after waiting for
+a random amount of time. This represents real world problem where tasks take an
+unforseeable amount of time and share resources like a database)
```
k4nar:
```suggestion
unforseeable amount of time and share resources like a database.
```

docs/discussions.rst
```diff
@@ -74,6 +74,53 @@ some projects that really stand out, to name a few:
 .. _Dramatiq: https://dramatiq.io/
 .. _dramatiq-pg: https://pypi.org/project/dramatiq-pg/
 
+.. _discussion-locks:
+
+About locks
+-----------
+
+Let's say we have a task that writes a character at the end of a file after waiting for
+a random amount of time. This represents real world problem where tasks take an
```
k4nar:
```suggestion
a random amount of time. This represents a real world problem where tasks take an
```

PR #107, opened 2019-10-29
Improve the schema

marco44:
- Add version table
- Remove unlogged
- Set missing primary key
- Set missing index on foreign key


ewjoachim:
Also, welcome to the project.

PR #113, opened 2019-11-25
Don't overwrite psycopg2 adapters when creating the connection

k4nar:
If an application had registered psycopg2 adapters, the next call to `get_connection` would overwrite them with aiopg's default. As we don't need the adapters registered by aiopg, we can safely tell it not to do that by default.

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -17,7 +17,7 @@ def wrap_json(arguments: Dict[str, Any]):
 
 def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
     # tell aiopg not to register adapters for hstore & json by default, as
-    # those are registered at the module level and could overwritte previously
+    # those are registered at the module level and could overwrite previously
     # defined adapters
     kwargs.setdefault("enable_json", False)
     kwargs.setdefault("enable_hstore", False)
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -16,6 +16,11 @@ def wrap_json(arguments: Dict[str, Any]):
 
 
 def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
+    # tell aiopg not to register adapters for hstore & json by default, as
+    # those are registered at the module level and could overwritte previously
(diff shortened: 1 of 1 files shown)


procrastinate/aiopg_connector.py
```diff
@@ -16,6 +16,11 @@ def wrap_json(arguments: Dict[str, Any]):
 
 
 def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
+    # tell aiopg not to register adapters for hstore & json by default, as
+    # those are registered at the module level and could overwritte previously
```
ewjoachim:
```suggestion
    # those are registered at the module level and could overwrite previously
```

elemoine:
@k4nar @ewjoachim I am not sure about this change.

`aiopg` does [`register_default_json(self._conn)`](https://github.com/aio-libs/aiopg/blob/7f0debf7a45626aa8ec85c8c714550734c0a8cae/aiopg/connection.py#L556), so the registration of the json typecaster only applies to the connection being created. And with procrastinate creating and using its own connections I don't see where there is a risk of overriding application settings. What do you think?

entwanne:
I don't exactly remember where was the global state, but the issue raised in PeopleCore where we explictly ask to not enable these adapters.

Without these two lines, the adapters were enabled when creating the connection in procrastinate, leading to a failure in PeopleCore.

elemoine:
> Without these two lines, the adapters were enabled when creating the connection in procrastinate,

Without these two lines, if the psycopg2 docs are correct, the json and hstore adapters will be enabled **only** for the procrastinate connection that is being created. So I don't see how this can lead to a failure at the application level.

k4nar:
Maybe it was only a failure in our tests then, because we re-use the same connection (due to a global transaction).

elemoine:
> Maybe it was only a failure in our tests then, because we re-use the same connection (due to a global transaction).

Maybe. I'd be tempted to remove these lines, unless someone comes up with a test-case demonstrating that they can have an impact on the application.

k4nar:
We'll test the PR where you do this with our project and let you know.

elemoine:
@k4nar see https://github.com/peopledoc/procrastinate/pull/133. Thanks.

Issue #102, opened 2019-10-27
Add more details to default log messages


ewjoachim:
In #4, we implemented "simple" message logs. They're so simple that they actually lack precision .The details are in the structured part, but it's not printed by defaut. Someone analyzing logs in a structured way will have everything they need, but someone just using the default log formatter will be in the dark.

I think it would be nice if the default logger could log the extra informations too
OR
The default message could be made dynamic (though f-strings) and include the extra info.

ewjoachim:
Closed by #117, thanks @sophie-ulti !

PR #117, opened 2019-12-06
102: add log formatting when deffering a task

sophie-ulti:
Cf. #102 

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -92,7 +92,10 @@ class JobDeferrer:
         job = self.make_new_job(**task_kwargs)
 
         context = job.get_context()
-        logger.debug(f"About to defer job {context['call_string']}")
+        logger.debug(
+            f"About to defer job {context['call_string']}",
+            extra={"action": "about_to_job_defer", "job": context},
+        )
         id = await self.job_store.defer_job(job=job)
         context["id"] = id
         logger.info(
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -59,7 +59,10 @@ class Job:
 
         if context["scheduled_at"]:
             context["scheduled_at"] = context["scheduled_at"].isoformat()
-
+        kwargs_string = ", ".join(
+            f"{key}={value}" for key, value in context["task_kwargs"].items()
+        )
+        context["call_string"] = f"{context['task_name']}({kwargs_string})"
         return context
 
 
@@ -88,13 +91,12 @@ class JobDeferrer:
 
         job = self.make_new_job(**task_kwargs)
 
-        logger.info(
-            "Deferring job",
-            extra={
-                "action": "job_defer",
-                "asynchronous": True,
-                "job": job.get_context(),
-            },
-        )
+        context = job.get_context()
+        logger.debug(f"About to defer job {context['call_string']}")
         id = await self.job_store.defer_job(job=job)
+        context["id"] = id
+        logger.info(
+            f"Deferred job {context['call_string']} with id: {id}",
+            extra={"action": "job_defer", "job": context},
+        )
         return id
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -134,13 +134,18 @@ class Worker:
         # a stop, we can get details on the currently running task
         # in the logs.
         start_time = time.time()
+
         job_context = log_context["job"] = {
             **job.get_context(),
             "start_timestamp": time.time(),
         }
         exc_info: Union[Exception, bool]
 
-        self.logger.info("Starting job", extra={"action": "start_job", **log_context})
+        logger.info(
+            f"Starting job {log_context['call_string']}",
+            extra={"action": "start_job", **log_context},
+        )
+
         try:
             task_result = task(**job.task_kwargs)
             if asyncio.iscoroutine(task_result):
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -36,6 +36,10 @@ class Worker:
     async def run(self) -> None:
         await self.job_store.listen_for_jobs(queues=self.queues)
 
+        self.logger.debug(
+            f"Log start one queues {self.queues}",
+            extra={"action": "start_log_on_queue", "queues": str(self.queues)},
+        )
         with signals.on_stop(self.stop):
             while not self.stop_requested:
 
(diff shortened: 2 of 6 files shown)


procrastinate/jobs.py
```diff
@@ -88,13 +91,12 @@ def make_new_job(self, **task_kwargs: types.JSONValue) -> Job:
 
         job = self.make_new_job(**task_kwargs)
 
+        context = job.get_context()
+        logger.debug(f"About to defer job {context['call_string']}")
```
pmourlanne: Do we want this debug?

sophie-ulti: This is a good question. We were not sure as well. But we had some issue were the job didn't defer, so I guess that can be a usefull DEBUG log.
I'll pass on logs than we will probably lighten a bit

ewjoachim: yeah for me the idea is that when we have a problem, switching to debug should help narrowing the source, so there's (almost) never "too much information" on debug logs. inf, though, should be filtered.

pmourlanne: I like having it here too :) Just wanted to make sure it wasn't a forgotten debug log :+1:

ewjoachim: Ah, no, `logger.debug()` is meant to stay, for the expert users to debug, not for us.

procrastinate/jobs.py
```diff
@@ -59,7 +59,10 @@ def get_context(self) -> types.JSONDict:
 
         if context["scheduled_at"]:
             context["scheduled_at"] = context["scheduled_at"].isoformat()
-
+        kwargs_string = ", ".join(
+            f"{key}={value}" for key, value in context["task_kwargs"].items()
+        )
+        context["call_string"] = f"{context['task_name']}({kwargs_string})"
```
k4nar: Can't we have sensitive information / PII in this? Another issue might be having long arguments, like something JSON encoded or a text.

I understand the value, but I wonder if it's not taking too much risks :) .

ewjoachim: > Can't we have sensitive information / PII in this?

There's nothing in there that won't also be in the extras ? Does it make a difference ?

> Another issue might be having long arguments, like something JSON encoded or a text.

That's right. Given the info will be present in the log, we could limit to, say 300 chars (configurable).

procrastinate/worker.py
```diff
@@ -153,21 +157,23 @@ def load_task(self, task_name) -> tasks.Task:
...
         finally:
-            end_time = log_context["end_timestamp"] = time.time()
-            log_context["duration_seconds"] = end_time - start_time
-            extra = {"action": log_action, "job": log_context, "result": task_result}
-            logger.log(log_level, log_title, extra=extra, exc_info=exc_info)
+            end_time = context["end_timestamp"] = time.time()
+            context["duration_seconds"] = end_time - start_time
+            extra = {"action": log_action, "job": context, "result": task_result}
+            seconds = round(context['duration_seconds'], 3)
+            ftext = f"{log_title} on job {context['call_string']} in {seconds}"
```
ewjoachim:
```suggestion
            ftext = f"{log_title} on job {context['call_string']} in {context['duration_seconds']:.3f}"
```
and you won't need the round anymore.

When in doubt, check https://pyformat.info/

sophie-ulti: Yes ! I forgot, I was maybe a bit lazy last week sorry

procrastinate/worker.py
```diff
@@ -81,8 +91,8 @@ def __init__(
                 job=job, status=status, scheduled_at=next_attempt_scheduled_at
             )
             logger.debug(
-                "Acknowledged job completion",
-                extra={"action": "finish_task", "status": status, **log_context},
+                f"Acknowledged job completion {context['call_string']} with id: {id}",
```
ewjoachim: Where does `id` come from ? And why just on that one log ?

procrastinate/jobs.py
ewjoachim: (let's keep this PR simple and leave this out of scope)

procrastinate/jobs.py
ewjoachim: We're missing the extras in this log.

procrastinate/worker.py
```diff
@@ -36,6 +36,9 @@ def __init__(
     async def run(self) -> None:
         await self.job_store.listen_for_jobs(queues=self.queues)
 
+        self.logger.debug(
+            f"Log start one queues {self.queues}", extra={"queues": str(self.queues)}
```
ewjoachim: All log extras should have an unique "action" (at least, that's what has been done elsewhere)

procrastinate/worker.py
```diff
@@ -169,7 +185,11 @@ def load_task(self, task_name) -> tasks.Task:
             end_time = job_context["end_timestamp"] = time.time()
             job_context["duration_seconds"] = end_time - start_time
             extra = {"action": log_action, **log_context, "result": task_result}
-            self.logger.log(log_level, log_title, extra=extra, exc_info=exc_info)
+            ftext = (
+                f"{log_title} on job {job_context['call_string']} "
```
ewjoachim: so that may create a log saying `Job error, to retry on job task(...)` and "to retry on job" feels strange.

Maybe it would seem more natural with:
```suggestion
                f"{log_title} - Job: {job_context['call_string']} "
```

procrastinate/jobs.py
```diff
@@ -88,13 +91,15 @@ def make_new_job(self, **task_kwargs: types.JSONValue) -> Job:
...
-            "Deferring job",
-            extra={
-                "action": "job_defer",
-                "asynchronous": True,
-                "job": job.get_context(),
-            },
+        context = job.get_context()
+        logger.debug(
+            f"About to defer job {context['call_string']}",
+            extra={"action": "about_to_job_defer", "job": context},
```
ewjoachim:
```suggestion
            extra={"action": "about_to_defer_job", "job": context},
```

procrastinate/worker.py
```diff
@@ -36,6 +36,10 @@ def __init__(
     async def run(self) -> None:
         await self.job_store.listen_for_jobs(queues=self.queues)
 
+        self.logger.debug(
+            f"Log start one queues {self.queues}",
```
elemoine: s/one/on

elemoine: `self.queues` is `None` if the worker listens on all the queues

elemoine:
@sophie-ulti I added a commit that makes the test_nominal more robust, by just checking the very first procrastinate log in stderr. The test passes with that.

procrastinate/worker.py
mgu: IMO the message could be changed, we don't want a log message that says we are (just) starting to log something :)

k4nar: Note that this is a debug log @mgu :) .

sophie-ulti: I'll change that. In fact, we were not sure about adding queues information, I had to removed some because It was too heavy. (like, two if to know if I can log something, for a debug log ...)

Issue #112, opened 2019-11-25
Disable the "Warning task name different from python path" upon request


ewjoachim:
It would be nice to provide a simple way to disable the warning that triggers everytime you define a task with a name that is not it's python import path. Of course, using a logging filter, people can already do that, but it lacks a bit of user friendliness. I'm thinking about a boolean parameter in `App.__init__()` ?

https://github.com/peopledoc/procrastinate/blob/81fe265a7a1865db77f93302cd82e8dcf9b22421/procrastinate/tasks.py#L83-L93

marqov:
hi @ewjoachim !! can I work on this issue and procrastinate :smile: with you ?

ewjoachim:
Absolutely, it would be a delight! There's already a bit of info on the issue, if you need anything, let me know !

PR #118, opened 2019-12-22
disable warning task name different from python path

marqov:
Cf. #112 

My solution is to add the boolean attribute on the decorator of a task.( `@app.task(disable_full_path_warning=True)`. I tried to add tests for this feature but I don't know how to mock ``__name__`` and ``__module__`` attributes of func in Task.

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -67,6 +67,7 @@ class App:
         queue: str = jobs.DEFAULT_QUEUE,
         name: Optional[str] = None,
         retry: retry_module.RetryValue = False,
+        disable_full_path_warning: bool = False,
     ) -> Any:
         """
         Declare a function as a task. This method is meant to be used as a decorator::
@@ -113,7 +114,8 @@ class App:
         def _wrap(func: Callable[..., "tasks.Task"]):
             from procrastinate import tasks
 
-            task = tasks.Task(func, app=self, queue=queue, name=name, retry=retry)
+            task = tasks.Task(func, app=self, queue=queue, name=name, retry=retry, disable_full_path_warning=
+                disable_full_path_warning)
             self._register(task)
 
             return functools.update_wrapper(task, func)
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -114,8 +114,9 @@ class App:
         def _wrap(func: Callable[..., "tasks.Task"]):
             from procrastinate import tasks
 
-            task = tasks.Task(func, app=self, queue=queue, name=name, retry=retry, disable_full_path_warning=
-                disable_full_path_warning)
+            task = tasks.Task(
+                func, app=self, queue=queue, name=name, retry=retry,
+                disable_full_path_warning=disable_full_path_warning)
             self._register(task)
 
             return functools.update_wrapper(task, func)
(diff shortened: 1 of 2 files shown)


ewjoachim:
Thanks a lot for your contribution !
I’ll do my best to have a look asap, but with the holidays I might not be in front on a computer for a few days.

Thanks anyway, and if you see nothing from me, it’s totally ok to ping me by the end of the week.

ewjoachim:
Regarding your commit "Fix flake8": all decisions regarding line lengths, spaces, etc are left to [black](https://github.com/psf/black).

You can run all the autoformatters in one go with:
```console
$ tox -e format
```

There's a section in the contribution doc regarding this:
https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#keep-your-code-clean

If there's anything we messed up (and I expect to be :) ) please let me know !

procrastinate/app.py
```diff
@@ -67,6 +67,7 @@ def task(
         queue: str = jobs.DEFAULT_QUEUE,
         name: Optional[str] = None,
         retry: retry_module.RetryValue = False,
+        disable_full_path_warning: bool = False,
```
ewjoachim: For boolean values, I often find it easier to have positive variables than negative ones, in order to avoid double negatives like `if not disable_full_path_warning` below.

A nice effect is that in "enable_full_path_warning", the "enable" word can be omitted, so `full_path_warning: bool=True`.

Then, as far as I understand, it's called `full_path` because in the code below, it uses a method we've named `full_path`. It's called `full_path` because it concatenates the module path and the name, but from an exterior point view, it's just a path.

I suggest `path_warning: bool = True` which I feel is shorter and as expressive. What do you think ? (it's totally ok if you'd rather go a different route)

One thing we'll need, though, is an explanation in the docstring just below this code. I'm thinking something along the lines of:

```
    path_warning :
        When True (by default), if the name of the task does not match
        the location where the task can be imported, procrastinate will
        emit a logger warning at definition time.
```

ewjoachim:
> I tried to add tests for this feature but I don't know how to mock __name__ and __module__ attributes of func in Task.

These are the existing tests you want to take inspiration from:
https://github.com/peopledoc/procrastinate/blob/master/tests/unit/test_tasks.py#L13-L29

This idea is to add a new test where you change the value of your boolean parameter, and verify that no warning is created.

If you want to unit test the change in app.py, it's going to be there:
https://github.com/peopledoc/procrastinate/blob/master/tests/unit/test_app.py#L13-L36

It will be a bit harder, given the value of the argument is not stored, so you may need to create a dedicated unit test with the `mocker` fixture where you mock `Task` and use `.assert_called_once_with()`. You'll find plenty of examples in the code.

I'm not sure it justifies an integration or acceptance test (but you may think otherwise and it's ok). If you want to write one, tell me, I'll guide you through it :)

ewjoachim:
@elemoine has suggested a much simpler approach, I'm tempted to go for it. You may want to have a look, in #123.

And even if we go this wait, I'd be delighted if you want to take a stab at any other issue in the tracker :)

Thanks again !

PR #123, opened 2020-01-17
Remove the Task name/full_path warning

### LINKED ISSUE #112: Disable the "Warning task name different from python path" upon request

elemoine:
Cf. #112 and #118.

This PR is an alternative to #118. It suggests to completely remove the warning that occurs when creating a task whose name is not its import path.

Motivation/rationale: when creating a task (for example using the `app.task` decorator) it is perfectly valid to provide an arbitrary task name. So, really, procrastinate shouldn't emit a warning when the task name doesn't correspond to the function's dotted name. And we already have a warning message on the worker side when the user didn't specify the task module (with `import_paths`) and when the task module was loaded dynamically.

Closes #112

### Successful PR Checklist:
- [X] Tests (no new tests, but existing tests pass, with no coverage regression)

### DIFF
--- a/procrastinate/tasks.py
+++ b/procrastinate/tasks.py
@@ -67,30 +67,7 @@ class Task:
         self.app = app
         self.func: Callable = func
         self.retry_strategy = retry_module.get_retry_strategy(retry)
-        self.name: str
-        if name:
-            self.name = name
-        else:
-            self.name = self.full_path
-
-        if name:
-            try:
-                full_path = self.full_path
-            except AttributeError:
-                # Can happen for functools.partial for example
-                full_path = ""
-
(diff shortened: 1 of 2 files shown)


elemoine:
This is ready for review.

Issue #87, opened 2019-09-20
Add a app.monitor() method


ewjoachim:
This should:

- Connect to the database and launch a simple query (`SELECT TRUE`)
- Check the existence (and version?) of the table `procrastinate_jobs`
- Count the number of jobs in each status

Return a dict describing all this.

```
def app.monitor(self) -> Dict[str, Any]:
```

ewjoachim:
Closed by #129

PR #129, opened 2020-01-24
First monitor and corresponding cli command

DainDwarf:
Cf. #87 

### Successful PR Checklist:
- [x] Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,3 +4,5 @@ services:
   postgres:
     image: postgres:10
     ports: ["5432:5432"]
+    environment:
+        POSTGRES_DB: procrastinate
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -209,7 +209,7 @@ class App:
         return migration.Migrator(job_store=self.job_store)
 
     @property
-    def health_check_runner(self) -> healthchecks.HealthCheckRunner:
+    def monitor(self) -> healthchecks.HealthCheckRunner:
         return healthchecks.HealthCheckRunner(job_store=self.job_store)
 
     async def close_connection_async(self):
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -207,7 +207,7 @@ class App:
         return migration.Migrator(job_store=self.job_store)
 
     @property
-    def monitor(self) -> healthchecks.HealthCheckRunner:
+    def health_check_runner(self) -> healthchecks.HealthCheckRunner:
         return healthchecks.HealthCheckRunner(job_store=self.job_store)
 
     async def close_connection_async(self):
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -2,9 +2,7 @@ import functools
 import logging
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Optional, Set
 
-from procrastinate import (
-    builtin_tasks, exceptions, healthchecks, jobs, migration,
-)
+from procrastinate import builtin_tasks, exceptions, healthchecks, jobs, migration
 from procrastinate import retry as retry_module
 from procrastinate import store, utils
 
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -2,7 +2,9 @@ import functools
 import logging
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Optional, Set
 
-from procrastinate import builtin_tasks, exceptions, jobs, migration
+from procrastinate import (
+    builtin_tasks, exceptions, healthchecks, jobs, migration,
+)
 from procrastinate import retry as retry_module
 from procrastinate import store, utils
 
@@ -205,6 +207,10 @@ class App:
     @property
     def migrator(self) -> migration.Migrator:
         return migration.Migrator(job_store=self.job_store)
+
+    @property
+    def health_check_runner(self) -> healthchecks.HealthCheckRunner:
+        return healthchecks.HealthCheckRunner(job_store=self.job_store)
 
     async def close_connection_async(self):
         await self.job_store.close_connection()
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -266,11 +266,11 @@ def healthchecks(app: procrastinate.App):
     """
     Check the state of procrastinate.
     """
-    checker = app.monitor
-    db_ok, reason = checker.check_connection()  # type: ignore
+    health_check = app.health_check_runner
+    db_ok, reason = health_check.check_connection()  # type: ignore
     click.echo(f"DB connection: {reason}")
 
-    schema_version = checker.get_schema_version()  # type: ignore
+    schema_version = health_check.get_schema_version()  # type: ignore
     migration_version = Migrator.version
     schema_ok = schema_version == migration_version
     if schema_ok:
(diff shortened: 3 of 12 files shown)


procrastinate/healthchecks.py
```diff
@@ -0,0 +1,37 @@
...
+class HealthCheckRunner:
+    def __init__(self, job_store: store.BaseJobStore):
+        self.job_store = job_store
+
+    async def check_connection_async(self) -> Tuple[bool, str]:
+        try:
+            result = await self.job_store.execute_query_one(
+                query=sql.queries["check_connection"],
+            )
+            return result["check"], "OK"
```
elemoine: `result["check"]` should always be `True` here no?

procrastinate/migration.py
```diff
@@ -5,6 +5,8 @@
 
 @utils.add_sync_api
 class Migrator:
+    version = "0.2.1"
```
elemoine: Will we need to ensure this version number is in sync with that in the SQL file?

elemoine:
Just two questions, otherwise looks good to me!

procrastinate/cli.py
```diff
@@ -258,6 +258,21 @@ def migrate(app: procrastinate.App, run: bool):
...
+@cli.command()
+@click.pass_obj
+@handle_errors()
+def healthchecks(app: procrastinate.App):
+    """
+    Check the state of procrastinate.
+    """
+    checker = app.health_check_runner
+    check_db = checker.check_connection()
+    click.echo(f"DB connection: {check_db}")
```
ewjoachim: I'd rather display "OK" than "True"

docker-compose.yml
```diff
@@ -4,3 +4,5 @@ services:
   postgres:
     image: postgres:10
     ports: ["5432:5432"]
+    environment:
+        POSTGRES_DB: procrastinate
```
ewjoachim: (hm, when you say "for testing" in the commit name, it's about running automated tests, or trying out procrastinate manually ?)

procrastinate/healthchecks.py
```diff
@@ -11,8 +12,8 @@ def __init__(self, job_store: store.BaseJobStore):
         )
         return result['check']
 
-    async def get_db_version_async(self):
+    async def check_db_version_async(self):
         result = await self.job_store.execute_query_one(
             query=sql.queries['get_latest_version'],
         )
-        return result['version']
+        return result['version'] == Migrator.version
```
ewjoachim: I wouldn't be opposed to leaving the `get_db_version_async` the way it was ?
Maybe have both ?
Also, maybe the check should rather raise than return False, so as to attach the details of the failure along with the result ?

procrastinate/cli.py
ewjoachim: Oops never mind

docker-compose.yml
DainDwarf: It's for trying out manually. With this, the postgres docker image automatically builds a `procrastinate` database when starting up. Otherwise, the demo app does not work out of the box.

I initially wanted to put it in another PR, trying to put most of the installation/manual testing in a Dockerfile and docker-compose. I guess I forgot when creating the PR. I can move this to another PR and rebase if you want, to keep the PRs well focused. 🙂

procrastinate/migration.py
DainDwarf: We do. Fortunately, the test `tests/integration/test_healthchecks.py::test_healthchecks` both test that the healthcheck work as expected with a working database, AND that running healthcheck on a new database works, including this version number comparison. So it tests that the version numbers are synced. 😄 

I do agree it's not crystal clear. I don't have a better idea atm, however. :/

docker-compose.yml
ewjoachim: It's totally ok to leave this in the PR. I'm just surprised that we never bumped into the issue before.

procrastinate/migration.py
ewjoachim: The alternative is to parameterize the migration, which we'll probably have to do anyway at some point, but if we can avoid that for a while, it might make things simpler.

A third option would be to remove the `INSERT INTO procrastinate_versions` query from the migration and manage it with a special system. But I'd rather have the whole migrations be "consistent". If there's a way we could have the migration in 1 place and that place be SQL code, I'd love that.

docker-compose.yml
DainDwarf: I think that's because the README tells to export some shell variables prior to launch the docker-compose. And _maybe_ I did not follow the README…

I may try to look into simplifying that part in another PR. Imo, it would be nice if the first launch or manual testing would only require you to do `docker-compose up -d`. We could setup environment in docker-compose and put the demo app in its own container, automating all setup needed.

ewjoachim: yes, I'm totally for it ! (in another PR maybe, though)

So far I've tried to optimize the main use case without closing the door for users who want to use a postgres instance of their own and not the docker one. But if we can simplify that as much as possible, I'll always support that.

procrastinate/healthchecks.py
DainDwarf: Yes, but I find it less surprising to read when using the result of the query.

DainDwarf: Also, it leaves several possible changes open without changing that line: For example, having a more complex check in SQL, or defering the actual DB request to the `job_store.__getitem__`.

procrastinate/healthchecks.py
DainDwarf: I changed into `get_schema_version_async`, without a try/except. It's now the cli that does the necessary check. 🙂

procrastinate/migration.py
DainDwarf: Maybe we could have the insert into the queries.sql, and let the Migrator run that query after doing its migrations?

procrastinate/cli.py
```diff
@@ -258,6 +259,33 @@ def migrate(app: procrastinate.App, run: bool):
...
+    """
+    checker = app.monitor
+    db_ok, reason = checker.check_connection()  # type: ignore
+    click.echo(f"DB connection: {reason}")
+
+    schema_version = checker.get_schema_version()  # type: ignore
+    migration_version = Migrator.version
+    schema_ok = schema_version == migration_version
+    if schema_ok:
+        click.echo(f"DB schema are up-to-date ({schema_version})")
```
ewjoachim:
```suggestion
        click.echo(f"DB schema is up-to-date ({schema_version})")
```

procrastinate/app.py
```diff
@@ -206,5 +206,9 @@ def perform_import_paths(self):
     def migrator(self) -> migration.Migrator:
         return migration.Migrator(job_store=self.job_store)
 
+    @property
+    def monitor(self) -> healthchecks.HealthCheckRunner:
```
ewjoachim: (rename ?)

procrastinate/healthchecks.py
```diff
@@ -0,0 +1,33 @@
...
+    async def check_connection_async(self) -> Tuple[bool, str]:
+        try:
+            result = await self.job_store.execute_query_one(
+                query=sql.queries["check_connection"],
+            )
+            return result["check"], "OK"
+        except Error as e:
+            return False, str(e)
+
+    async def get_schema_version_async(self) -> Tuple[bool, str]:
```
ewjoachim: (wrong type)

tests/integration/test_healthchecks.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+async def test_healthchecks(pg_job_store):
+    """Running healthchecks should be OK on a clean install."""
+    checker = HealthCheckRunner(pg_job_store)
+
+    assert await checker.check_connection_async() == (True, "OK")
+    assert await checker.get_schema_version_async() == Migrator.version
+
+    status_count = await checker.get_status_count_async()
```
ewjoachim: split into 3 tests

PR #130, opened 2020-01-24
Fix bug in get_connection

elemoine:
This fixes a bug in `get_connection` where `get_connection` may return a closed connection.

@ewjoachim please confirm that `get_connection` is meant to return an open connection, and that `get_connection` is supposed to return a new connection when the previous one was closed.

### Successful PR Checklist:
- [X] Tests
- [X] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation)) **not relevant**

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -24,6 +24,10 @@ def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
     return aiopg.connect(dsn=dsn, **kwargs)
 
 
+def connection_is_open(connection: aiopg.Connection) -> bool:
+    return not connection.closed
+
+
 async def execute_query(
     connection: aiopg.Connection, query: str, **arguments: Any
 ) -> None:
@@ -97,7 +101,7 @@ class PostgresJobStore(store.BaseJobStore):
         self.socket_timeout = socket_timeout
 
     async def get_connection(self):
-        if not self._connection:
+        if not self._connection or not connection_is_open(self._connection):
             self._connection = await get_connection(**self._connection_parameters)
         return self._connection
 
--- a/tests/integration/test_pg_store.py
+++ b/tests/integration/test_pg_store.py
@@ -419,3 +419,19 @@ async def test_close_connection_no_connection(pg_job_store):
 async def test_stop_no_connection(pg_job_store):
     pg_job_store.stop()
     # Well we didn't crash. Great.
+
+
+async def test_get_connection_called_twice(pg_job_store):
+    conn1 = await pg_job_store.get_connection()
+    assert not conn1.closed
+    conn2 = await pg_job_store.get_connection()
+    assert conn2 is conn1
+
+
+async def test_get_connection_after_close(pg_job_store):
+    conn1 = await pg_job_store.get_connection()
+    assert not conn1.closed
+    await pg_job_store.close_connection()
+    conn2 = await pg_job_store.get_connection()
+    assert not conn2.closed
+    assert conn2 is not conn1


procrastinate/aiopg_connector.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, *, socket_timeout: float = store.SOCKET_TIMEOUT, **kwargs: An
         self.socket_timeout = socket_timeout
 
     async def get_connection(self):
-        if not self._connection:
+        if not self._connection or self._connection.closed:
```
ewjoachim: On the one hand, I’d rather that function wouldn’t mix logic on our side (re-opening if closed) and intimate knowledge about psycopg2 (detecting if a connection is open or closed by checking the `closed` parameter). Would you mind splitting those 2 (a `is_openned` function that takes an aiopg connection) ? I

tests/integration/test_pg_store.py
```diff
@@ -419,3 +419,16 @@
...
+    conn1 = await pg_job_store.get_connection()
+    assert not conn1.closed
+    conn2 = await pg_job_store.get_connection()
+    assert conn2 is conn1
+    await pg_job_store.close_connection()
+    assert conn1.closed
+    conn2 = await pg_job_store.get_connection()
+    assert conn2 is not conn1
+    await pg_job_store.close_connection()
+    assert conn2.closed
```
ewjoachim: Why the 2 last lines ?
Do we get an error if we don’t close the connection ? In this case, I’d rather have them closed by a fixture so as to not make the test more complicated.

Also, this feels like 2 distinct tests put together (testing that when getting twice it’s the same object, testing that when getting when it’s closed, we get a new one), could we do this as 2 distinct tests ?
That point is debatable though, it’s integration tests, not unit tests, but I feel this is almost unit-test-like-ish. The only reason this is not unit test is that it needs IO. That being said, if we implement the other change above, it would nearly make sense to mock I/Os.

A debate to have :) It’s probably a bit of bike shedding but software design practices at the beginning of a project can be critical down the road.

procrastinate/aiopg_connector.py
elemoine: `self._connection` is an aiopg object!

tests/integration/test_pg_store.py
elemoine: No I think you're right on all points. I'll do as you suggest it.

procrastinate/aiopg_connector.py
elemoine: and `closed` is a a documented property: https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Connection.closed.

ewjoachim: Yes, it's not about it being documented, it's about putting all the code that includes knowledge about a 3rd party in the same place.

Hm, maybe it was the "intimate" part that made my comment harder to understand :/

For me it's all about isolating 3rd party dependencies. I want to try and decouple as much as possible procrastinate from psycopg2/aiopg.

ewjoachim: It doesn't have to be this this way if you feel otherwise. I'm totally OK going your way. If you have specific reasons in mind to do so, it's even better.

elemoine: I must admit that I don't understand what why you'd like :)

Are you suggesting to add a `connection_is_open` function? Like this:

```py
def connection_is_open(self):
    return not self._connection.closed
```

And then? What about `get_connection`? Would you change it to create a new connection when `self.connection_is_open()` returns `False`? If not, how can one get a new connection when the current one has been closed? Should `close_connection` set `self._connection` to `None`?

tests/integration/test_pg_store.py
elemoine: As discussed I've splitted the test into two tests. And there's indeed no need to close the last connection, as it will be closed by the `pg_job_store` fixture.

procrastinate/aiopg_connector.py
ewjoachim: The module is split between the functions (at the top) that do interact with psycopg2 and aiopg, and the class implementation at the bottom that calls the functions, but doesn't really know about psycopg/aiopg. I'm suggesting:

```python
def connection_is_open(connection: aiopg.Connection) -> bool:
    return not connection.closed
```
and then get_connection would ask for `connection_is_open(self._connection)` instead of `self._connection.closed`.

elemoine: Ohh, got you. Sorry for being slow on this!

elemoine:
Branch updated. Thank you for the review  @ewjoachim.

elemoine:
There's another place where we may want to check that the connection has not been closed: the [`stop method`](https://github.com/peopledoc/procrastinate/blob/18e0c42a9b89127521218f53037238982f4d3cae/procrastinate/aiopg_connector.py#L133-L135). I'll take a look at it, and create a new PR if a fix is necessary.

procrastinate/aiopg_connector.py
```diff
@@ -24,6 +24,10 @@ def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
     return aiopg.connect(dsn=dsn, **kwargs)
 
 
+def connection_is_open(connection: aiopg.Connection) -> bool:
+    return not connection.closed
```
k4nar: Is there a rationale to create a separate function for this? :)

elemoine: See the discussion with @ewjoachim above. The functions above the `PostgresJobStore` class definition are the ones dealing with aiopg objects.

k4nar: Ok :+1: .

ewjoachim: If you both think it's a bad idea, maybe it's time I reconsider :D

elemoine: I think it does not hurt at this point really.

k4nar: I find it a bit overzealous but it makes sense, so it's a +0 for me :) .

PR #108, opened 2019-11-15
work in progress: Custom json encoder/decoder

linoor:
Cf. #99 

### Successful PR Checklist:
- [ ] Tests
- [ ] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation))
- [ ] Had a good time contributing? (if not, feel free to give some feedback)

Issue #99, opened 2019-10-21
Custom json encoder/decoder


ewjoachim:
Let users define how they want job payloads to be encoded/decoded, especially if they want to instantiate / format special types.

Ping @hsmett who was up for contributing :)

ewjoachim:
@linoor you mentionned you were interested by this ticket

elemoine:
Closed with #131.

PR #131, opened 2020-01-24
Custom JSON encoders and decoders

elemoine:
Cf. #99 (and #108 for the preliminary work on this).

This PR makes it possible for the user to provide custom JSON encoders and decoders.

The custom encoder and decoder are provided as JSON `dumps` and `loads` functions on the `App` object.

Here's an example dealing with `datetime` objects:

```py
import functools
import json

from procrastinate import App, PostgresJobStore

# Function used for encoding datetime objects
def encode(obj):
    if isinstance(obj, datetime.datetime):
        return obj.isoformat()
    raise TypeError()

dumps = functools.partial(json.dumps, default=encode)

# Function used for decoding datetime objects
def decode(dict_):
    if "dt" in dict_:
        dict_["dt"] = datetime.datetime.fromisoformat(dict_["dt"])
    return dict_

loads = functools.partial(json.loads, object_hook=decode)

app = App(job_store=PostgresJobStore(), dumps=dumps, loads=loads)
```

I think the approach provides a maximum of flexibility to the users. They can use the `json` module the way they want. They can even use a different JSON implementation than `json`, such as [UltraJSON](https://pypi.org/project/ujson/), if they want to.

I am creating this PR for discussion. The code in my branch (attached to this PR) basically works and is tested, although there's still work to do (typing, documentation, more tests maybe, etc.).

Any feedback?

### Successful PR Checklist:
- [x] Tests

### DIFF
--- a/docs/howto/custom_json_encoder_decoder.rst
+++ b/docs/howto/custom_json_encoder_decoder.rst
@@ -1,0 +1,54 @@
+Define custom JSON encoders and decoders
+----------------------------------------
+
+When calling ``mytask.defer()`` to create a job, the task arguments are serialized into
+a JSON string for storing into the Postgres database.
+
+And after fetching a job from the database Procrastinate workers need to deserialize
+the task arguments before calling the task.
+
+By default Procrastinate relies on the JSON dumps and loads functions used by psycopg2.
+(See the `psycopg2 documentation`_.) Procrastinate makes it possible to specify custom
+encoders and decoders.
+
+Here is an example involving serializing/deserializing ``datetime`` objects::
+
+    import functools
+    import json
+
+    from procrastinate import App, PostgresJobStore
+
+    # Function used for encoding datetime objects
+    def encode(obj):
+        if isinstance(obj, datetime.datetime):
+            return obj.isoformat()
+        raise TypeError()
+
+
+    # Function used for decoding datetime objects
+    def decode(dict_):
+        if "dt" in dict_:
+            dict_["dt"] = datetime.datetime.fromisoformat(dict_["dt"])
+        return dict_
+
+    json_dumps = functools.partial(json.dumps, default=encode)
+    json_loads = functools.partial(json.loads, object_hook=decode)
+
+    app = App(job_store=PostgresJobStore(json_dumps=json_dumps, json_loads=json_loads))
+
+In this example the custom JSON dumps and loads functions are based on the standard
+``json`` module's ``dumps`` and ``loads`` functions, with specific ``default`` and
+``object_hook`` for serializing and deserializing ``datetime`` objects, respectively.
+(See the `Python 3 json documentation`_ for more detail.)
+
+This mechanism even makes it possible to use a different JSON implemation than ``json``,
+such as `UltraJSON`_ for example.
+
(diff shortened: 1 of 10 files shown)


DainDwarf:
I was wondering, would it be better to store the functions in the `job_store` rather than the app? I think it would reduce passing around those functions through all the wrapping, and keep the code cleaner. And _imo_ it makes sense that "encode/decoding functions" are configured to the instance of what is dealing with matching jobs on postgresql.

Takin your example, you would only need to change the last line to 
```python
app = App(job_store=PostgresJobStore(dumps=dumps, loads=loads))
```

elemoine:
> I was wondering, would it be better to store the functions in the `job_store` rather than the app?

Great suggestion! I'll look into it…

docs/howto/custom_json_encoder_decoder.rst
```diff
@@ -0,0 +1,49 @@
+Define custom JSON encoders and decoders
+----------------------------------------
+
+When calling ``mytask.defer()`` to create a job, the task arguments are serialized into
+a JSON string for storing into the Postgres database.
+
+And after fetching a job from the database a Procrastinate worker needs to deserialize
```
ewjoachim:
```suggestion
And after fetching a job from the database, Procrastinate workers need to deserialize
```

docs/howto/custom_json_encoder_decoder.rst
```diff
@@ -0,0 +1,49 @@
...
+the task arguments before calling the task.
+
+By default Procrastinate relies on the JSON dumps and loads functions used by psycopg2.
+See the `psycopg2 documentation`_. But Procrastinate makes it possible to specify custom
```
ewjoachim: `See the `psycopg2 documentation` ` entre parenthèses, et ne pas couper la phrase en 2 ou enlever "But"

docs/howto/custom_json_encoder_decoder.rst
```diff
@@ -0,0 +1,49 @@
...
+
+    app = App(job_store=PostgresJobStore(json_dumps=json_dumps, json_loads=json_loads))
+
+In this example the custom JSON dumps and loads functions are based on the standard
+``json`` module's ``dumps`` and ``loads`` functions, with specific ``default`` and
+``object_hook`` for serializing and deserializing ``datetime`` objects, respectively.
+See the `Python 3 json documentation`_ for more detail.
+
+This mechanism even makes it possible to use a different JSON implemation than ``json``,
+such as `UltraJSON`_ for example.
```
ewjoachim: "If your encoding function starts resembling a long list of `if isisntance` calls, you may want to have a look at `functools.singledispatch` (link) for a cleaner way."

elemoine: Done.

docs/howto/custom_json_encoder_decoder.rst
elemoine: Done.

docs/howto/custom_json_encoder_decoder.rst
elemoine: Done.

PR #132, opened 2020-01-24
Add name attribute to workers and print it in the logs

entwanne:
It would be great, when running multiple workers and joining all the logs, to identify which worker produces which log.
The purpose of this PR is to add an optional name parameter to workers, and use this name to format logs.

### Successful PR Checklist:
- [x] Tests

### DIFF
--- a/docs/howto/worker.rst
+++ b/docs/howto/worker.rst
@@ -5,10 +5,11 @@ You can either go towards the CLI route with:
 
 .. code-block:: console
 
-    $ procrastinate --verbose --app=dotted.path.to.app worker [queue [...]]
+    $ procrastinate --verbose --app=dotted.path.to.app worker [--name=worker-name] [queue [...]]
 
 or, identically, use the code way::
 
-    app.run_worker(queues=["queue", ...])
+    app.run_worker(queues=["queue", ...], name="worker-name")
 
 In both cases, not specifying queues will tell Procrastinate to listen to every queue.
+Naming the worker is optional.
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -157,10 +157,12 @@ class App:
 
         return tasks.configure_task(name=name, job_store=self.job_store, **kwargs)
 
-    def _worker(self, queues: Optional[Iterable[str]] = None) -> "worker.Worker":
+    def _worker(
+        self, queues: Optional[Iterable[str]] = None, name: Optional[str] = None
+    ) -> "worker.Worker":
         from procrastinate import worker
 
-        return worker.Worker(app=self, queues=queues)
+        return worker.Worker(app=self, queues=queues, name=name)
 
     @functools.lru_cache(maxsize=1)
     def perform_import_paths(self):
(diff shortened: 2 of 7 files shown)


procrastinate/worker.py
```diff
@@ -25,6 +25,10 @@ def __init__(
         self.app.perform_import_paths()
         self.job_store = self.app.job_store
 
+        self.logger = logging.getLogger(
+            __name__ + f"[{name}]" if name is not None else ""
+        )
```
entwanne: I'm not sure it's a good practice to define a logger per object.
Maybe this can be done with a logging-context variable?

elemoine: I don't know either. I am not too shocked by this object-specific logger :)

elemoine:
+1 from me for the general idea of making it possible to give workers a name.

docs/howto/worker.rst
```diff
@@ -5,10 +5,11 @@ You can either go towards the CLI route with:
 
 .. code-block:: console
 
-    $ procrastinate --verbose --app=dotted.path.to.app worker [queue [...]]
+    $ procrastinate --verbose --app=dotted.path.to.app worker [queue [...]] [--name=worker-name]
```
ewjoachim: In an ideal world, flags should come before positional parameters in a cli

docs/howto/worker.rst
```diff
@@ -5,10 +5,11 @@ You can either go towards the CLI route with:
...
 
 or, identically, use the code way::
 
-    app.run_worker(queues=["queue", ...])
+    app.run_worker(queues=["queue", ...], name="worker-name")
 
 In both cases, not specifying queues will tell Procrastinate to listen to every queue.
+And naming the worker is optional.
```
ewjoachim:
```suggestion
Naming the worker is optional.
```

procrastinate/app.py
```diff
@@ -157,10 +157,12 @@ def configure_task(self, name: str, **kwargs: Any) -> jobs.JobDeferrer:
 
         return tasks.configure_task(name=name, job_store=self.job_store, **kwargs)
 
-    def _worker(self, queues: Optional[Iterable[str]] = None) -> "worker.Worker":
+    def _worker(
+        self, queues: Optional[Iterable[str]] = None, name: Optional[str] = None
```
ewjoachim: I always wonder if optional parameters that are passed should be optional all the way of just at the first step. Any opinion on this ?

procrastinate/worker.py
ewjoachim: I think it would be better if each logger was the child of the « main » logger. Logger paternity is defined by dots in the name. Thus, procrastinate.worker.<name> may be better.

Though, that might be more easily achieved with logger.getChild():
https://docs.python.org/3/library/logging.html#logging.Logger.getChild

ewjoachim:
If it doesn’t add too much complexity, I’d like the worker name to be added in the log context, to ease filtering.

procrastinate/app.py
entwanne: No strong opinion, but yes on internal methods maybe it's not very useful (and a bit error prone) to have default values.

ewjoachim:
Run `tox -e format`, rebase your fixup and we're all good.

entwanne:
Ah, yes, I only run it on `procrastinate` directory and forgot the tests.

PR #133, opened 2020-01-28
Don't set enable_json and enable_hstore to False

elemoine:
Don't set `enable_json` and `enable_hstore` to `False` when calling `aiopg.connect`.

aiopg does `register_default_json(self._conn)`, so the registration of the json typecaster only applies to the connection being created.  And with Procrastinate creating its own connections there's no risk of overriding global application settings.

Cf. #113 

### Successful PR Checklist:
- [ ] Tests
- [X] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation)) **Not relevant**

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -16,11 +16,6 @@ def wrap_json(arguments: Dict[str, Any]):
 
 
 def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
-    # tell aiopg not to register adapters for hstore & json by default, as
-    # those are registered at the module level and could overwrite previously
-    # defined adapters
-    kwargs.setdefault("enable_json", False)
-    kwargs.setdefault("enable_hstore", False)
     return aiopg.connect(dsn=dsn, **kwargs)
 
 


k4nar:
This PR breaks the tests of one of my app (cf #113), but now I'm not sure why :thinking: . The issue is not with `enable_json` but with `enable_hstore`. I do have two separate connections, so it's really not supposed to happen.

ewjoachim:
Let's investigate what aiopg/psycopg2 really does in terms of connection-based registration. If we find a global state of any kind, it might be dangerous to play with does (actually... It might be dangerous to use aiopg/psycopg2 at all :/ )

elemoine:
Interesting. I've looked at the json typecaster, but not so much at the hstore typecaster. I'll do that in the coming days. Thanks for testing!

elemoine:
@k4nar does your app so something special with hstore? I mean does it call `register_hstore` or something to register an hstore typecaster?

elemoine:
@ewjoachim at this point I see no global state setting from aiopg/psycopg2. There must be something somewhere though :) Let's dig this one out!

k4nar:
Ok I've found the culprit, it's this line: https://github.com/psycopg/psycopg2/blob/e14e3385b4809ec4223894f8c7a009b1560eb41d/lib/extras.py#L994

The function `register_hstore` basically does two things: register a type so that hstore are properly returned as a dict-like object (instead of a string), and register an adapter so that dicts passed as argument of a query are converted as hstore. This second part is not specific to a connection, so it happen globally.

Turns out that in our project we also register an adapter for dicts, but one that convert them to JSON. When procrastinate creates its connection, our adapter is overridden & our dicts are converted to hstore instead, which causes syntax errors in our queries.

Adapters seems to be global, not bound to a connection. I don't think there is a simple workaround here (other than re-overriding the adapter in our project :smile: ).

Therefore I think we should continue to avoid registering hstore & json in Procrastinate, and manage the type conversion ourselves.

elemoine:
Good spotting @k4nar! Don't think that we should report this as a bug in aiopg? aiopg really shouldn't set psycopg2 global states.

k4nar:
If you want, but there is not much they can do. Other than re-defining `register_hstore` so that it doesn't register an adapter.

elemoine:
Don't you think that aiopg should not call register_json and register_hstore by default?

k4nar:
Probably yes, but that's a breaking change though :/ .

elemoine:
True. Dead end.

elemoine:
@k4nar @ewjoachim aiopg also [calls](https://github.com/aio-libs/aiopg/blob/7f0debf7a45626aa8ec85c8c714550734c0a8cae/aiopg/connection.py#L558) `register_uuid()`, which also globally [registers](https://github.com/psycopg/psycopg2/blob/e14e3385b4809ec4223894f8c7a009b1560eb41d/lib/extras.py#L685) an adapter. So we probably want to set `enable_uuid` to `False` as well. If you agree I'll create a PR, and possibly add tests.

ewjoachim:
Please go ahead :)

I hope we’ll have a better postgres driver soon ! 🤞

elemoine:
> I hope we’ll have a better postgres driver soon !

Couldn't we add an asyncpg connector to Procrastinate? Why was aiopg preferred over asyncpg? Just curious…

ewjoachim:
Honnestly, it's just because we had already worked with aiopg before, but I'm open to any change that makes our lives easier !

PR #134, opened 2020-01-31
Set enable_uuid to False when calling aiopg.connect

elemoine:
This PR follows up on #113 and #133.

As discussed in #133 the psycopg2.extras `register_hstore` and `register_uuid` functions call `psycopg2.extensions.register_adapter` which register adapter globally. We already set `enable_hstore` to `False` to prevent the global registration of the hstore adapter. This PR does the same for uuid. This PR also adds a test verifying that calling `get_connection` doesn't lead global adapter registrations.

### Successful PR Checklist:
- [X] Tests
- [X] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation)) **not relevant**

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -16,11 +16,12 @@ def wrap_json(arguments: Dict[str, Any]):
 
 
 def get_connection(dsn="", **kwargs) -> Awaitable[aiopg.Connection]:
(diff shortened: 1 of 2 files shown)


ewjoachim:
At this point, a review from @k4nar is probably more useful than from me :)

PR #135, opened 2020-01-31
Retry on specific exception type 84

### LINKED ISSUE #84: Retry depending on the exception
1. The retry strategy should be provided the exception that triggers the retry
2. A builtin mechanism should allow to retry only if the exception is of a given type or among a few types.

ewjoachim:
Fixes #84 

Pair programmed with @emmanuelleq 

### Successful PR Checklist:
- [x] Tests

### DIFF
--- a/docs/howto/retry.rst
+++ b/docs/howto/retry.rst
@@ -34,11 +34,20 @@ Simple strategies
 Advanced strategies
 ^^^^^^^^^^^^^^^^^^^
 
-You can get a more precise strategy using a :py:class:`RetryStrategy` instance::
+Advanced strategies let you:
+- define a maximum number of retries
+- define the retry delay, with constant, linear and exponential backoff options
+- define the exception types you want to retry on
+
+Define your precise strategy using a :py:class:`RetryStrategy` instance::
 
     from procrastinate import RetryStrategy
 
-    @app.task(retry=procrastinate.RetryStrategy(max_attempts=10, wait=5))
+    @app.task(retry=procrastinate.RetryStrategy(
+        max_attempts=10,
+        wait=5,
+        retry_exceptions=[ConnectionError, IOError]
+    ))
     def my_other_task():
         print("Hello world")
 
@@ -62,9 +71,12 @@ Implementing your own strategy
         min = 1
         max = 10
 
-        def get_schedule_in(self, attempts: int) -> int:
+        def get_schedule_in(self, exception:Exception, attempts: int, **kwargs) -> int:
             if attempts >= max_attempts:
                 return None
 
             return random.uniform(self.min, self.max)
 
+
+It's interesting to add a catch-all parameter ``**kwargs`` to make your strategy more
+resilient to possible changes of Procrastinate in the future.
--- a/docs/howto/retry.rst
+++ b/docs/howto/retry.rst
@@ -50,7 +50,7 @@ Define your precise strategy using a :py:class:`RetryStrategy` instance::
     @app.task(retry=procrastinate.RetryStrategy(
         max_attempts=10,
         wait=5,
-        retry_exceptions=[ConnectionError, IOError]
+        retry_exceptions={ConnectionError, IOError}
     ))
     def my_other_task():
         print("Hello world")
--- a/docs/howto/retry.rst
+++ b/docs/howto/retry.rst
@@ -36,12 +36,12 @@ Advanced strategies
 
 Advanced strategies let you:
 
-- define a maximum number of retries (if you don't, the job will be retried
-  indefinitely until it passes)
+- define a maximum number of retries (if you don't, jobs will be retried indefinitely
+  until they pass)
 - define the retry delay, with constant, linear and exponential backoff options (if
-  you don't, jobs will retry immediately)
-- define the exception types you want to retry on (if you don't, retry will occur on
-  any type of exceptions)
+  you don't, jobs will be retried immediately)
+- define the exception types you want to retry on (if you don't, jobs will be retried
+  on any type of exceptions)
 
 Define your precise strategy using a :py:class:`RetryStrategy` instance::
 
--- a/docs/howto/retry.rst
+++ b/docs/howto/retry.rst
@@ -75,7 +75,7 @@ Implementing your own strategy
         min = 1
         max = 10
 
-        def get_schedule_in(self, exception:Exception, attempts: int, **kwargs) -> int:
+        def get_schedule_in(self, *, exception:Exception, attempts: int, **kwargs) -> int:
             if attempts >= max_attempts:
                 return None
 
--- a/docs/howto/retry.rst
+++ b/docs/howto/retry.rst
@@ -35,9 +35,13 @@ Advanced strategies
 ^^^^^^^^^^^^^^^^^^^
 
 Advanced strategies let you:
-- define a maximum number of retries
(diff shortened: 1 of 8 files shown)


docs/howto/retry.rst
```diff
@@ -34,11 +34,20 @@ Simple strategies
 Advanced strategies
 ^^^^^^^^^^^^^^^^^^^
 
-You can get a more precise strategy using a :py:class:`RetryStrategy` instance::
+Advanced startegies let you:
+- define a maximum number of retries
+- define the retry delay, with constant, linear and exponential backoff options
+- define the exception types you want to retry on
```
pmourlanne: We should point out that defining the exception types is optional, and that if not given, we will retry on every type of exception I think :O

docs/howto/retry.rst
```diff
@@ -62,9 +71,12 @@ Implementing your own strategy
...
-        def get_schedule_in(self, attempts: int) -> int:
+        def get_schedule_in(self, exception:Exception, attempts: int, **kwargs) -> int:
             if attempts >= max_attempts:
                 return None
 
             return random.uniform(self.min, self.max)
 
+
+It's interesting to add a catch-all parameter ``**kwargs`` to make your strategy more
+resilient to possible evolutions of Procrastinate in the future.
```
pmourlanne: Even though we're changing the `*args` part of the method here :> They'll never be safe!

procrastinate/retry.py
```diff
@@ -17,15 +17,17 @@ class BaseRetryStrategy:
...
+    ) -> Optional[exceptions.JobRetry]:
+        schedule_in = self.get_schedule_in(exception=exception, attempts=attempts)
         if schedule_in is None:
             return None
 
         schedule_at = pendulum.now("UTC").add(seconds=schedule_in)
         return exceptions.JobRetry(schedule_at)
 
-    def get_schedule_in(self, attempts: int) -> Optional[float]:
+    def get_schedule_in(self, exception: Exception, attempts: int) -> Optional[float]:
```
pmourlanne: We're giving the exception here because why not? I don't mind it, just making sure I'm not missing anything

docs/howto/retry.rst
ewjoachim: Maybe adding a `*,` would be worth it, then...

procrastinate/retry.py
ewjoachim: We'll be doing calls with "exception" here, because the concrete implementation needs to look at the exception to make the call whether it needs to retry or not. Changes in the concrete implementation imply changes in the abstract one too, ohterwise:
- mypy complains
- people will implement it wrongly.

pmourlanne: Right, I missed that we actually use it in the implementation (obviously.) My bad, thanks for the explanation :+1:

ewjoachim:
@pmourlanne both problems fixed

ewjoachim:
We could merge but @elemoine said he wanted to make a review, so let's wait until Monday/Tuesday.

tests/acceptance/test_nominal.py
```diff
@@ -82,6 +82,17 @@ def test_nominal(defer, worker):
...
+    defer("multiple_exception_failures")
+    stdout, stderr = worker()
+    assert (
+        stdout
+        == """Launching a worker on all queues
+Try 0
+Try 1
+Try 2
+"""
+    )
```
k4nar: Shouldn't we also test stderr, to make sure it failed as expected?

ewjoachim: Indeed, costs nothing, and it's highly relevant.

ewjoachim: @emmanuelleq How about you take care of this fix ? I'll be available to help, if needed!

procrastinate/retry.py
```diff
@@ -17,15 +17,19 @@ class BaseRetryStrategy:
     Children classes only need to implement `get_schedule_in`.
```
elemoine: nit: s/Children/Child (while you're at it)

docs/howto/retry.rst
```diff
@@ -34,11 +34,24 @@ Simple strategies
...
+
+Define your precise strategy using a :py:class:`RetryStrategy` instance::
 
     from procrastinate import RetryStrategy
 
-    @app.task(retry=procrastinate.RetryStrategy(max_attempts=10, wait=5))
+    @app.task(retry=procrastinate.RetryStrategy(
+        max_attempts=10,
+        wait=5,
+        retry_exceptions=[ConnectionError, IOError]
```
elemoine: I'd use a `tuple` here instead of a `list`. To advertise better practices to our users.

ewjoachim: In what way is this a better practice to use a tuple here ? This is an array of homogeneous elements with arbitrary length, which fits better as a list IMHO, contrary to a tuple which works best as a fixed-length iterable where each element's type and signification is described through their position, no ?

ewjoachim: To be more precise, my rule of thumb:
- If you have very specific performance constraints, you may explore the collections zoo (deque, ...)
- If it makes no sense to have duplicates and order doesn't matter, it's a set
- If any 2 elements could be swapped and it would still make sense, it's a list
- if it can be used as basis for a namedtuple, it's a tuple (and you should consider using a namedtuple)
- Else... I'm interested to know what iterable you've found that fail all of the above tests.

elemoine: I did not go that far :-)

I just think a fixed-length, immutable, data structure makes more sense in such cases. In the same way `isinstance` receives a tuple as the second argument. Or when you write `elem in ("foo", "bar")`.

But I don't mind to be too pedantic. Using a list is fine too.

ewjoachim: Ha sorry for being pedantic too :D

I think the fact `ìsinstance()` takes a tuple and not an iterable is an error from Python's youth.

Also, I think you were totally right questioning my choice of using a list: from my own rationale just above, a set would actually be much more appropriate (order doesn't matter and duplicates make no sense) 😆

And finally, regarding `elem in ("foo", "bar")`, I was curious if the most "appropriate" structure (here, a set) was also the most performant and it kinda looks like it is (with huge question marks regarding the validity of my benchmark):
```
In [1]: %timeit "baz" in {"foo", "bar"}
29.6 ns ± 0.79 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [2]: %timeit "baz" in ("foo", "bar")
57 ns ± 0.656 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit "baz" in ["foo", "bar"]
58 ns ± 1.9 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
(this holds true if we're looking for the 2nd element but the 3 are as long if we're looking for the 1st element)
(Just to be clear, I'm not advocating for using one or another way just to earn a few nanoseconds as long as it's not on the critical path)

elemoine:
@ewjoachim feel free to merge this whenever you think this is good to go.

procrastinate/retry.py
k4nar: Isn't `children` the proper form as `classes` is plural? :thinking:

k4nar: Ah maybe not, as `child` is an adjective here :) .

ewjoachim:
Yes, just waiting for @CorBott to approve :)

PR #144, opened 2020-02-07
Adding a first version of dockerized development environment.

### LINKED ISSUE #139: Dockerize the procrastinate development environment
To be able to quickly setup a development environment, we could wrap the procrastinate demo app in its own container in the `docker-compose.yml`. This would avoid the use all the exports in the `CONTRIBUTING.rst` and maybe some of the installs of the `README.rst`.

The new image should have the necessary variable to connect to the database and use the demo app, and its entrypoint should be the procrastinate command. Of course, it should bind the volume to the procrastinate directory and use a develop install, so that changes of code can be instantly checked on the demo app.

DainDwarf:
Closes #139 

This adds a development docker environment that should let you get to it immediately.

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -16,8 +16,8 @@ TL;DR
     $ # Export libpq env vars for PG connection
     $ export PGDATABASE=procrastinate PGHOST=localhost PGUSER=postgres
 
-    $ # Launch PostgreSQL within Docker
-    $ docker-compose up -d
+    $ # Start PostgreSQL in a Docker container
+    $ docker-compose up -d postgres
 
     $ # Explore tox entrypoints
     $ tox -l
@@ -43,7 +43,7 @@ Environment variables
 The `export` command below will be necessary whenever you want to interact with
 the database (using the project locally, launching tests, ...).
 These are standard ``libpq`` environment variables, and the values used below correspond
-to the docker setup. Feel free to adjust them as necessary.
+to the Docker setup. Feel free to adjust them as necessary.
 
 .. code-block:: console
 
@@ -52,13 +52,13 @@ to the docker setup. Feel free to adjust them as necessary.
 Create your development database
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The development database can be launched using docker with a single command.
+The development database can be launched using Docker with a single command.
 The PostgreSQL database we used is a fresh standard out-of-the-box database
 on the latest stable version.
 
 .. code-block:: console
 
-    $ docker-compose up -d
+    $ docker-compose up -d postgres
 
 If you want to try out the project locally, it's useful to have ``postgresql-client``
 installed. It will give you both a PostgreSQL console (``psql``) and specialized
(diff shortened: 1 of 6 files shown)


DainDwarf:
Still TODO:
* Find a way to play migrations if needed, before starting the worker
* Add documentation!

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.7-buster
```
ewjoachim: What about just "python:3" ? (once we support 3.8)

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+WORKDIR /procrastinate_dev/
+COPY requirements.txt ./
+COPY setup.* ./
+COPY procrastinate ./procrastinate/
```
ewjoachim: maybe `COPY . .` and let dockerignore do the rest ?

.dockerignore
```diff
@@ -0,0 +1,25 @@
...
+/.coverage
+.DS_Store
+/.mypy_cache
+/.pytest_cache
+/.tox
+*.egg-info
+/htmlcov
+/dist
+/build
+/.venv.egg-info/
```
ewjoachim: Woulid it be easier to ignore `*` and then unignore things we want to keep (with `!`) ?

Dockerfile
ewjoachim: Ah, we fixed 3.8 ?
(not sure we published since)

DainDwarf: 3.8 is not fixed on pypi (yet), but as this image is used within docker-compose where we install the repository version, it works. :)

.dockerignore
DainDwarf: I am not sure what we want to ignore or keep. As it's only a development docker, we could just keep everything.

CONTRIBUTING.rst
```diff
@@ -113,11 +96,11 @@ Install local dependencies:
 Run the project automated tests
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-With a running database:
+Within docker container:
 
 .. code-block:: console
 
-    (venv) $ pytest  # Test the code with the current interpreter
+    $ docker-compose exec procrastinate pytest  # Test the code with the latest python
```
DainDwarf: with the latest python -> within a python docker

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3
+
+WORKDIR /procrastinate_dev/
+COPY . ./
```
DainDwarf: Go back to copying only requirements/setup/procrastinate, and remove dockerignore.

CONTRIBUTING.rst
```diff
@@ -13,42 +13,28 @@ TL;DR
...
-
-The `export` command below will be necessary whenever you want to interact with
-the database (using the project locally, launching tests, ...).
-These are standard ``libpq`` environment variables, and the values used below correspond
-to the docker setup. Feel free to adjust them as necessary.
-
-.. code-block:: console
-
-    $ export PGDATABASE=procrastinate PGHOST=localhost PGUSER=postgres
-
```
DainDwarf: Keep env variables for outside-of-docker testing,

.dockerignore
DainDwarf: As discussed with @elemoine , I removed the `.dockerignore` file and only copied the files necessary for the pip install in Dockerfile.

As we later mount the whole repository anyway, we can only including what is strictly necessary for install in the image.

elemoine:
Ok, testing now…

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3
+
+WORKDIR /procrastinate_dev/
+COPY requirements.txt ./
+COPY setup.* ./
+COPY procrastinate ./procrastinate/
```
elemoine: You can have just one COPY instruction I guess:

```
COPY requirements.txt setup.* procrastinate ./
```

no?

(I am not sure about the recursive copy of the procrastinate directory.)

CONTRIBUTING.rst
```diff
@@ -13,42 +13,28 @@ TL;DR
 
 .. code-block:: console
 
-    $ # Export libpq env vars for PG connection
-    $ export PGDATABASE=procrastinate PGHOST=localhost PGUSER=postgres
-
-    $ # Launch PostgreSQL within Docker
+    $ # Launch services within Docker
     $ docker-compose up -d
```
elemoine: When doing `docker-compose up -d` the first time the procrastinate worker will fail to start because the Procrastinate database objects have not been created yet. No?

CONTRIBUTING.rst
```diff
@@ -13,42 +13,28 @@ TL;DR
...
 
-    $ # Explore tox entrypoints
-    $ tox -l
+    $ # Create procrastinate databases
+    $ docker-compose run --rm procrastinate migrate
+
+    $ # Launch demo
```
elemoine: Why "Launch demo" here? I'd say "Display the procrastinate command's help message" or something…

CONTRIBUTING.rst
```diff
@@ -13,42 +13,28 @@ TL;DR
...
+    $ docker-compose run --rm procrastinate -h
 
-    $ # You can do things without tox too:
+    $ # Launch tests from the docker container
+    $ docker-compose exec procrastinate pytest
 
+    $ # Outside of docker containers:
```
elemoine: I find this a bit confusing. It sounds like the "outside of docker containers" commands are the next commands to run, while they only make sense if Docker is not used.

CONTRIBUTING.rst
```diff
@@ -60,14 +46,11 @@ on the latest stable version.
 
     $ docker-compose up -d
```
elemoine: Again this is going to start the procrastinate worker, which will fail to start because the database structure has not been created yet.

Dockerfile
DainDwarf: I'll try that, thanks!

CONTRIBUTING.rst
```diff
@@ -113,11 +96,11 @@ Install local dependencies:
...
 
-With a running database:
+Within docker container:
 
 .. code-block:: console
 
-    (venv) $ pytest  # Test the code with the current interpreter
+    $ docker-compose exec procrastinate pytest  # Test the code with the latest python
 
 Or
```
elemoine: Before, with this `Or`, we were comparing the execution of the tests to the execution of all the checks. Now this is confusing because we use docker-compose for the execution of pytest, but not for the execution of tox.

CONTRIBUTING.rst
```diff
@@ -211,21 +194,23 @@ If you need to add words to the spell checking dictionary, it's in
 Try our demo
 ------------
 
-With a running database:
-
-Launch a worker with:
+You can see the docker worker with:
```
elemoine: "You can see the docker worker **logs** with:"
No ?

Dockerfile
DainDwarf: I cannot copy the `procrastinate/` directory on the same line, but I merged the first two copies.

CONTRIBUTING.rst
DainDwarf: As you suggested on meet, I separated the commands to first launch only postgres and run the migrations, before starting the worker.

CONTRIBUTING.rst
DainDwarf: I separated the commands in here as well. :)

elemoine:
This is now ready for review again. As discussed with @DainDwarf and @ewjoachim I created in the CONTRIBUTING guide a section dedicated to working with Docker, the goal being to avoid clobbering the CONTRIBUTING guide with Docker-related stuff that developers not using Docker don't care about.

CONTRIBUTING.rst
```diff
@@ -274,6 +274,94 @@ Schedule some tasks with:
...
+
+.. code-block:: console
+
+    $ export UID
+    $ export GID=$(id -g)
+    $ docker-compose run --rm procrastinate pytest
+
+The ``UID`` and ``GID`` environment variables are set and exported for the Procrastinate
+container to be run with the current user id and group id. If not set or exported, the
+Procrastinate container will run as root, which isn't recommended.
```
ewjoachim: (I think this statement should be sourced. It's not universally recognized as true)

docker-compose.yml
```diff
@@ -1,6 +1,19 @@
 version: '2.1'
 services:
-
+  procrastinate:
+    build: .
+    volumes:
+      - "./:/procrastinate_dev/"
```
ewjoachim:
```suggestion
      - ".:/procrastinate_dev"
```

CONTRIBUTING.rst
elemoine: Actually, there's a real problem related to running the container as root: that will create in the developer's repo files and directories owned by root, which is certainly not desired. I'll change the wording.

docker-compose.yml
elemoine: Done. Thanks.

CONTRIBUTING.rst
ewjoachim: At what point does procrastinate creates a file ?

elemoine: `mypy` creates a `.mypy_cache` dir for example.

Issue #100, opened 2019-10-21
Migrations


ewjoachim:
For now, it's all fun and games, no one uses Procrastinate yet, and we haven't  really thought about migrations, but it's probably time to setup something.

- People might hop in and install procrastinate at any version (starting at 1.0, say)
- Starting from there, people will need to upgrade procrastinate (including the schema)
- Ideally, in a simple manner
- Ideally, without impact and downtime
- Optionally, using https://github.com/peopledoc/septentrion/ given we're developing this tool too.

One way of doing it could be:
- Adapt the migration files to septentrion. Remove `procrastinate migrate` in favor of `septentrion migrate`.
- Create a first migration for something trivial, just to test it.
- Document the hell out of it, so that it's easy to understand. Don't forget Septentrion is by far not as properly documented 1. as it should be and 2. as procrastinate is.
- (make it easy for people who want to read migrations instead of executing them)

(Note that in PeopleDoc, we actually don't need this part at all, because our fantastic DBA team will take care of the migrations for us, so this is strictly an effort to make procrastinate more usable by other people)

ewjoachim:
Is it worth making a procrastinate cli command that spits out a septentrion.ini file ?

ewjoachim:
Should `procrastinate` explicitely depend on `septentrion` (maybe using an extra) and `procrastinate migrate` be instead a wrapper around `septentrion migrate` that autoconfigures septentrion ?

mgu:
I like the way that is described in the description :)
(plain sql files, septentrion as an extra requirement)

ewjoachim:
@elemoine has come with a very very interesting proposal:

Given:
- Septentrion is not and has not been written to be a tool for running migrations in production
- Septentrion is very linked to PeopleDoc's way of writing migrations (including, for example, `--meta-psql:do-until-0`, etc)
- Septentrion is not as clean as Procrastinate is
- This is one of the last release blockers for 1.0
- There is an external tool (https://github.com/opengisch/pum/) that actually does what we wished Septentrion did.

Pum seems like a nice, much much more mature tool, with plenty of goodies.

Then, what about the following plan:

- We don't link Procrastinate to septentrion
- We publish up-to-date schemas, and a simple way to apply it ot an empty database
- We publish simple agnostic SQL migration files
- We advertise the use of pum in the documentation to use the migration files in real-life use cases

- Maybe, later, in a future that is not planned yet,we may self-migrate on startup using pum.
[...]

PR #145, opened 2020-02-07
Remove the procrastinate_jobs.started_at column

### LINKED ISSUE #143: Remove the started_at column from the procrastinate_jobs table
Currently the `procrastinate_jobs` table includes a `started_at` column that records the time at which jobs were started. The only reader of that column is the `select_stalled_jobs` SQL query, which, currently, is not used at all by the Procrastinate code. Even if we want to keep that SQL query, we can still remove the `started_at` column and rewrite that query with a JOIN between the `procrastinate_jobs` and `procrastinate_events` tables.

elemoine:
Closes #143 

This PR removes the unneeded `procrastinate_jobs.started_at` column. The `select_stalled_jobs`, which is the only user of that column, is rewritten in terms of a JOIN between `procrastinate_jobs` and `procrastinate_events`.

TODO with a separate PR: document how to use Pum to do database migrations.

### Successful PR Checklist:
- [x] Tests
- [ ] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation)) **not relevant** + I propose to add an "Use Pum for migrations" How-to section with a separate PR

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -208,6 +208,71 @@ not related to your PR (but please put it in a dedicated commit).
 If you need to add words to the spell checking dictionary, it's in
 ``docs/spelling_wordlist.txt``. Make sure the file is alphabetically sorted!
 
+Migrations
+----------
+
+Create database migration scripts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you make changes to the database structure (``procrastinate/sql/structure.sql``) you
+also need to create a corresponding migration script in the
+``procrastinate/sql/migrations`` directory.
+
+For example, let's say you want to add column named ``extra`` to the
+``procrastinate_jobs`` table. You would obviously edit
+``procrastinate/sql/structure.sql`` and change the definition of the table to add that
+column. But you'd also need to create a migration script, whose content would look like
+this:
+
+.. code-block:: sql
+
+    ALTER TABLE procrastinate_jobs ADD COLUMN extra TEXT;
+
+The name of migration scripts must follow a specific pattern:
+
+.. code-block::
+
+    delta_x.y.z_your_migration_script_name.sql
+
+We follow the conventions and rules of the `Pum`_ (PostgreSQL Updates Manager) project.
+
+.. _`Pum`: https://github.com/opengisch/pum/
+
+``x.y.z`` is a number associated with your migration script (e.g. ``1.0.1``).
+``your_migration_script_name`` provides a short description of your change; it is
+important to use underscores rather than hyphens between words.
+
+The series of migration numbers must be strictly increasing. So you must pick for your
+migration script a number that is higher than all the existing ones.
+
+Also, we use a semantic-versioning-like scheme, so you'd increment ``x`` (MAJOR) for a
+breaking change, ``y`` (MINOR) for a non-breaking change (e.g.  adding a column), and
+``z`` (PATCH) for a bug fix (e.g. fixing a bug in an SQL function).
+
+For example, let's say the ``migrations`` directory includes migration scripts numbered
+``1.0.0`` and ``1.0.1``, and you're fixing a bug in an SQL function. You will then use
+number ``1.0.2`` for your migration script.
+
+Finally, don't forget to update the version number in ``migration.py`` and at the top
+of ``structure.sql``.
+
+Find a given schema version
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To extract the full SQL schema at a given schema version, do:
+
+.. code-block:: console
+
+    $ VERSION="1.0.0"
+    $ git show $( \
+        git log \
+            --pretty=format:"%h" \
+            -S "Schema version $VERSION" \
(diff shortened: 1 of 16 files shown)


k4nar:
I see we don't have a Changelog yet, but this is exactly the kind of things we'll want to put in it :) . Although this isn't a breaking change as the column is nullable.

ewjoachim:
@k4nar Planning on doing a changelog in the doc that mirrors https://github.com/peopledoc/procrastinate/releases

How does that sounds/looks ?
Capture d’écran 2020-02-08 à 00 12 23

k4nar:
I always look at a file named CHANGELOG or HISTORY at the root first, but I guess using the Releases as a changelog is fine too (even more so if the changelog entries are in the tag description in git).

My only concern is that Changelog entries can easily be forgotten, and they're not reviewed at the same time than the PR. But at least there is no towncrier and/or rebase madness to confuse contributors.

ewjoachim:
We can continue over here :) https://github.com/peopledoc/procrastinate/pull/146

elemoine:
This is ready for review.

Note that I added a migration script (in `procrastinate/sql/migrations/`). This migration script is plain SQL, and it's compatible with Pum. Based on the discussion in https://github.com/peopledoc/procrastinate/issues/100, the next step is to add an How-to section to describe how to use Pum to migrate the Procrastinate database. I'll do with a separate PR.

procrastinate/sql/migrations/delta_3.0.1_drop-started-at.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE procrastinate_jobs DROP COLUMN started_at;
```
ewjoachim: You have to update the procrastinate_fetch_job function first.
Also, given we don't know what the next version number will be, maybe it would make more sense to not name the file after a version number that hasn't been choosen yet ? Maybe 00001 ?

ewjoachim:
Maybe in a different PR, but I wonder if it would make sense that for some migrations (not necessarily all), we would track the commit hashes where one could extract the schema that this migration was planned to work against.

(if Pum doesn't already has something for our use case)

procrastinate/sql/migrations/delta_3.0.1_drop-started-at.sql
elemoine: > You have to update the procrastinate_fetch_job function first.

Oops yes, good catch.

elemoine: > Also, given we don't know what the next version number will be, maybe it would make more sense to not name the file after a version number that hasn't been choosen yet ? Maybe 00001 ?

Pum requires that the names of migration files (called delta files in the Pum jargon) have this form:

```
delta_x.x.x_deltaname.sql
```

where `x.x.x` is the version number.

We can indeed number our delta files the way we want, and independently from the Python package version if we want to.

When using Pum the very first step is to create a "baseline" in the so-called Pum information table. That baseline represents the base version, from which delta files will be applied.

For example, if my baseline is 0.0.1, `pum upgrade` will apply all the delta files that are greater than or equal to 0.0.1. And all the applied delta files will be recorded in the Pum information table. The next time `pum upgrade` is executed the new delta files whose versions are greater or equal to the delta versions recorded in the Pum information table will be applied.

The hard part for users is to pick a baseline version. If the Procrastinate version I'm installing has deltas that go up to 0.0.4 then I should set 0.0.5 as my baseline. In this way I am assured that Pum will not attempt to apply deltas that bring migrations that I already have in the database.

For this to work our responsibility as Procrastinate developers will be to have a series of deltas that is strictly increasing.

elemoine:
> Maybe in a different PR, but I wonder if it would make sense that for some migrations (not necessarily all), we would track the commit hashes where one could extract the schema that this migration was planned to work against.

I am not sure how we could set up something like that.
> 
> (if Pum doesn't already has something for our use case)

See my comment above to understand how Pum works.

ewjoachim:
If our version numbers are distinct, then we could go with semver-like:
- Major for breaking changes
- Minor for non-breaking changes (adding fields, etc)
- Incremental for bugfixes (like fixing bugs in the PLPGSQL procedures)

Our first version could be 1.0.0 ?

ewjoachim:
Also, the "current baseline" should definitely be written in the schema.

elemoine:
> Our first version could be 1.0.0 ?

Sounds good.

ewjoachim:
I think you'll need to fix https://github.com/peopledoc/procrastinate/blob/master/procrastinate/migration.py#L9 too :)

elemoine:
> I guess the next step is to remove our migration table in favor of Pum's one, but that can be done in another PR.

Yep! I plan to get rid of the Migrator class completely, and I think we should remove the command from `migrate` to `createdb` as this is just about the initial creation of the schema.

ewjoachim:
> I plan to get rid of the Migrator class completely

Well, we'll still need somewhere to put the code for creating the DB

PR #146, opened 2020-02-08
Automated release process

### LINKED ISSUE #98: Automate release
What I'd really like is:
- PR merged should automatically add a line in a draft release in GitHub
- Tags created in GitHub should trigger the corresponding version to be released on pypi

(in this schema, both changelog and version releasing are expected not to require any code change)

ewjoachim:
Closes #98 

- Adds an automatic changelog in the doc, built from the releases page when a tag is pushed
- Adds an automatic version number from the tag
- Adds an automatic release on pypi when a tag is pushed
- Document that process

Obviously the release part hasn't been tested, but isn't that exciting :)
The Changelog part has been though.
We'll need to setup 2 envvars named `CHANGELOG_GITHUB_TOKEN` and `PYPI_TOKEN` in the CI.
First one is just for rate limiting.  

### Successful PR Checklist:
- [ ] Tests - not relevant
    - Yes, except when I was writing some docutils. This was hell. Honestly, if there's a secret way to not dies on the inside anytime I have to do the simplest thing using docutils, please tell me, I wanna know.


ewjoachim:
That's a bummer! By default, the API doesn't sort the releases properly. We can sort them on our side. We could also do a much better request with graphql, but that requires to be authenticated, and even if it's going to work on the CI, it will be much harder to generate the doc locally. Too bad for graphql.

For the record, that would have let us get the HTML version of the release body, which would have removed the need for m2r.

Here's the graphql query:
```graphql
{
  repository(owner: "peopledoc", name: "procrastinate") {
    releases(orderBy: {field: CREATED_AT, direction: DESC}, first:100) {
      nodes {
        name, descriptionHTML, url, tagName
      }
    }
  }
}
```

ewjoachim:
Ok, I've changed my mind. Graphql.
Also, improve the wording & content for the "package" parameter to, e.g.:

`pypi_package: <url to pypi>`

ewjoachim:
@k4nar so to answer your questions:

- There is a top level CHANGELOG.rst file, and I've made it so that it contains a link to the releases page when not built (when built, it contains the changelog). This way, your first idea of "checking the CHANGELOG root file" will get you the info you need.

Regarding the fact changelog entry can be forgotten: it's possible, but my idea is that:
1. Release is done by publishing a GitHub release so we won't forget to write a changelog
2. In an ideal world, publishing a PR would write a blurb in the current release draft so that we don't forget something in the changelog. That part is still left to automate though.

docs/sphinxext/changelog.py
```diff
@@ -0,0 +1,152 @@
...
+        query = """
+        query {
+            repository(owner: "%(owner)s", name: "%(repo)s") {
+                releases(orderBy: {field: CREATED_AT, direction: DESC}, first:100) {
+                    nodes {
+                        name, descriptionHTML, url, tagName
+                    }
+                }
+            }
+        }
```
ewjoachim: Procrastinate, a project so cool it even has graphql.

setup.py
```diff
@@ -1,3 +1,19 @@
...
+
+def get_version():
+    try:
+        return (
+            subprocess.check_output(["git", "describe"])
+            .decode("utf-8")
+            .strip()
+            .replace("-", "+", 1)
+            .replace("-", ".")
+        )
```
ewjoachim: This should ensure we have a PEP 440 valid version number

ewjoachim:
(Env vars setup in the CI and in ReadTheDocs, ready for review)

docs/sphinxext/changelog.py
```diff
@@ -0,0 +1,152 @@
...
+    def run(self) -> Iterable[nodes.Node]:
+        # sett = self.state.document.settings
+        # env = self.state.document.settings.env
+
+        # gives you access to the parameter stored
+        # in the main configuration file (conf.py)
+        # config = env.config
+
+        # # we create a title and we add it to section
+        # section += nodes.title("bla")
```
k4nar: should all this code stay commented? :)

docs/sphinxext/changelog.py
```diff
@@ -0,0 +1,152 @@
...
+        url_is_correct = stripped_url.startswith(prefix) and stripped_url.endswith(
+            postfix
+        )
+        if not url_is_correct:
+            raise self.error(
+                "Changelog needs a Github releases URL "
+                f"(https://github.com/:owner/:repo/releases). Received {url}"
+            )
+
+        return stripped_url[len(prefix) : -len(postfix)]  # noqa
```
k4nar: why the `noqa` ?

setup.py
k4nar: Wait, doesn't this mean that we can't install the project with a meaningful version from pip without using the wheel?

docs/sphinxext/changelog.py
ewjoachim: https://github.com/psf/black#slices but you're right, it's not the correct fix for that.

docs/sphinxext/changelog.py
ewjoachim: hm, nope

setup.py
ewjoachim: Oh, it had completely slipped my mind that this code is executed at install time for sdists :o you're right!

Let me give it a thought, I'll make a new proposal.

ewjoachim: @k4nar I've added a new commit for that.

I've added a dynamic VERSION.txt file at release time, which gets removed right after release. Being present in the sdist (and incidentally in the whell too, but too meh), it's read at install time.

I'm not a huge fan, ...
... but my life as a wind turbine would probably have been very different. ;)

k4nar: Works for me :) .

PR #147, opened 2020-02-08
Composition postgres 142

### LINKED ISSUE #142: Composition instead of inheritance for PostgresJobStore
We're dealing with fine software design issues, it's really no big deal if we don't go this way.

I think we could make a last (haha, last...) modification on how our job store handles being plugged with a real `Postgres` database versus being in memory for tests and such . Today, it's done via inheritance, with a `BaseJobStore`, a `PostgresJobStore` and an `ImMemoryJobStore`. I've finally realized that it would be much cleaner in terms of separation of concerns if the whole idea of store was uncorrelated from the idea of DB connection. We could have a `JobStore`, and a `PostgresConnector` and an `InMemoryConnector`, and the `JobStore` would receive an `...Connector` instance at creation time (same way as the App currently receives a Job Store instance).

That would mean the App would be given a connector instead of a `JobStore`. If we change the name, that would be backwards incompatible, but we could still keep the name `PostgresJobStore` as an alias of `PostgresConnector` for at least a few releases.

ewjoachim:
Closes #142 

Includes breaking changes (but I kept a compat layer and added a deprecation warning)

### Successful PR Checklist:
- [x] Tests
- [x] Had a good time contributing? (if not, feel free to give some feedback) < **Not as bad as expected :D**


procrastinate/aiopg_connector.py
```diff
@@ -131,3 +131,12 @@ def interrupt_wait(self):
...
 
+
+class PostgresJobStore(PostgresConnector):
+    def __init__(self, *args, **kwargs):
+        message = (
+            "Use procrastinate.PostgresConnector(...) "
+            "instead of procrastinate.PostgresJobStore(...), with the same arguments"
+        )
+        logger.warn(f"Deprecation Warning: {message}")
+        warnings.warn(DeprecationWarning(message))
```
ewjoachim: super()

procrastinate/store.py
```diff
@@ -13,37 +11,22 @@ def get_channel_for_queues(queues: Optional[Iterable[str]] = None) -> Iterable[s
         return ["procrastinate_queue#" + queue for queue in queues]
 
 
-class BaseJobStore:
+class JobStore:
     json_dumps: Optional[Callable] = None
     json_loads: Optional[Callable] = None
```
elemoine: To be removed.

procrastinate/app.py
```diff
@@ -209,11 +214,15 @@ def perform_import_paths(self):
 
     @property
     def migrator(self) -> migration.Migrator:
-        return migration.Migrator(job_store=self.job_store)
+        return migration.Migrator(connector=self.connector)
+
+    @property
+    def job_store(self) -> store.JobStore:
+        return store.JobStore(connector=self.connector)
```
ewjoachim: instantiate in constructor

tests/integration/test_pg_connector.py
```diff
@@ -8,42 +8,45 @@
 pytestmark = pytest.mark.asyncio
 
 
-async def test_get_connection(connection):
-    dsn = connection.dsn
-    async with await aiopg_connector.get_connection(dsn=dsn) as new_connection:
+@pytest.fixture
```
ewjoachim: rename file as test_aiopg_connector

ewjoachim:
@elemoine @k4nar Shall I fixup my fixups ?

k4nar:
You can squash them if you want.

ewjoachim:
LGTY ?

docs/howto/testing.rst
```diff
@@ -1,20 +1,20 @@
 Test your code that uses Procrastinate
 --------------------------------------
 
-Procrastinate defines an `InMemoryJobStore` that will speed up your tests,
-remove dependency to PostgreSQL and allow you to have tasks run in a
+Procrastinate defines an `InMemoryConnector` that will speed up your tests,
+remove dependency to PostgreSQL and allows you to have tasks run in a
```
elemoine: s/allows/allow, because of `will` earlier in the sentence.

procrastinate/aiopg_connector.py
```diff
@@ -125,60 +44,100 @@ def __init__(
...
+    def interrupt_wait(self):
+        if not self._connection:
+            return
+        asyncio.get_event_loop().call_soon_threadsafe(
+            self._connection.notifies.put_nowait, "s"
         )
 
-    def stop(self):
-        if self._connection:
-            interrupt_wait(connection=self._connection)
```
elemoine: Did you purposefully remove `stop` here?

procrastinate/app.py
```diff
@@ -51,13 +57,29 @@ def __init__(
             A :py:func:`App.task` that has a custom "name" parameter, that is not
             imported and whose module path is not in this list will
             fail to run.
+        job_store:
+            **Deprecated**: Old name of ``connector``
```
elemoine: Final dot for consistency with the other docstrings.

procrastinate/aiopg_connector.py
ewjoachim: It has moved to the store: https://github.com/peopledoc/procrastinate/pull/147/files/9683ef2955cf5ba9f0128c4220f200c347007167#diff-32d29d678d9b442460d4b55326ed2104R23

procrastinate/app.py
```diff
@@ -51,13 +57,29 @@ def __init__(
...
+        if job_store:
+            message = (
+                "Use App(connector=procrastinate.PostgresConnector(...)) "
+                "instead of App(job_store=procrastinate.PostgresJobStore())"
+            )
+            logger.warn(f"Deprecation Warning: {message}")
+            warnings.warn(DeprecationWarning(message))
+            connector = job_store
+        if not connector:
+            raise TypeError("App() missing 1 required argument: 'connector'")
```
elemoine: Do we want to raise an Exception when connector and job_store are both set in the constructor arguments?

tests/integration/test_pg_connector.py
elemoine: do you plan to do it with a separate PR?

procrastinate/app.py
ewjoachim: Do you think people could end up being confused to the point they submit those 2 parameters and are not happy with the result ?

procrastinate/aiopg_connector.py
```diff
@@ -125,60 +44,100 @@ def __init__(
...
+            )
+        except asyncio.TimeoutError:
+            pass
 
-    async def wait_for_jobs(self):
-        return await wait_for_jobs(
-            connection=await self.get_connection(), socket_timeout=self.socket_timeout
+    # This can be called from a signal handler, better not do async stuff
+    def interrupt_wait(self):
+        if not self._connection:
```
elemoine: I think we need to have `if not self._connection or self._connection.closed` here.

procrastinate/app.py
elemoine: Not sure. Up to you really.

tests/integration/test_pg_connector.py
ewjoachim: Damn, you're right

procrastinate/app.py
ewjoachim: I vote for keeping the code simple and avoiding having one more test to write :)

procrastinate/aiopg_connector.py
elemoine: In that case there shouldn't be a `stop` method in the `InMemoryConnector` either, no?

PR #150, opened 2020-02-10
WIP Using an Aiopg Pool 148

### LINKED ISSUE #148: Procrastinate should be using a pool
It just occured to me that we're not using an aiopg pool, but a connection directly. In the defer side, there may be hundreds of concurrent coroutines that try to defer at the same time. As of today, I think they will crash. We need a pool instead (which could very well be of size 1 if people don't want to create too many connections). Having a pool means aiopg manages the "semaphore" for us and we don't have to deal with it.

Who's in for a pool party ?
@elemoine / @k4nar: I believe this will be blocking for you quickly, so let's address that!

Also, we need a parameter to configure the pool size.
Finally, I'd advise against doing anything not based on top of #147.
(note: I have the beginning of a PR, but nothing that cannot be redone)

ewjoachim:
Closes #148 

Based on #147 

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -1,106 +1,25 @@
 import asyncio
+import logging
+import warnings
 from typing import Any, Callable, Dict, List, Optional
 
 import aiopg
 import psycopg2.sql
 from psycopg2.extras import Json, RealDictCursor
 
-from procrastinate import store
+from procrastinate import connector
+
+logger = logging.getLogger(__name__)
 
 
-def wrap_json(arguments: Dict[str, Any], json_dumps: Optional[Callable]):
-    return {
-        key: Json(value, dumps=json_dumps) if isinstance(value, dict) else value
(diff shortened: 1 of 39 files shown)


procrastinate/aiopg_connector.py
```diff
@@ -122,20 +123,23 @@ def make_dynamic_query(self, query: str, **identifiers: str) -> str:
...
-        pool = await self._get_pool()
-        try:
-            await asyncio.wait_for(pool.notifies.get(), timeout=self.socket_timeout)
-        except asyncio.TimeoutError:
-            pass
+    async def listen_notify(
+        self, event: asyncio.Event, channels: Iterable[str]
+    ) -> NoReturn:
+
+        async with (await self._get_pool()).acquire() as connection:
```
ewjoachim: (we'll need to make sure maxsize is at least 2)

ewjoachim:
CF #166

PR #156, opened 2020-02-15
Acquire lock for connecting and executing queries

elemoine:
Cf. #148 

### Successful PR Checklist:
- [x] Tests
- [ ] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation))  **Not relevant**

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -47,6 +47,7 @@ class PostgresConnector(connector.BaseConnector):
         kwargs["dsn"] = dsn
         self._connection_parameters = kwargs
         self._connection: Optional[aiopg.Connection] = None
+        self._lock = asyncio.Lock()
         self.socket_timeout = socket_timeout
         self.json_dumps = json_dumps
         self.json_loads = json_loads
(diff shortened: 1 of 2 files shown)


elemoine:
Thanks @ewjoachim!
I'm not sure about this change really. I think relying on a connection pool (at least on the defer side) makes for a better solution. To handle concurrent accesses, but also for robustness - open new connections in case of network issues.

elemoine:
Note to self: the locking could be more fine-grained: with one lock for the connection (`lock_connection`), and one lock for the queries (`lock_query`).

ewjoachim:
Yeah, I'm all for a pool too, but until we have a pool PR ready, I'm guessing it doesn't hurt.

(In my humblest opinion, code itself has very little value, but thought process and engineering has a lot of value. I'm ok throwing away some code, even lots of it, or going back and forth, as long a we continue to learn how to make things better and improve ourselves and the project. I have no problem merging this today and undoing it for a better pool-based solution in a week)

elemoine:
Before merging I'd like to use two different locks, because it's really two separate things that we need to lock here. I'll also add a test for the "connection locking" part.

ewjoachim:
I'm not really sure what you mean ?
Can you be more specific on the usecase of your 2 locks ?

elemoine:
> I'm not really sure what you mean ?
> Can you be more specific on the usecase of your 2 locks ?

Look at my changes.

The lock is taken in `_get_connection`. This lock is to avoid the situation where two (or more) coroutines would simultaneously create a connection, which would lead to the creation of multiple connections. This is the part for which I'd like to add a test. The lock is also taken in `execute_query` and friends. This one is to avoid creating multiple cursors on the same connection, which would result in `ResoureWarning: You can only have one cursor per connection` errors. The `test_execute_query_simultaneous` test that I've already added covers that case.

Hope that clarifies.

k4nar:
I'm unsure of the advantage of using two separate locks though. You cannot use the second one until the first one has been used successfully.

procrastinate/aiopg_connector.py
```diff
@@ -66,45 +67,49 @@ def _wrap_json(self, arguments: Dict[str, Any]):
...
-        kwargs.setdefault("enable_hstore", False)
-        kwargs.setdefault("enable_uuid", False)
-        conn = await aiopg.connect(**kwargs)
-        if self.json_loads:
-            psycopg2.extras.register_default_jsonb(conn.raw, loads=self.json_loads)
-        self._connection = conn
-        return conn
+        async with self.lock:
+            if self._connection and not self._connection.closed:
+                return self._connection
```
k4nar: This is vital that this part is inside the lock, but I wonder if we could add a fast path by copying this just before taking the lock.

The idea is that once the connection is created (and not closed), it's useless to wait on the lock. It's not even two possible to have two coroutine inside the `if` at the same time as there is no async op. So instead of waiting + taking the lock every time we get the connection, we could return immediately.

elemoine:
> I'm unsure of the advantage of using two separate locks though. You cannot use the second one until the first one has been used successfully.

coroutine 2 can get a connection while coroutine 1 is querying. I agree, there's no performance benefit here, as you need the connection for querying. Using two locks just makes it clearer in the code that we have two separate things to lock.

procrastinate/aiopg_connector.py
elemoine: :+1:

elemoine: Actually it does work. Here's what happens:

- coroutine 1 acquires the lock, and calls `aiopg.connect`
- coroutine 2 doesn't take the fast path because the connection is not yet open, so it blocks on the lock aquisition
- when coroutine 1 releases the lock, coroutine 2 acquires it and calls `aiopg.connect` **BUG**

ewjoachim:
I see what you mean.

If we switch to a pool, and we find a way to create a pool without awaiting (which seems to be possible by instantiating one, at the risk of not respcting `minsize`) we could just save the trouble and instantiate the pool in the constructor, which would mean no need for a specific lock.

(otherwise, what you're describing might happen with pools too, calling create_pool() twice)

elemoine:
Based on the review comments I decided to use one lock only. I also added a `test_get_connection_simultaneous` test.

This is ready for final review.

procrastinate/aiopg_connector.py
k4nar: That's why I said "copying this" :) . We need the `if …: return …` both before & inside the lock.

elemoine: Oh ok, I had missed that part from your comment. Sorry. Sounds like an unnecessary optimization to me, but why not…

elemoine: I added the fast path.

Issue #157, opened 2020-02-16
Getting "unclosed connection" errors when running the demo


elemoine:
We get "unclosed connection" errors when running `python -m procrastinate_demo` to schedule tasks.

```
$ python -m procrastinate_demo 
DEBUG:asyncio:Using selector: EpollSelector
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
INFO:procrastinate.jobs:Deferring job
/home/elemoine/.virtualenvs/procrastinate/lib/python3.8/site-packages/aiopg/connection.py:513: ResourceWarning: Unclosed connection <aiopg.connection::Connection isexecuting=False, closed=True, echo=False, cursor=<aiopg.cursor::Cursor name=None, closed=True>>
ERROR:asyncio:Unclosed connection
connection: <aiopg.connection::Connection isexecuting=False, closed=True, echo=False, cursor=<aiopg.cursor::Cursor name=None, closed=True>>
```

ewjoachim:
That line should take care of that, though... 
https://github.com/peopledoc/procrastinate/blob/master/procrastinate/cli.py#L109

elemoine:
The cli is not used when we do `python -m procrastimate_demo` as indicated in the contributing file.

ewjoachim:
Then maybe we should either:
- add "try: finally:" blocks in app's worker (and maybe migrator and health checker)
- Add a decorator that does this

I feel that when interacting with the app, the app's responsible for requesting the connection to open, so it should be also the one that closes it.

(maybe transparently opening the connection was not a great idea)

Point being: I'm not sure relying on the GC to close the connection is my favorite approach here :/

elemoine:
> Then maybe we should either:
> 
>    * add "try: finally:" blocks in app's worker (and maybe migrator and health checker)
>    * Add a decorator that does this

I don't understand that part. This issue is about the defer side, so I don't understand why you're talking about the worker.

> I feel that when interacting with the app, the app's responsible for requesting the connection to open, so it should be also the one that closes it.

What does `app` refer to in the above statement? The `procrastinate.App` object, or the user application? I am not sure…

[...]

PR #158, opened 2020-02-16
Call close_connection from procrastinate_demo main

elemoine:
Cf. #157 

Note: `aiopg.Connection.close` is a regular function in aiopg 0.5+. See https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Connection.close.

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -53,9 +53,8 @@ class PostgresConnector(connector.BaseConnector):
         self.json_loads = json_loads
 
     async def close_connection(self) -> None:
-        if not self._connection:
+        if not self._connection or self._connection.closed:
             return
-
         await self._connection.close()
 
     def _wrap_json(self, arguments: Dict[str, Any]):
--- a/procrastinate_demo/__main__.py
+++ b/procrastinate_demo/__main__.py
@@ -1,6 +1,7 @@
 import logging
 
 from procrastinate_demo import tasks
+from procrastinate_demo.app import app
 
 
 def main():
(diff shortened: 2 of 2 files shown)


procrastinate/connector.py
```diff
@@ -7,7 +7,7 @@ class BaseConnector:
     json_dumps: Optional[Callable] = None
     json_loads: Optional[Callable] = None
 
-    async def close_connection(self) -> None:
+    def close_connection(self) -> None:
```
ewjoachim: Please don't make close a sync function in the abstract class. Asyncpg's close as well as Aiopg's pool close are async, and making it sync will make things quite harder I believe.

procrastinate/app.py
```diff
@@ -237,5 +237,8 @@ def migrator(self) -> migration.Migrator:
     def health_check_runner(self) -> healthchecks.HealthCheckRunner:
         return healthchecks.HealthCheckRunner(connector=self.connector)
 
-    async def close_connection_async(self):
-        await self.connector.close_connection()
+    def close_connection(self):
+        self.connector.close_connection()
+
+    def __del__(self):
+        self.close_connection()
```
ewjoachim: https://github.com/peopledoc/procrastinate/blob/master/procrastinate/cli.py#L109 should be taking care of that, I'd like to understand why it doesn't.

elemoine:
Ok, with the approval from @ewjoachim, I am merging this as it does fix a problem in the procrastinate demo. But we need to revisit this some time, because it does not feel right that the user needs to close connections that she has not opened herself.

PR #159, opened 2020-02-19
Small second pass on logs

ewjoachim:
Cf. #102 

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -34,11 +34,17 @@ class Worker:
             self.logger = logger
 
     async def run(self) -> None:
-        await self.job_store.listen_for_jobs(queues=self.queues)
+        queues = self.queues
+        await self.job_store.listen_for_jobs(queues=queues)
 
-        self.logger.debug(
-            f"Log start one queues {self.queues}",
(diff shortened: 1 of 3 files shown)


ewjoachim:
In wonder if each log should be tested...

mgu:
> In wonder if each log should be tested...

I wonder why no linters have seen the issues

procrastinate/worker.py
```diff
@@ -34,11 +34,17 @@ def __init__(
...
-            f"Log start one queues {self.queues}",
-            extra={"action": "start_log_on_queue", "queues": str(self.queues)},
+        if queues:
+            queue_names = f"queues {', '.join(queues)}"
+        else:
+            queue_names = "all queues"
+
+        self.logger.info(
+            f"Starting worker on queues {queue_names}",
+            extra={"action": "start_worker", "queues": queues},
```
elemoine: You really want to use `queues` rather than `queue_names` here in the extra's?

ewjoachim: yes, for me the structured log is supposed to be structured, no ?

elemoine: Ok I see.

PR #162, opened 2020-02-21
Derive schema version from migration scripts

ewjoachim:
With this PR the schema version is derived from the versions found in migration scripts.

### Successful PR Checklist:
- [x] Tests


procrastinate/schema.py
```diff
@@ -16,6 +21,17 @@ def __init__(self, connector: connector_module.BaseConnector):
...
 
+    @property
+    def version(self):
+        version_ = pkg_resources.parse_version("1.0.0")
+        for script in migrations_dir.glob("*.sql"):
+            m = migration_script_pattern.match(script.name)
+            if m:
+                v = pkg_resources.parse_version(m.group(1))
+                if v > version_:
+                    version_ = v
```
ewjoachim: I think I know you enough so you'll pardon me for being teasing but...
You know about [max](https://docs.python.org/3/library/functions.html#max) ?
```python
matches = [migration_script_pattern.match(script.name) for script in migrations_dir.glob("*.sql")]
versions = [match.group(1) for match in matches if match}
return max(versions, key=pkg_resources.parse_version)
```

tests/integration/test_cli.py
```diff
@@ -63,7 +63,7 @@ def test_healthchecks(entrypoint, click_app, mocker):
     check_version = mocker.patch(
         "procrastinate.healthchecks.HealthCheckRunner.get_schema_version"
     )
-    check_version.return_value = SchemaManager.version
+    check_version.return_value = SchemaManager(click_app.connector).version
```
ewjoachim: why the `()` ?

procrastinate/schema.py
```diff
@@ -16,6 +21,17 @@ def __init__(self, connector: connector_module.BaseConnector):
     def get_schema() -> str:
         return read_text("procrastinate.sql", "schema.sql")
 
+    @property
+    def version(self):
```
ewjoachim: Seems to me like a staticmethod `get_version()` would make your life easier in a few places.
You can even implement this property as calling the staticmethod if you want, but you don't even have to.

procrastinate/schema.py
elemoine: Creating two lists just and doing three passes over the data, just to use `max`!!!

procrastinate/schema.py
elemoine: Yep, I was actually considering doing it.

procrastinate/schema.py
elemoine: I can do it if you prefer, but I would find it ugly.

ewjoachim: 0 lists now :)
```python
matches = (migration_script_pattern.match(script.name) for script in migrations_dir.glob("*.sql"))
versions = (match.group(1) for match in matches if match)
return max(versions, key=pkg_resources.parse_version)
```
For the rest:
- Passes over the data are irrelevant, as far as I know (please prove me wrong)
- Also, max is ~40% faster than iterating manually (because it does so in C)
- Collectively, `pkg_resources.parse_version`, `migrations_dir.glob` and `migration_script_pattern.match` take more time that we could hope to win by optimizing the loop 

I guess it's not really about performance, but about readability.
I find reimplementing max not very pythonic, and I tend to prefer "functional" (I'll dare say "vectorized") ways of writing code (acting on whole iterables rather than on individual elements).

That being said, I don't want to impose my views, I wouldnt want you to just give up just because I said so, if you find this ugly, I'm really interested in knowing why. Maybe we can find a more pythonic version that is pleasing to both our eyes :)

elemoine: You've actually convinced me! See my last commit.

PR #163, opened 2020-02-21
Remove the procrastinate_version table

elemoine:
### Successful PR Checklist:
- [x] Tests

### DIFF
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -10,7 +10,6 @@ import pendulum
 
 import procrastinate
 from procrastinate import connector, exceptions, jobs, types
-from procrastinate.schema import SchemaManager
 
 logger = logging.getLogger(__name__)
 
@@ -278,18 +277,6 @@ def healthchecks(app: procrastinate.App):
         click.echo("Cannot connect to DB")
         return  # No need to go further
     click.echo("DB connection: OK")
-
-    db_schema_version = health_check.get_schema_version()  # type: ignore
-    file_schema_version = SchemaManager.get_version()
-    schema_ok = db_schema_version == file_schema_version
-    if schema_ok:
-        click.echo(f"DB schema is up-to-date ({db_schema_version})")
-    else:
-        click.echo(
-            "There are schema migrations to apply! "
-            f"{db_schema_version} => {file_schema_version}"
-        )
-        return  # No need to go further
 
     status_count = health_check.get_status_count()  # type: ignore
     for status, count in status_count.items():
--- a/procrastinate/healthchecks.py
+++ b/procrastinate/healthchecks.py
@@ -15,12 +15,6 @@ class HealthCheckRunner:
         )
         return result["check"]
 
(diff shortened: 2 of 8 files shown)


procrastinate/sql/migrations/delta_1.2.0_drop_version.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS procrastinate_version;
```
ewjoachim: Isn't that a breaking change according to our "semver-like" plan ? I think we can do 2.0.0.

(version numbers are cheap :) )

elemoine: You mean it's a breaking change because migrating the database would require updating the Procrastinate code?

ewjoachim: I guess so, but maybe this needs a better definition ? What should be considered breaking ?

elemoine:
> PUM stands for Pum stands for "Postgres Upgrades Manager". (source: https://github.com/opengisch/pum/blob/master/README.md) and not Updates Manager.

Where is "Updates Manager" mentioned instead of "PostgreSQL Updates Manager"? I am failing to understand where the problem is. Thanks!

ewjoachim:
@elemoine https://github.com/peopledoc/procrastinate/pull/163/files#diff-37aeb18b690f8baaabb37a9f0e4cbedfR17

elemoine:
@CorBott @ewjoachim understood and fixed. Thanks :)

procrastinate/sql/migrations/delta_1.2.0_drop_version.sql
elemoine: That's a good question. Here's my take on it:

I think what's important is the perspective of Procrastinate users. And I think a Procrastinate user would never attempt to migrate the Procrastinate database schema independently from upgrading Procrastinate (the Python package). What users want is upgrade to a more recent version of Procrastinate, which may imply migrating the Procrastinate database schema.

As I see it, a PATCH release of Procrastinate must not imply a data schema migration. A MINOR release of Procrastinate may imply a database schema migration, but the migration shouldn't be mandatory, meaning that Procrastinate would still work as expected if the user didn't migrate the database schema. And a MAJOR release of Procrastinate does imply a database migration, with the risk that Procrastinate won't function properly if the user didn't not apply the migration.

With this definition, the removal of a table (as done in this PR) is not a breaking change.

ewjoachim: Are you're talking about the procrastinate version or the schema version ?

elemoine: I am talking about the Procrastinate version. I think the Procrastinate version is what users see and care about the most.

elemoine:
I finally decided to close this in favor of #167. The latter includes the changes that this PR had.

Issue #164, opened 2020-02-23
Fix doc build and deploy


ewjoachim:
- https://readthedocs.org/projects/procrastinate/builds/
- https://travis-ci.org/peopledoc/procrastinate/builds/651658412

ewjoachim:
Doc was likely broken by the SSO on the peopledoc org. Should be fixed now (I've granted the token SSO rights)
(note: the token has no rights on private repos, but with SSO, even read-only on public repos need SSO approval)

ewjoachim:
Closed by #165 (until next version)

PR #165, opened 2020-02-23
Fix builds 164

ewjoachim:
Cf. #164 

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,8 @@ after_success:
 deploy:
   provider: pypi
   user: __token__
-  password: $PYPI_TOKEN
+  password:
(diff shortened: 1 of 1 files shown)


ewjoachim:
Absolutely!

In #98 / #146 , I've changed the release process to make Travis release and deploy on pypi for us. In order to do so, I've used the [Travis PyPI deploy](https://docs.travis-ci.com/user/deployment/pypi/) feature, but I didn't read it well enough: I tried to use a "standard" env var, but travis dosen't support env var in deploy passwords.

Because of that, the automated deployment of 0.5.0 [failed](https://travis-ci.org/peopledoc/procrastinate/builds/651658412). 

Instead, the doc linked above explains explicitely what needs to be done: use the travis CLI tool to encrypt our token with Travis's public key, and use `secure:` to let travis know that they need to decrypt the token.

Another thing I noticed in the failed build is that Travis tries to deploy after each matrix element. Instead, we need to run the whole matrix and, if it worked (and if we're on master with a tag) then deploy. This can be achived with the use of build stages as described in [this example](https://docs.travis-ci.com/user/build-stages/deploy-rubygems/)

PR #169, opened 2020-03-06
Add a "Set database schema" how-to

elemoine:
This adds an how-to section describing how to make Procrastinate use a specific database schema rather the `public` default schema.

### Successful PR Checklist:
- [ ] Tests **Not relevant**

### DIFF
--- a/docs/howto/set_database_schema.rst
+++ b/docs/howto/set_database_schema.rst
@@ -1,0 +1,26 @@
+Set the database schema
+-----------------------
+
+Any PostgreSQL database contains one or more *schemas*. Schemas are PostgreSQL way of
+implementing namespaces for database objects. See the `PostgreSQL Schema documentation`_
+for more detail on schemas. (See the :ref:`glossary <glossary_schema>` for a note on the
+overloading of the term *schema*.)
+
+By default Procrastinate uses the ``public`` schema, which is PostgreSQL's default
+schema, i.e. the schema that every new database contains.
+
+To have Procrastinate use another schema than ``public``, change the *schema search
+path* when creating the :py:class:`procrastinate.PostgresConnector`::
+
+    app = procrastinate.App(
+        connector=procrastinate.PostgresConnector(
+            host="localhost",
+            options="-c search_path=myschema"
+        )
+    )
+
+With this the ``procrastinate schema --apply`` command will create the Procrastinate
+database objects (tables, functions, etc.) into the ``myschema`` schema. And all the
+SQL queries issued by Procrastinate will apply to the database objects of that schema.
+
+.. _PostgreSQL Schema documentation: https://www.postgresql.org/docs/current/ddl-schemas.html
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -62,3 +62,10 @@ Schema
     database. Applying the schema means installing all those objects in the database.
     A evolution in the schema (modifying the table structure, or the procedures) is
     called a migration.
+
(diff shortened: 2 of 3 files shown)


docs/howto/set_database_schema.rst
```diff
@@ -0,0 +1,25 @@
+Set the database schema
+-----------------------
+
+Any PostgreSQL database contains one or more *schemas*. Schemas are namespaces really,
+not much more than that. See the `PostgreSQL Schema documentation`_ for more detail on
```
ewjoachim: This tone is a little bit too harsh. 

> Schemas are Postgres way of implementing namespaces.

Also, you'll need to update the glossary, we now have 2 clashing definitions for what a "schema" is :)

docs/howto/set_database_schema.rst
```diff
@@ -0,0 +1,25 @@
...
+schemas.
+
+By default Procrastinate uses the ``public`` schema, which is PostgreSQL's default
```
ewjoachim: > By default Procrastinate uses the ``public`` schema, which is PostgreSQL's default
schema

A bit tautological :)

docs/howto/set_database_schema.rst
```diff
@@ -0,0 +1,25 @@
...
+schema, that is the schema that every new database contains.
+
+To make Procrastinate use a specific schema, use the ``-c search_path=`` option at
+``PostgresConnector`` creation time::
```
ewjoachim:
```suggestion
To have Procrastinate use a specific schema, pass it when creating the :py:class:`procrastinate.PostgresConnector`::
```

docs/howto/set_database_schema.rst
```diff
@@ -0,0 +1,25 @@
...
+    app = procrastinate.App(
+        connector=procrastinate.PostgresConnector(
+            host="localhost",
+            options="-c search_path=myschema"
+        )
+    )
+
+With this the ``procrastinate schema --apply`` command will create the Procrastinate
+database objects (tables, functions, etc.) into the ``myschema`` schema. And all the
+SQL queries issued by Procrastinate will apply to the database objects of that schema.
```
ewjoachim: > If the connector receives an existing Pool, there is currently no way to have Procrastinate use a different schema from the rest of your application.

docs/howto/set_database_schema.rst
ewjoachim: Ah, I misread sorry

ewjoachim: That being said... It's not 100% true, I think (not sure). Procrastinate doesn't care about its schema. I think the DB can be set so that the default schema is different, and then by default Procrastinate will use the default schema, no ?

elemoine: Not sure. You can change the `search_path` for a given session (using `SET search_path TO myschema`) but I am not sure you can attach a default schema to a database.

So, to my current knowledge, I think it's correct to say that Procrastinate will use the `public` schema by default.

docs/howto/set_database_schema.rst
elemoine: I don't want to mention pools at this point, as we don't have pools in Procrastinate for the moment.

docs/howto/set_database_schema.rst
elemoine: Yep, I added a note to the glossary about that. I just mention that the term "schema" is also by Postgres for something else.

elemoine:
Happy to make post-merge changes if you have more comments. Thanks a lot!

PR #166, opened 2020-02-23
(Deprecated PR) Use a pool instead of a single connection, and reorganize the worker around that

ewjoachim:
Cf. #148

I've tried to redo and complete my previous approach without touching the "run multiple jobs at the same time" part.
I think the next step will be the separation between sync connector and async connector.

### Successful PR Checklist:
- [x] Tests
- [ ] *Documentation*: Need to write more of it
- [ ] Had a good time contributing? Let's say I'll be glad when this will be behind us.

PR #173, opened 2020-03-20
Use a pool instead of a single connection, and reorganize the worker around that

### LINKED ISSUE #148: Procrastinate should be using a pool

ewjoachim:
Closes #148
Closes #166 

Sorry 100000000 times @elemoine 

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -1,12 +1,13 @@
+import asyncio
 import logging
 import warnings
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, Iterable, List, NoReturn, Optional
 
 import aiopg
 import psycopg2.sql
 from psycopg2.extras import Json, RealDictCursor
 
-from procrastinate import connector, utils
+from procrastinate import connector, sql, utils
 
 logger = logging.getLogger(__name__)
 
@@ -16,7 +17,6 @@ class PostgresConnector(connector.BaseConnector):
     def __init__(
         self,
         pool: aiopg.Pool,
-        socket_timeout: float = connector.SOCKET_TIMEOUT,
         json_dumps: Optional[Callable] = None,
         json_loads: Optional[Callable] = None,
     ):
@@ -33,7 +33,6 @@ class PostgresConnector(connector.BaseConnector):
 
         """
         self._pool = pool
-        self.socket_timeout = socket_timeout
         self.json_dumps = json_dumps
         self.json_loads = json_loads
 
@@ -55,7 +54,6 @@ class PostgresConnector(connector.BaseConnector):
     @classmethod
     async def create_with_pool_async(
         cls,
-        socket_timeout: float = connector.SOCKET_TIMEOUT,
         json_dumps: Optional[Callable] = None,
         json_loads: Optional[Callable] = None,
         **kwargs,
@@ -75,12 +73,6 @@ class PostgresConnector(connector.BaseConnector):
         .. __: https://aiopg.readthedocs.io/en/stable/core.html#aiopg.create_pool
         .. _psycopg2 doc: https://www.psycopg.org/docs/extras.html#json-adaptation
 
-        socket_timeout:
-            This parameter should generally not be changed.
-            It indicates the maximum duration (in seconds) procrastinate workers wait
-            between each database job pull. Job activity will be pushed from the db to
-            the worker, but in case the push mechanism fails somehow, workers will not
-            stay idle longer than the number of seconds indicated by this parameters.
         json_dumps:
             The JSON dumps function to use for serializing job arguments. Defaults to
             the function used by psycopg2. See the `psycopg2 doc`_.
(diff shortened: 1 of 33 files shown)


tests/unit/test_signals.py
```diff
@@ -32,3 +33,31 @@ def stop():
...
+
+def test_on_stop_signal_twice(kill_own_pid):
+    with pytest.raises(KeyboardInterrupt):
+        with signals.on_stop(callback=lambda: None):
+            kill_own_pid(signal=signal.SIGINT)
+            kill_own_pid(signal=signal.SIGINT)
+
+
+@pytest.mark.asyncio
+def test_on_stop_work_with_asyncio(kill_own_pid):
```
ewjoachim: await

procrastinate/app.py
```diff
@@ -203,32 +217,23 @@ def perform_import_paths(self):
...
+        wait: bool = True,
     ) -> None:
         """
-        Run a worker. This worker will run in the foreground
-        and the function will not return until the worker stops
-        (most probably when it receives a stop signal) (except if
-        `only_once` is True).
+        Run a worker. This worker will run in the foreground and the function will not
+        return until the worker stops (most probably when it receives a stop signal)
+        (except if `only_once` is True).
```
ewjoachim: to fix

procrastinate/aiopg_connector.py
```diff
@@ -120,30 +166,35 @@ def make_dynamic_query(self, query: str, **identifiers: str) -> str:
...
+                )
+
+            # We'll leave this loop with a CancelError, when we get cancelled
+            while True:
+                event.set()
+                # TODO: because of https://github.com/aio-libs/aiopg/issues/249,
+                # we could get stuck in here forever if the connection closes.
+                # Maybe we should set a timeout and if connection is closed, reopen
+                # a new one.
+                await connection.notifies.get()
```
ewjoachim: (why did we change the order of event.set() and notifies.get() ? I can't remember)

procrastinate/aiopg_connector.py
```diff
@@ -120,30 +166,35 @@ def make_dynamic_query(self, query: str, **identifiers: str) -> str:
...
+        async with self._pool.acquire() as connection:
+            for channel_name in channels:
+                await self._execute_query_connection(
+                    connection=connection,
+                    query=self.make_dynamic_query(
+                        query=sql.queries["listen_queue"], channel_name=channel_name
+                    ),
+                )
+
+            # We'll leave this loop with a CancelError, when we get cancelled
```
ewjoachim:
```suggestion
            # We'll leave this loop with a CancelledError, when we get cancelled
```

procrastinate/app.py
```diff
@@ -57,6 +60,14 @@ def __init__(
             A :py:func:`App.task` that has a custom "name" parameter, that is not
             imported and whose module path is not in this list will
             fail to run.
+        worker_timeout:
+            This parameter indicates the maximum duration (in seconds) procrastinate
+            workers wait between each database job pull. Job activity will be pushed
+            from the db to the worker, but in case the push mechanism fails somehow,
+            workers will not stay idle longer than the number of seconds indicated by
+            this parameters.
```
ewjoachim:
```suggestion
            this parameter.
```

procrastinate/signals.py
```diff
@@ -1,17 +1,29 @@
...
+# This all mean we have to save the previous signals before installing ours, and for
+# uninstalling, we need to remove the async handler and only then re-add the normal
+# one. And hope that there was no previously set async handler.
 
 
 @contextmanager
-def on_stop(callback: Callable[[signal.Signals, FrameType], None]):
+def on_stop(callback: Callable[[], None]):
     if threading.current_thread() is not threading.main_thread():
         logger.debug(
```
ewjoachim: warning

procrastinate/signals.py
```diff
@@ -22,19 +34,29 @@ def on_stop(callback: Callable[[signal.Signals, FrameType], None]):
...
         yield
     finally:
         logger.debug(
             "Resetting previous signal handler",
             extra={"action": "uninstall_signal_handlers"},
         )
+        loop.remove_signal_handler(signal.SIGINT)
+        loop.remove_signal_handler(signal.SIGTERM)
         signal.signal(signal.SIGINT, sigint_handler)
         signal.signal(signal.SIGTERM, sigterm_handler)
```
ewjoachim: These 4 lines should be in `else`

procrastinate/utils.py
```diff
@@ -67,21 +67,35 @@ def wrap_one(cls: Type, attribute_name: str):
...
     # Keep only async def methods
-    if not asyncio.iscoroutinefunction(attribute):
+    if not asyncio.iscoroutinefunction(wrapped):
         return
 
     # Create a wrapper that will call the method in a run_until_complete
-    @functools.wraps(attribute)
-    def wrapper(self, *args, **kwargs):
-        awaitable = attribute(self, *args, **kwargs)
+    # @functools.wraps(attribute)
```
ewjoachim:
```suggestion
    @functools.wraps(attribute)
```

tests/unit/test_signals.py
```diff
@@ -32,3 +33,31 @@ def stop():
...
+
+    def stop():
+        event.set()
+
+    async def wait_and_kill():
+        await asyncio.sleep(0.001)
+        kill_own_pid()
+
+    try:
+        with signals.on_stop():
```
ewjoachim:
```suggestion
        with signals.on_stop(stop):
```

procrastinate/app.py
ewjoachim:
```suggestion
        return until the worker stops (most probably when it receives a stop signal).
```

ewjoachim:
Ok, I added a bunch of commits to finish fixing tests and because there was a lot missing for coverage.

I think we're good now, @elemoine you can re-review if you want.

ewjoachim:
New implementation of signals fails with Py3.6 only 😭

elemoine:
> New implementation of signals fails with Py3.6 only sob

The issue is that `asyncio.get_event_loop()` does not return `None` if not running in the context of an asyncio task. Instead, it creates a event loop and returns it.

See the commit that I added to the branch for a fix. The fix involves using `asyncio.Task.current_task()` on Python 3.6 to test whether running in task context or not.

elemoine:
Might be worth cleaning up the Git commit history before merging. I let you decide @ewjoachim.

ewjoachim:
> Might be worth cleaning up the Git commit history before merging.

I could but I think the risk of screwing up is higher than the gain. Let the commits tell the story ! :D

Issue #176, opened 2020-03-27
Postgres image now requires `POSTGRES_PASSWORD`


ewjoachim:
Source: https://github.com/docker-library/postgres/pull/658#pullrequestreview-336007842
> This is a breaking change, but I think it's one that makes sense overall (and that the benefit outweighs the pain in this instance).

We need to add a password in `docker-compose.yml` and make sure it is passed properly in the contribution doc (`CONTRIBUTING.rst`)

ewjoachim:
@charlesaracil-ulti up for it ?

charlesaracil-ulti:
@ewjoachim of course !

ewjoachim:
Closed by #177

PR #177, opened 2020-03-27
add POSTGRES_PASSWORD to docker-compose.yml and in contributing docum…

charlesaracil-ulti:
…entation

Cf. #176 

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,7 @@ services:
       PGDATABASE: procrastinate
       PGHOST: postgres
       PGUSER: postgres
+      PGPASSWORD: password
     depends_on:
       - postgres
     user: ${UID:-0}:${GID:-0}
@@ -19,3 +20,4 @@ services:
     ports: ["5432:5432"]
     environment:
         POSTGRES_DB: procrastinate
+        POSTGRES_PASSWORD: password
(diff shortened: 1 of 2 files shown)


elemoine:
Bonus: the change is backward compatible, in the sense that `docker-compose up postgres` will work even without pulling the new `postgres:10` image.

docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
       PGDATABASE: procrastinate
       PGHOST: postgres
       PGUSER: postgres
+      POSTGRES_PASSWORD: password
```
elemoine: This should actually be set at postgres service level.

charlesaracil-ulti: you mean alongside the variable `POSTGRES_DB` ? It's not a environment variable ?

elemoine: You need `PGPASSWORD` in the procrastinate env vars (in the same way we set `PGPASSWORD` for `psql`), and `POSTGRES_PASSWORD` in the postgres env vars.

charlesaracil-ulti: it should be done thank you :)

elemoine:
Thanks @charlesaracil-ulti!

charlesaracil-ulti:
Thanks for you reviews and explanations !

PR #181, opened 2020-04-10
Admin shell 38

### LINKED ISSUE #38: Implement an administration prompt with cmd
Python's standard lib has https://docs.python.org/3/library/cmd.html , which will help us implement easily an interactive prompt.

We could probably use the following features:
- List queues (and number of queued & processing jobs per queue, with stats like processing rate, average wait time etc.)
- List tasks (and ... same stats)
- List jobs & search for a specific job (by task, queue, lock, id, maybe args)
- List stalled jobs and error jobs

(it would be nice if those pages could stay on and autorefresh)
(also, use colors ?)

- Display a job, and provide specific actions: discard it, retry it (now or later), manually set it to success or failure (well it's all about changing the state of the job)
- Batch action ? 

And finally, the cmd prompt should just be an interface, and the real actions should be implemented independently in an `admin` module, because we'll probably have exactly the same needs for the API and/or the web admin panel.

ewjoachim:
Closes #38 by @entwanne 

Just creating the PR for readability.

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/admin.py
+++ b/procrastinate/admin.py
@@ -1,0 +1,43 @@
+from procrastinate import connector as connector_module
+from procrastinate import sql, utils
+
+
+@utils.add_sync_api
+class Admin:
+    def __init__(self, connector: connector_module.BaseConnector):
+        self.connector = connector
+
+    async def list_jobs_async(self):
+        return [
+            dict(row) for row in await self.connector.execute_query_all(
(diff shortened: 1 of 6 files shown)


ewjoachim:
@entwanne would you mind re-opening a draft PR ? I'm using the "My PRs" view a lot and this one is cluttering it a bit :)

PR #182, opened 2020-04-10
Create Aiopg pool lazily

k4nar:
I was playing with the new pools (cf #173) in my project, and I had a few issues with the new design.

In order to create an App, I need a Connector. And in order to create a Connector, I either need to create my own pool, or to use `PosgresConnector.create_with_pool_async` which will create one for me. This has two implications:
- I can't create a pool without an event loop, so I can't instantiate an App before starting my event loop. Instantiating an App usually happen very early (before defining tasks), so it should not have the side effect of creating an event loop, nor require one.
- I can't instantiate an App without connecting to my DB. I like to be able to start my services without a DB connection for resiliency. This makes it close to impossible. Also it's very simple to end up in a situation where you have side-effects at import time, which is usually a bad idea.

So I've tried to mess around a bit with the `PostgresConnector` in order to create the loop lazily, like the connection is in Procrastinate 0.6. My solution is probably not perfect, and it will require some doc changes (but the current doc is not up-to-date with master anyway).

Please let me now what you think :) .  

### Successful PR Checklist:
- [ ] Tests

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -18,7 +18,8 @@ LISTEN_TIMEOUT = 30.0
 class PostgresConnector(connector.BaseConnector):
     def __init__(
         self,
-        pool: aiopg.Pool,
+        pool: Optional[aiopg.Pool] = None,
+        pool_kwargs: Optional[Dict] = None,
         json_dumps: Optional[Callable] = None,
         json_loads: Optional[Callable] = None,
     ):
@@ -30,11 +31,13 @@ class PostgresConnector(connector.BaseConnector):
         Parameters
         ----------
         pool:
-            An aiopg pool, either externally configured or passed by
-            :py:func:`PostgresConnector.create_with_pool`.
+            Optional. An aiopg pool externally configured
+        pool_kwargs:
+            Optional. Arguments passed to the aiopg pool during creation.
 
         """
         self._pool = pool
+        self._pool_kwargs = pool_kwargs or {}
         self.json_dumps = json_dumps
         self.json_loads = json_loads
 
@@ -42,8 +45,9 @@ class PostgresConnector(connector.BaseConnector):
         """
         Closes the pool and awaits all connections to be released.
         """
-        self._pool.close()
-        await self._pool.wait_closed()
+        if self._pool:
+            self._pool.close()
+            await self._pool.wait_closed()
 
     def _wrap_json(self, arguments: Dict[str, Any]):
         return {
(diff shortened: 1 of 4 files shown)


procrastinate/aiopg_connector.py
```diff
@@ -54,12 +58,12 @@ def _wrap_json(self, arguments: Dict[str, Any]):
         }
 
     @classmethod
-    async def create_with_pool_async(
+    def create_with_pool(
```
k4nar, line 61: To be honest, this method is pretty useless now. I think merging this with the init method would make things easier to understand.

ewjoachim:
We can discuss this but, why not consider creating the app lazily then ?

elemoine:
> We can discuss this but, why not consider creating the app lazily then ?

I think I agree with @k4nar that application developers would expect to be able to create the Application instance very early in the startup process, and possibly before the event loop is created.

If we don't want to go the "create pool lazily" way we could expose `set_pool` and `create_pool` methods.

```py
app = procrastinate.Application(connector=procrastinate.PostgresConnector())
```

And then, after the event loop has been created:

```py
pool = await app.connector.create_pool(**pool_args)
app.connector.set_pool(pool)
```

`PostgresConnector.create_pool()` would be provided as a convenience. The application developer could also create the pool by herself:

```py
pool = await aiopg.create_pool(**pool_args)
app.connector.set_pool(pool)
```

ewjoachim:
I do agree that we may have to end up doing this.

I feel a bit sad about the fact you can't instantiate a pool synchronously (even if you tell it not to connect right away, it insists on being created with a coroutine, and even if you tried calling the constructor yourself, it expects a running loop). So Aiopg gets away while saying "I want to be started after the loop", but it looks like we can't and I'm not sure what's the big difference that makes it acceptable for them and not us. It's because they refuse to instantiate a pool synchronously before starting the loop that we end up with the same problem.

That being said, I can accept that we're simply trying to do better than aiopg.

elemoine:
FWIW I'd rather expose `create_pool` and `set_pool` functions than creating the pool lazily in `execute_query`. Because the user can control when the pool is created. And because that gives an API for passing ones own pool to Procrastinate.

elemoine:
What I have in mind: https://github.com/peopledoc/procrastinate/compare/ele_pool.

k4nar:
That's a good alternative indeed. In my project I try to connect to the DB as late as possible, in order to be able to start before the DB without issues. I'll try to see if I can still do it with `set_pool`.

PR #185, opened 2020-04-17
Create aiopg pool lazily

elemoine:
Closes #182

### Successful PR Checklist:
- [x] Tests

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -7,7 +7,7 @@ import aiopg
 import psycopg2.sql
 from psycopg2.extras import Json, RealDictCursor
 
-from procrastinate import connector, sql, utils
+from procrastinate import connector, exceptions, sql, utils
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


elemoine:
This is ready for review.

Thanks @ewjoachim for pair-programming on this!

procrastinate/aiopg_connector.py
```diff
@@ -18,70 +18,23 @@
...
         self,
-        pool: aiopg.Pool,
+        *,
         json_dumps: Optional[Callable] = None,
         json_loads: Optional[Callable] = None,
+        **kwargs: Any,
     ):
         """
-        The pool connections are expected to have jsonb adapters.
+        Create a PostgreSQL connector. The connector is based on an
```
k4nar, line 27: "is based on", or "needs"?

procrastinate/aiopg_connector.py
```diff
@@ -103,31 +56,86 @@ async def create_with_pool_async(
...
+        if self._pool:
+            raise exceptions.PoolAlreadySet
+        self._pool = pool
 
-        return cls(pool=pool, json_dumps=json_dumps, json_loads=json_loads)
+    async def _get_pool(self) -> aiopg.Pool:
+        async with self._lock:
+            if not self._pool:
+                self.set_pool(await self._create_pool(self._pool_args))
+        return self._pool
```
k4nar, line 138: I think it doesn't hurt to add a fast path to avoid taking the lock every single time we need the pool:
```python
if self._pool:
    return self._pool

async with self._lock:
    ...
```

elemoine: Yep. We added the fast path, then got confused and thought it wouldn't work. But you're right, that works.

procrastinate/aiopg_connector.py
elemoine: I decided to use "uses" here, with other improvements to the docstrings. See 537ee70c9323017a500cca49b793b6d1006bac48.

elemoine:
Thanks for your review comments @k4nar. I took them into account, and made other improvements to the PostgresConnector docstrings.

procrastinate/aiopg_connector.py
ewjoachim: It works only if you re-test the value of self._pool inside the lock. We were confused because we tried moving the if instead of duplicating it.

ewjoachim: An alternative would be to put the lock in the `_create_pool` function. That would allow limiting logic & IO being mixed up !

ewjoachim: Ah no forget it !

elemoine: Yep.

PR #190, opened 2020-04-24
Create and use a decorator to wrap aiopg and pyscop2 erros as connect…

### LINKED ISSUE #140: Wrap errors in aiopg connector
Currently, database errors are raised from `psycopg2` or `aiopg`. However, we want to use our own errors, so that we have less dependancy with those librairies.

We need to catch all psyopg2/aiopg errors at the source so that we use our own custom-defined errors, with a `from` argument to not lose the base error.

This can probably be done inside a decorator that we apply on all `aiopg_connector` functions.

SBillion:
Closes #140 

### Successful PR Checklist:
- [x] Tests

This PR adds a decorator to wrap psycopg2 and aiopg errors as connector exceptions. This decorator is expected to be used on coroutine functions only. It has been used on PostgresConnector methods doing psycopg2 or aiopg potential errors

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -49,8 +49,8 @@ class PostgresConnector(connector.BaseConnector):
         py:func:`PostgresConnector.set_pool` method.
 
         All other arguments than ``json_dumps`` and ``json_loads`` are passed to
-        ``aiopg.create_pool`` (see aiopg documentation__), with default values that
-        may differ from those of ``aiopg`` (see the list of parameters below).
+        :py:func:`aiopg.create_pool` (see aiopg documentation__), with default values
+        that may differ from those of ``aiopg`` (see the list of parameters below).
 
         .. _psycopg2 doc: https://www.psycopg.org/docs/extras.html#json-adaptation
         .. __: https://aiopg.readthedocs.io/en/stable/core.html#aiopg.create_pool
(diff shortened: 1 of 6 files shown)


procrastinate/aiopg_connector.py
```diff
@@ -1,18 +1,34 @@
 import asyncio
+import contextlib
```
thomasperrot, line 2: Cet import n'est plus nécessaire :)

ewjoachim:
Ready for review !

ewjoachim:
:o coverage has dropped below 100%

I'll fix that elsewhere.

procrastinate/exceptions.py
```diff
@@ -23,4 +30,19 @@ def __init__(self, scheduled_at: datetime.datetime):
 
 
 class PoolAlreadySet(ProcrastinateException):
+    """
+    Indicates a call on connector.set_pool() was done but the
+    pool already had a set. Chaging the pool of a connector is not
```
CorBott, line 35:
```suggestion
    pool already had a set. Changing the pool of a connector is not
```

procrastinate/aiopg_connector.py
```diff
@@ -12,6 +14,25 @@
...
+    """
+
+    @functools.wraps(coro)
+    async def wrapped(*args, **kwargs):
+        try:
+            return await coro(*args, **kwargs)
+        except psycopg2.Error as exc:
+            raise exceptions.ConnectorException from exc
+
+    wrapped._exceptions_wrapped = True  # type: ignore
```
elemoine, line 33: Is this only for the test? If so I think this deserves a comment in the code.

ewjoachim, line 33:
```suggestion
    # Attaching a custom attribute to ease testability and make the
    # decorator more introspectable
    wrapped._exceptions_wrapped = True  # type: ignore
```

PR #191, opened 2020-04-24
Fix procrastinate_fetch_job

marco44:
This function works as is, but because of a side effect: when storing a
record into a composite type, if there are more attributes than
necessary, the remaining attributes are simply lost. So it worked
because the columns came from the query in the same order as needed by
the composite type. With this commit, no side effect anymore.

Closes no ticket. We detected this on our schemas because we started linting plpgsql in all our functions, using plpgsql_check


ewjoachim:
Added missing migration.

PR #192, opened 2020-04-24
Add a prompt to remember writing migrations

ewjoachim:
Also, encourage people not to discard the list completely.

### Successful PR Checklist:
- [ ] Tests < not applicable

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,10 @@
 Closes #<ticket number>
 
+<!-- Please do not remove this, even if you think you don't need it -->
 ### Successful PR Checklist:
+<!-- In case of doubt, we're here to help. CONTIRBUTING.md might help too -->
 - [ ] Tests
-- [ ] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation))
-- [ ] Had a good time contributing? (if not, feel free to give some feedback)
+  - [ ] (not applicable?)
+- [ ] Documentation
+  - [ ] (not applicable?)
+- [ ] Had a good time contributing?


ewjoachim:
Hmm, we can do better !

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,8 @@
 Closes #<ticket number>
 
+<!-- Please do not remove this, even if you think you don't need it -->
 ### Successful PR Checklist:
 - [ ] Tests
 - [ ] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation))
+- [ ] Migration, if applicable
```
CorBott, line 7:
```suggestion
- [ ] Migration procedure, if applicable
```

ewjoachim: I've changed it to "SQL Migrations". This refers to [this part](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#migrations) in the contributing doc.

k4nar:
I think that editing might be a bit confusing for some contributors. It's the last step in the process, so it shouldn't be confusing :) .

Maybe instead we could add a CI check with an error if you touch the schema but not the migrations? Or if you touch a migration corresponding to an earlier version.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,6 +1,12 @@
 Closes #<ticket number>
 
+<!-- Please do not remove this, even if you think you don't need it -->
 ### Successful PR Checklist:
+<!-- In case of doubt, we're here to help. CONTIRBUTING.md might help too -->
 - [ ] Tests
-- [ ] Documentation (optionally: [run spell checking](https://github.com/peopledoc/procrastinate/blob/master/CONTRIBUTING.rst#build-the-documentation))
-- [ ] Had a good time contributing? (if not, feel free to give some feedback)
+  - ([ ] not applicable ?)
```
elemoine, line 7: (nit) no space before question mark

ewjoachim:
> I think that editing might be a bit confusing for some contributors. It's the last step in the process, so it shouldn't be confusing :) .
Maybe instead we could add a CI check with an error if you touch the schema but not the migrations? Or if you touch a migration corresponding to an earlier version.

Ok so I I understood correctly, the idea is that once we'll have the pum checks, we won't need this checkbox, and I agree. Let's check with pum !

PR #202, opened 2020-05-04
Worker should instantiate Event lazily, and should use wait_for, not wait

### LINKED ISSUE #201: Worker should instantiate Even lazily, and should use wait_for, not wait
2 small errors that can create problems:

- Workers create their Event far too early at instanciation, instead of at proper runtime in the Run function. This shouldn't change many things if the worker is launched via the app as documented, but it can be problematic if the worker is instanciated and ran by the user. Even if this should not be the case, we'll be better off with the event created at the proper place
- Workers use asyncio.wait instead of wait_for because it doesn't need the Timeout exception. But when we use wait, we're responsible for the tasks afterwards (mainly cancelling them). We'd be much better off using wait_for and disregading the timeout exception. Note: as it it, this could be the cause for a memory leak.

ewjoachim:
Closes #201 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)
- [x] Had a good time contributing? < kinda. Not astonishing. Our tests there are a bit complicated but it's still the best we could imagine.

### DIFF
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -47,10 +47,11 @@ class Worker:
 
         self.current_job_context: Optional[Dict] = None
         self.stop_requested = False
-        self.notify_event = asyncio.Event()
+        self.notify_event: Optional[asyncio.Event] = None
 
     @contextlib.contextmanager
     def listener(self):
+        assert self.notify_event
         notifier = asyncio.ensure_future(
             self.job_store.listen_for_jobs(event=self.notify_event, queues=self.queues)
         )
@@ -62,6 +63,8 @@ class Worker:
     async def run(self) -> None:
         queues = self.queues
         display = queues_display(queues)
+        self.notify_event = asyncio.Event()
+        self.stop_requested = False
 
         self.logger.info(
             f"Starting worker on {display}",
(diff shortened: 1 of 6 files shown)


ewjoachim:
(sorry for the strange branch name)

ewjoachim:
Tests pass but missing coverage.

procrastinate/worker.py
```diff
@@ -102,8 +106,14 @@ async def single_worker(self):
                     },
                 )
                 self.notify_event.clear()
-                await asyncio.wait([self.notify_event.wait()], timeout=self.timeout)
-                self.notify_event.clear()
+                try:
+                    await asyncio.wait_for(
+                        self.notify_event.wait(), timeout=self.timeout
+                    )
+                    self.notify_event.clear()
```
elemoine, line 113: Any reason for calling `clear` within the `try` block?

ewjoachim: It's either this or in the `else`, because if we go through `except`, there's no need to clear because the event wasn't set.

So we could:
- Leave as-is
- Move it to `else` (but it's a bit less readable, but more correct)
- Move it out of the `try` but then we'll clear even when there's a timeout, which is useless

elemoine: My preference is either "move it to `else`" or "move it out of the `try`". I don't like "leave as-is" too much, but reading the code you don't quite know what can cause the `TimeoutError` exception (although it's quite clear in that particular case).

ewjoachim:
Maybe one way to simplify testing would be to put the loop in a different function from its contents, but that would make it quite harder to read.

ewjoachim:
For some reason, coverage is on strike.

procrastinate/worker.py
```diff
@@ -83,29 +86,38 @@ async def run(self) -> None:
...
-        display = queues_display(queues)
-
         while not self.stop_requested:
             job = await self.job_store.fetch_job(self.queues)
-            if not job:
-                if not self.wait:
+            if job:
+                await self.process_job(job=job)
+            else:
+                if not self.wait or self.stop_requested:
```
elemoine, line 97: Out of curiosity, is this `stop_requested` test required here, or is just an optimization or something? I find it weird that we test `stop_requested` at two different places in the loop.

ewjoachim: It's an optimization, so that we don't wait if stop was requested during the job fetching coroutine.

PR #204, opened 2020-05-07
Implement an administration prompt

### LINKED ISSUE #38: Implement an administration prompt with cmd

entwanne:
Closes #38 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/docs/howto/shell.rst
+++ b/docs/howto/shell.rst
@@ -1,0 +1,27 @@
+Use the administration shell
+----------------------------
+
+The procrastinate shell is a tool to administrate jobs and overview queues and tasks.
+It is an interactive shell that you can run with the following command.
+
+*Experimental feature*.
+
+.. code-block:: console
+
+    $ procrastinate shell
+    Welcome to the procrastinate shell.   Type help or ? to list commands.
+
+    procrastinate> help
+
+    Documented commands (type help <topic>):
+    ========================================
+    EOF  cancel  exit  help  list_jobs  list_queues  list_tasks  retry
+
+As usual, you should use ``--app`` argument or ``PROCRASTINATE_APP`` environment
+variable to specify the application you want to use (see `command_line`).
+
+There are commands to list all the jobs (``list_jobs``), tasks (``list_tasks``)
+& queues (``list_queues``).
+And commands to retry (``retry``) & cancel (``cancel``) a specific job.
+
+You can get help for a specific command *cmd* by typing ``help cmd``.
--- a/procrastinate/admin.py
+++ b/procrastinate/admin.py
@@ -1,3 +1,5 @@
+from typing import Any, Dict, Iterable
+
 from procrastinate import connector as connector_module
 from procrastinate import sql, utils
 
@@ -23,8 +25,13 @@ class Admin:
         self.connector = connector
 
     async def list_jobs_async(
-        self, id=None, queue=None, task=None, status=None, lock=None,
-    ):
+        self,
+        id: int = None,
+        queue: str = None,
+        task: str = None,
+        status: str = None,
+        lock: str = None,
+    ) -> Iterable[Dict[str, Any]]:
         """
         List all procrastinate jobs given query filters.
 
@@ -69,8 +76,8 @@ class Admin:
         ]
 
     async def list_queues_async(
-        self, queue=None, task=None, status=None, lock=None,
-    ):
+        self, queue: str = None, task: str = None, status: str = None, lock: str = None,
+    ) -> Iterable[Dict[str, Any]]:
         """
         List all queues and number of jobs per status for each queue.
 
@@ -110,8 +117,8 @@ class Admin:
         ]
 
     async def list_tasks_async(
-        self, queue=None, task=None, status=None, lock=None,
-    ):
+        self, queue: str = None, task: str = None, status: str = None, lock: str = None,
+    ) -> Iterable[Dict[str, Any]]:
         """
         List all tasks and number of jobs per status for each task.
 
@@ -150,7 +157,7 @@ class Admin:
             )
         ]
 
-    async def set_job_status_async(self, id, status):
+    async def set_job_status_async(self, id: int, status: str) -> Dict[str, Any]:
         """
         Set/reset the status of a specific job.
 
--- a/procrastinate/admin.py
+++ b/procrastinate/admin.py
@@ -1,0 +1,81 @@
+from procrastinate import connector as connector_module
+from procrastinate import sql, utils
+
+
+@utils.add_sync_api
+class Admin:
+    def __init__(self, connector: connector_module.BaseConnector):
+        self.connector = connector
+
+    async def list_jobs_async(
+        self, id=None, queue=None, task=None, status=None, lock=None,
+    ):
+        return [
+            {
+                "id": row["id"],
+                "queue": row["queue_name"],
+                "task": row["task_name"],
+                "lock": row["lock"],
+                "args": row["args"],
+                "status": row["status"],
+                "scheduled_at": row["scheduled_at"],
+                "attempts": row["attempts"],
+            }
+            for row in await self.connector.execute_query_all(
+                query=sql.queries["list_jobs"],
+                id=id,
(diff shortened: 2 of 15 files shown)


docs/howto/shell.rst
```diff
@@ -0,0 +1,25 @@
+Use the administration shell
+----------------------------
+
+The procrastinate shell is a tool to administrate jobs and overview queues and tasks.
+Itis an interactive shell that you can run with the following command.
```
ewjoachim, line 5:
```suggestion
It is an interactive shell that you can run with the following command:
```

entwanne:
> * We mentionned saying the shell is experimental for now in the intro, but we can add it quickly later.

Ok, I've just added it.

> * Did you have an opportunity to check if the new queries were likely to be problematic ?

Just a rapid check, I didn't notice anything wrong.

> * Would be equally nice to check with people who are going to use this in the PeopleDoc stack, to know if this would fit their first level needs (knowing that future evolutions will address those needs much better, but this is a first iteration, and a brilliant one ! )

Ok, I'll see further for that, but I don't think it's a blocking point.

docs/howto/shell.rst
```diff
@@ -0,0 +1,27 @@
...
+    $ procrastinate shell
+    Welcome to the procrastinate shell.   Type help or ? to list commands.
+
+    procrastinate> help
+
+    Documented commands (type help <topic>):
+    ========================================
+    EOF  cancel  exit  help  list_jobs  list_queues  list_tasks  retry
+
+As usually, you should use ``--app`` argument or ``PROCRASTINATE_APP`` environment
```
elemoine, line 20:
```suggestion
As usual, you should use ``--app`` argument or ``PROCRASTINATE_APP`` environment
```

procrastinate/admin.py
```diff
@@ -0,0 +1,174 @@
...
+            Filter by job task name
+        status: str
+            Filter by job status (*todo*/*doing*/*succeeded*/*failed*)
+        lock: str
+            Filter by job lock
+
+        Returns
+        -------
+        ``List[Dict[str, Any]]``
+            A list of dictionnaries representing queues stats (``name``, ``nb_jobs``,
```
elemoine, line 91: Minor but I think `num_jobs` makes for a better variable name. `nb_` is just in french, but I don't think it's common as an abbreviation for number in english. I might be wrong.

procrastinate/admin.py
```diff
@@ -0,0 +1,174 @@
...
+        connector:
+            Instance of a subclass of :py:class:`procrastinate.connector.BaseConnector`,
+            typically `PostgresConnector`. It will be responsible for all communications
+            with the database. Mandatory.
+        """
+        self.connector = connector
+
+    async def list_jobs_async(
+        self, id=None, queue=None, task=None, status=None, lock=None,
+    ):
```
elemoine, line 27: Do you want to add types for the function arguments and return values?

elemoine:
This looks great already. Thanks @entwanne!

procrastinate/admin.py
entwanne: Yes, I think it's relevant for this class.

procrastinate/admin.py
entwanne: Maybe it's just me but when I read `num` I think about "numero", so just a number without any context.
When I read `nb` I think about a counter. Then maybe `count_` would be more suitable?

elemoine: > Maybe it's just me but when I read num I think about "numero", so just a number without any context.

Yes, I know :-)

https://stackoverflow.com/questions/3742650/naming-conventions-for-number-of-foos-variables, for example, doesn't even mention `nb_` as a possible prefix. I really thing "nb_" is a french thing.

> Then maybe count_ would be more suitable?

Or as a suffix: `jobs_count`.

@ewjoachim any opinion about this?

ewjoachim: `jobs_count` is nice. It's not something we can't do on a second pass. If you want to change it, great, otherwise, I suggest we keep this in mind and make a follo-up PR. It's 1000 lines already !

ewjoachim:
(don't forget to squash your fixups)

docs/howto/shell.rst
```diff
@@ -0,0 +1,27 @@
...
+    Documented commands (type help <topic>):
+    ========================================
+    EOF  cancel  exit  help  list_jobs  list_queues  list_tasks  retry
+
+As usual, you should use ``--app`` argument or ``PROCRASTINATE_APP`` environment
+variable to specify the application you want to use (see `command_line`).
+
+There are commands to list all the jobs (``list_jobs``), tasks (``list_tasks``)
+& queues (``list_queues``).
+And commands to retry (``retry``) & cancel (``cancel``) a specific job.
```
elemoine, line 25: Given that we have `list_jobs` & co. I've been wondering if having `retry_job` instead of `retry`, and `cancel_job` instead of `cancel`, would make things more consistent and explicit. When I open the shell for the first time and see `cancel` as a possible command I was not sure what that command would cancel. Curious what others think…

procrastinate/admin.py
```diff
@@ -0,0 +1,181 @@
...
+        task: str = None,
+        status: str = None,
+        lock: str = None,
+    ) -> Iterable[Dict[str, Any]]:
+        """
+        List all procrastinate jobs given query filters.
+
+        Parameters
+        ----------
+        id: int
```
elemoine, line 40: We use spaces around `:` elsewhere in the code, and double-backticks around native types. So
```
id : ``int``
```
Not sure how important it is for the doc generation. @ewjoachim?

entwanne: Not everywhere: https://github.com/peopledoc/procrastinate/blob/64472ff6559dffaffb3d39f53217c1c2a99e60b7/procrastinate/app.py#L184-L189

But I'm ok to add spaces and backticks.

docs/howto/shell.rst
entwanne: I'm not sure if it's relevant because I see no confusion here.
What would be the meaning of "retry"/"cancel" applied to a task or a queue?

elemoine: Right. When I saw `cancel` for the first time I thought it would cancel the shell. But it may be just me :-) And `help cancel` immediately shed light on my confusion. Up to you really.

procrastinate/admin.py
elemoine: Yep, not that consistent indeed.

ewjoachim: I'd say: don't bother with that yet. If we want to enforce that someday, we'll do it with linters. Meanwhile, as long as sphinx & napoleon are happy, that's enough to me.

ewjoachim:
@entwanne please squash your fixup :) (btw if you force-push, GitHub still allow to see the diff, so pushing fixups is becoming less and less interesting with time IMHO)

entwanne:
I didn't see the last review, I was waiting for it before squashing & merge.

> btw if you force-push, GitHub still allows to see the diff, so pushing fixups is becoming less and less interesting with time IMHO

For me it's not really-well handled and it sometimes lose some comments, so I prefer to keep the fixup commits until the last step.

ewjoachim:
:clap: thank you !

PR #205, opened 2020-05-07
Closes #203 - Refactor builtin tasks to use pass_context

### LINKED ISSUE #203: builtin tasks should use pass_context
That would simplify builtin tasks a lot. For now, not having the context means they need to be created as closures, which make the whole thing much more complicated than it should be.

If builtin tasks received the context, then they could get the app, and from this, interact with the store.

Also, app is available in the context through the worker, but in reality, app should probably be passed to the context directly.

Apps will still be responsible for registering builtin tasks on startup, but at least, they will be "normal" tasks without "magic" parameters.

ewjoachim:
Closes #203

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/__init__.py
+++ b/procrastinate/__init__.py
@@ -1,10 +1,12 @@
 from procrastinate import metadata as _metadata_module
 from procrastinate.aiopg_connector import PostgresConnector
 from procrastinate.app import App
+from procrastinate.job_context import JobContext
 from procrastinate.retry import BaseRetryStrategy, RetryStrategy
 
 __all__ = [
     "App",
+    "JobContext",
     "BaseRetryStrategy",
     "PostgresConnector",
     "RetryStrategy",
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -21,6 +21,11 @@ Tasks
 .. autoclass:: procrastinate.tasks.Task
     :members: defer, defer_async, configure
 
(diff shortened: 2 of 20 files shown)


ewjoachim:
Hm, I've added some stuff to the PR. Coverage seems good, but it may still need some dedicated unit tests.

ewjoachim:
I can easily split the PR into multiple ones if need be

procrastinate/__init__.py
```diff
@@ -1,10 +1,12 @@
 from procrastinate import metadata as _metadata_module
 from procrastinate.aiopg_connector import PostgresConnector
 from procrastinate.app import App
+from procrastinate.job_context import JobContext
 from procrastinate.retry import BaseRetryStrategy, RetryStrategy
 
 __all__ = [
     "App",
+    "JobContext",
```
elemoine, line 9: Don't you think `JobDeferrer` should be added as well? Not saying it should be done with this PR though…

ewjoachim: When would it be needed in client code ?

elemoine: For type-checking. `JobDeferrer` is `configure_task`'s return type, and `configure_task` is part of the public API.

PR #207, opened 2020-05-15
Use click's version_option for cli

### LINKED ISSUE #111: Version cli flag should use click's `version_option`
Click already has an builtin function to implement the version flag for us.

https://click.palletsprojects.com/en/7.x/api/#click.version_option

SBillion:
Closes #111 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -58,14 +58,6 @@ def handle_errors():
         )
 
 
-def print_version(ctx, __, value):
-    if not value or ctx.resilient_parsing:
-        return
-    click.echo(f"{PROGRAM_NAME} {procrastinate.__version__}")
-    click.echo(f"License: {procrastinate.__license__}")
-    ctx.exit()
-
-
 @click.group(context_settings=CONTEXT_SETTINGS)
 @click.pass_context
 @click.option("--app", "-a", help="Dotted path to the Procrastinate app")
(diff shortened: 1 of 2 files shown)


procrastinate/cli.py
```diff
@@ -77,14 +69,7 @@ def print_version(ctx, __, value):
...
 )
-@click.option(
-    "-V",
-    "--version",
-    is_flag=True,
-    callback=print_version,
-    expose_value=False,
-    is_eager=True,
-)
+@click.version_option(None, '-V', '--version')
```
ewjoachim, line 72: So this:
```suggestion
@click.version_option('-V', '--version', version=None)
```
doesn't work ?

procrastinate/cli.py
```diff
@@ -77,13 +69,11 @@ def print_version(ctx, __, value):
...
-    "--version",
-    is_flag=True,
-    callback=print_version,
-    expose_value=False,
-    is_eager=True,
+@click.version_option(
+    procrastinate.__version__,
+    '-V',
+    '--version',
+    prog_name=PROGRAM_NAME
```
ewjoachim, line 76: Curiosity: why is `prog_name` required?

procrastinate/cli.py
SBillion: Nope:
`version_option() got multiple values for argument 'version'`

ewjoachim: Ah yes. ok.

procrastinate/cli.py
SBillion: For the test_version. Without it, it thinks cli is the program name so the test was broken. 2 choiches, define prog_name here or change the test to assert on cli instead of procrastinate prog_name

ewjoachim: nah, I like it that way.

ewjoachim:
@SBillion 
```console
$ tox -e format
$ git commit -a --amend
$ git push --force-with-lease
```
And you're good to go.

PR #215, opened 2020-05-16
Simplify migration tests

ewjoachim:
No ticket

Ping @elemoine : Starting from your work, I simplified the process a bit. It does the same, but a bit differently. You may or may not like 100%, we can discuss any part you think is still improvable, as usual :)

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/tests/migration/test_migration.py
+++ b/tests/migration/test_migration.py
@@ -43,8 +43,8 @@ def procrastinate_version():
 @pytest.fixture
 def pg_service_file(tmp_path):
     file_path = tmp_path / "pg_service.conf"
-    with open(file_path, "w") as file_:
-        file_.write(PG_SERVICES)
+    with open(file_path, "w") as file:
+        file.write(PG_SERVICES)
     yield file_path
 
 
--- a/tests/migration/test_migration.py
+++ b/tests/migration/test_migration.py
@@ -49,14 +49,14 @@ def pg_service_file(tmp_path):
 
 
 @pytest.fixture
-def environment(pg_service_file):
+def pg_service_env(pg_service_file):
     env = os.environ.copy()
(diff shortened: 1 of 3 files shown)


tests/migration/test_migration.py
```diff
@@ -1,114 +1,76 @@
...
+BASELINE = "0.5.0"
 
+# Pum config file is currently broken (https://github.com/opengisch/pum/issues/5)
+# When it's fixed, we can add a config file here, and then get rid of omit_table_dir
 @pytest.fixture
-def migrations_directory():
-    return pkg_resources.resource_filename("procrastinate", "sql/migrations")
+def pum():
 
+    env = {**os.environ, "PGSERVICEFILE": "tests/migration/pgservice.ini"}
```
ewjoachim, line 15: @elemoine This way, we don't even need to name the fixture

ewjoachim: in case you doubt,
```pycon
Python 3.6.9 (default, Dec 14 2019, 18:23:02)
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.8)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> {**{1: 2}, 3:4}
{1: 2, 3: 4}
>>> {**{3: 2}, 3:4}
{3: 4}
```

ewjoachim:
Dammit wrong button sorry.

tests/migration/test_migration.py
elemoine: I keep forgetting about this syntax. Looks good.

PR #219, opened 2020-05-22
Add notion of "queueing lock" 

### LINKED ISSUE #217: Add defer lock: keep some tasks for being enqueued more than once
(Edit @ewjoachim: Original title was "Introduce queue size?")

Here's my use-case:

Worker healthcheck tasks, launched periodically (every 2 minutes) by a cron job, through a dedicated Procrastinate queue. If the worker is down, healthcheck jobs will accumulate in the queue. When the Procrastinate worker is up again, it will execute all the healthcheck tasks in the queue, resulting in a burst of healthchecks. Instead, I'd like the worker to execute only one healthcheck task.

For this problem, I've been contemplating the idea of introducing the notion of queue size. On a defer, if the queue size is exceeded, the defer operation will basically be a noop, meaning that no new job will be inserted into the procrastinate_jobs table.

And this is the "defer_job" SQL query I've been considering for the case when a queue size is defined on the defer operation:

```sql
INSERT INTO procrastinate_jobs (queue_name, task_name, lock, args, scheduled_at)
SELECT %(queue)s, %(task_name)s, %(lock)s, %(args)s, %(scheduled_at)s
WHERE (
    SELECT COUNT(*)
        FROM procrastinate_jobs
       WHERE status = 'todo' AND queue_name = %(queue)s
) < %(queue_size)s RETURNING id;
```

I think there's a possible race condition, when multiple defer operations occur at the time. Simultaneous defer operations may see a queue that is not full, and therefore insert a new job into the queue, possibly exceeding the queue size. But that might not be a problem for the use-cases we want to support with this new functionality.

Interested in discussing this use-case, and the "queue size" functionality I've been considering for it.

elemoine:
Closes #217 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)


procrastinate/cli.py
```diff
@@ -122,6 +122,11 @@ def worker(app: procrastinate.App, queue: Iterable[str], name: Optional[str]):
 @click.option(
     "--lock", help="A lock string. Jobs sharing the same lock will not run concurrently"
 )
+@click.option(
+    "--defer-lock",
+    help='A defer lock string. The defer operation will trigger a "DeferLockTaken" '
+    'exception if there already is a job in the "todo" state with the same defer lock',
```
ewjoachim, line 128:
```suggestion
    'exception if there already is a job waiting in the queue with the same defer lock',
```

elemoine:
@ewjoachim before writing docs for this change, I'd like to know if you agree with the implementation and terminology: the column name "defer_lock" and the exception class name "DeferLockTaken".

tests/unit/test_store.py
```diff
@@ -25,6 +26,36 @@ async def test_store_defer(job_store, job_factory, connector):
...
+        constraint_name="procrastinate_jobs_defer_lock_idx"
+    )
+
+    connector.execute_query_one = mocker.Mock(side_effect=exc)
+    mocker.patch.object(store.JobStore, "UniqueViolation", UniqueViolation)
+
+    with pytest.raises(exceptions.DeferLockTaken) as excinfo:
+        await job_store.defer_job(job=job_factory(task_kwargs={"a": "b"}))
+
+        assert excinfo.value.__cause__ is exc.__cause__
```
elemoine, line 56: @ewjoachim you want to like the mocks :-) If you don't I see no other ways than removing these unit tests entirely, and just relying on the corresponding integration tests.

elemoine:
Thanks a lot @CorBott!

procrastinate/tasks.py
```diff
@@ -126,6 +128,9 @@ def configure(
         ----------
         lock :
             No two jobs with the same lock string can run simultaneously
+        defer_lock :
+            No two jobs with the same defer lock can be in the "todo" state in
```
ewjoachim: (reword)

procrastinate/jobs.py
```diff
@@ -48,6 +48,9 @@ class Job:
     lock :
         Lock object in database that prevents other jobs with the same lock from running
         concurrently.
+    defer_lock:
+        Defer lock object in database that prevents having several jobs with the same
+        defer lock and in the "todo" state.
```
ewjoachim: (reword)

procrastinate/exceptions.py
```diff
@@ -46,3 +46,12 @@ class ConnectorException(ProcrastinateException):
     """
 
     pass
+
+
+class DeferLockTaken(ProcrastinateException):
+    """
+    Indicates that a job in todo state with the same defer lock
+    already exists in the database.
```
ewjoachim, line 54: (reword)

tests/integration/test_store.py
```diff
@@ -387,14 +387,39 @@ async def test_defer_job(pg_job_store, get_all):
...
+    )
+    with pytest.raises(exceptions.DeferLockTaken):
+        await pg_job_store.defer_job(
+            jobs.Job(
+                id=2,
+                queue="queue_a",
+                task_name="task_1",
+                lock="lock_1",
+                defer_lock="defer_lock_1",
+                task_kwargs={"a": "b"},
```
ewjoachim, line 430: Please change all params except defer_lock

procrastinate/cli.py
```diff
@@ -122,6 +122,11 @@ def worker(app: procrastinate.App, queue: Iterable[str], name: Optional[str]):
...
+)
```
ewjoachim, line 130: Add `"-i", "--ignore-already-enqueued/--no-ignore-already-enqueued"`

tests/integration/test_cli.py
```diff
@@ -126,7 +128,9 @@ def mytask(a):
 
 def test_defer_unknown(entrypoint, click_app, connector):
     # No space in the json helps entrypoint() to simply split args
-    result = entrypoint("""-a yay defer --unknown --lock=sherlock hello {"a":1}""")
+    result = entrypoint(
+        """-a yay defer --unknown --lock=sherlock --defer-lock=houba hello {"a":1}"""
```
ewjoachim: Remove all unrelevant params.

procrastinate/store.py
```diff
@@ -17,15 +19,24 @@ def __init__(self, connector: connector.BaseConnector):
...
+                queue=job.queue,
+            )
+        except exceptions.ConnectorException as exc:
+            cause = exc.__cause__
+            if (
+                isinstance(cause, psycopg2.errors.UniqueViolation)
+                and cause.diag.constraint_name == "procrastinate_jobs_defer_lock_idx"
+            ):
+                raise exceptions.DeferLockTaken from exc.__cause__
+            raise
```
ewjoachim:
```suggestion
        except exceptions.ConnectorException as exc:
            try:
                cause =  exc.__cause__
                raise cause
            except self.UniqueViolation:
                if cause.diag.constraint_name == "procrastinate_jobs_queueing_lock_idx":
                    raise exceptions.AlreadyEnqueued from cause
```

tests/integration/test_store.py
```diff
@@ -423,3 +424,8 @@ async def test_defer_job_violate_defer_lock(pg_job_store):
                 task_kwargs={"a": "b"},
             )
         )
+        assert isinstance(excinfo.value.__cause__, psycopg2.errors.UniqueViolation)
+        assert (
+            excinfo.value.__cause__.diag.constraint_name
+            == "procrastinate_jobs_defer_lock_idx"
+        )
```
ewjoachim:
```suggestion
    assert isinstance(excinfo.value.__cause__, psycopg2.errors.UniqueViolation)
    assert (
        excinfo.value.__cause__.diag.constraint_name
        == "procrastinate_jobs_defer_lock_idx"
    )
```

procrastinate/exceptions.py
```diff
@@ -55,3 +55,17 @@ class AlreadyEnqueued(ProcrastinateException):
     """
 
     pass
+
+
+class UniqueViolation(ConnectorException):
+    """
+    A unique constraint is violated. The constraint name is availabe in
```
elemoine, line 62:
```suggestion
    A unique constraint is violated. The constraint name is available in
```

elemoine: I fixed it.

elemoine:
Now with docs! @ewjoachim this is ready for review again. Once reviewed by you, I'll request a final review from @CorBott.

tests/acceptance/test_nominal.py
```diff
@@ -62,29 +62,29 @@ def func(*queues, name="worker"):
 def test_nominal(defer, worker):
     from .param import Param
 
-    defer("sum_task", a=5, b=7)
-    defer("sum_task_param", p1=Param(3), p2=Param(4))
-    defer("increment_task", a=3)
+    defer("sum_task", [], a=5, b=7)
```
ewjoachim, line 65: Ah damn. Maybe this should be the default ?

ewjoachim: Sorry about that :)

elemoine: I did not want to change your work :-) But yes, I can change this.

procrastinate/exceptions.py
```diff
@@ -50,18 +50,17 @@ class ConnectorException(ProcrastinateException):
 
 class AlreadyEnqueued(ProcrastinateException):
     """
-    Indicates that a job in todo state with the same queueing lock
-    already exists in the database.
+    Indicates that there already is job waiting in the queue with the same queueing
```
ewjoachim, line 53:
```suggestion
    Indicates that there already is a job waiting in the queue with the same queueing
```

procrastinate/jobs.py
```diff
@@ -49,8 +49,8 @@ class Job:
         Lock object in database that prevents other jobs with the same lock from running
         concurrently.
     queueing_lock:
-        Queueing lock object in database that prevents having several jobs with the same
-        queueing lock and in the "todo" state.
+        Lock object in database that prevents having several jobs waiting in the queue
```
ewjoachim, line 52: Lock is not exactly an "object". Maybe "lock name" ?

docs/glossary.rst
```diff
@@ -62,4 +62,10 @@ Let's go through a few words and their meaning.
         sequential manner. No two jobs with the same lock can run simultaneously. See
         the `howto-locks` section for more information.
 
+    Queueing Lock
+        When configuring a job using `Task.configure` you can attach a queueing lock to
+        the job. No two jobs with the same queueing lock can be waiting in the queue.
+        And `Task.defer` will raise an `AlreadyEnqueued` exception if there already is
```
ewjoachim, line 68:
```suggestion
        `Task.defer` will raise an `AlreadyEnqueued` exception if there already is
```

docs/glossary.rst
```diff
@@ -62,4 +62,10 @@ Let's go through a few words and their meaning.
...
+        a job waiting in the queue with the same queueing lock.
```
ewjoachim, line 69: Not sure this sentence belongs in the glossary.

docs/howto/cron.rst
```diff
@@ -16,9 +16,9 @@ already has, in the form of `unix cron`_ and `systemd timers`_, to name just two
 That being said, Procrastinate tries to ease the use of these solutions by providing
 a means to easily schedule jobs from the command line.
 
-It also includes "queueing locks" to avoid accumulating jobs in the queue when the
-Procrastinate workers are down or too busy, and preventing bursts of job executions
-when workers are up again.
+It also features :term:`queueing lock` to avoid accumulating jobs in the queue when the
```
ewjoachim, line 19:
```suggestion
It also features :term:`queueing locks <queueing lock>` to avoid accumulating jobs in the queue when the
```

docs/howto/remove_old_jobs.rst
```diff
@@ -16,8 +16,8 @@ For more information about this task's parameter,
 see :py:func:`~procrastinate.builtin_tasks.remove_old_jobs`
 
 If you launch ``remove_old_jobs`` from a cron you may want to avoid inserting a new job
-when there is one already waiting in the queue. You can rely on a "queueing lock" for
-that:
+when there is one already waiting in the queue. You can rely on a :term:`queueing lock`
```
ewjoachim, line 19:
```suggestion
when there is one already waiting in the queue. You can rely on a :term:`queueing locks <queueing lock>`
```

docs/howto/remove_old_jobs.rst
```diff
@@ -36,7 +36,7 @@ In Python code
 You can access the builtin task through `App.builtin_tasks`.
 The parameters are the same than when accessing the task through the CLI.
 
-For example, to use a "queueing lock":
+For example, to use a :term:`queueing lock`:
```
ewjoachim, line 39: No need to link if the word has already been used in the page.

docs/spelling_wordlist.txt
```diff
@@ -18,9 +18,18 @@ Dramatiq
...
+programmatically
+pum
+pycharm
+queueing
 quickstart
 reimplement
+schemas
 sync
+tox
 Travis
```
ewjoachim, line 35: Ha, you managed to get the spell check working on your host ? Nice job !

procrastinate/jobs.py
elemoine: I've used the same wording as for the `lock` argument, but I can add a commit to change the wording of both. I'll do that.

docs/glossary.rst
elemoine: Yeah, you're right. I'll remove it.

docs/spelling_wordlist.txt
elemoine: Yep, I needed the aspell-en Arch package.

procrastinate/jobs.py
elemoine: Done. I've used the same docstrings as in the Task class, which brings extra consistency in our docs :-)

elemoine:
@ewjoachim thanks for the thorough review, I've taken all your comments into account.

docs/glossary.rst
```diff
@@ -61,4 +61,14 @@ Let's go through a few words and their meaning.
...
 
+    Lock
+        When configuring a job using `Task.configure` you can attach a lock to the job.
+        Jobs with the same lock are guaranteed to be executed in defer order, and in a
+        sequential manner. No two jobs with the same lock can run simultaneously. See
+        the `howto-locks` section for more information.
+
+    Queueing Lock
+        When configuring a job using `Task.configure` you can attach a queueing lock to
+        the job. No two jobs with the same queueing lock can be waiting in the queue.
```
k4nar, line 72: Shouldn't we talk about `AlreadyEnqueued` here? Otherwise what happens if we try to enqueue two jobs is left unclear.

elemoine: @ewjoachim said otherwise so I removed it with https://github.com/peopledoc/procrastinate/pull/219/commits/98a068a0b4fa2a2401227ac4b8e059aec4351910. But happy to add it back if others think it's worth mentioning `AlreadyEnqueued` here.

ewjoachim: I think the glossary should be there to explain the terms, not the behaviour. That being said, we can link to pages that explain this. For this feature, I think there is no dedicated page. Maybe we could link to the ref doc...

k4nar: Ok, makes sense :+1: . I think I was biased because this was the first thing I read from this PR, but this will probably not be the typical user experience :) .

Issue #101, opened 2019-10-24
Complete/fix mypy integration


ewjoachim:
This is a 2-fold ticket:

- Without a py.typed, our type annotations cannot be used by projects using procrastinate, and it's quite sad. Let's add a py.typed (see https://www.python.org/dev/peps/pep-0561/ )
- Mypy doesn't like our metaprogramming trick @async.add_sync_api and frankly, I understand it makes its job harder. Let's help it by implementing a mypy plugin. We can take direct inspiration from https://github.com/strawberry-graphql/strawberry/blob/4d11e2d26cf3f8f1453c95106606acede85125ac/strawberry/ext/mypy_plugin.py and https://github.com/strawberry-graphql/strawberry/tree/4d11e2d26cf3f8f1453c95106606acede85125ac#type-checking (doc: https://mypy.readthedocs.io/en/latest/extending_mypy.html )

If this ticket is succesful, then from a new python project using mypy, installing procrastinate from PyPI, on the following lines:

```python
app = procrastinate.App(...)
app.run_worker(["hello"])  # should pass mypy
app.run_worker("hello")  # should fail mypy
```

ewjoachim:
So, we're using `add_sync_api` less and less in our external API, so this doesn't make as much sense as it used to.

I wonder if we could just provide type stubs somewhere...

ewjoachim:
We're using pyright instead of mypy now, and we've removed most of the metaprogramming.

PR #224, opened 2020-05-29
Fixes 223

### LINKED ISSUE #223: Improvments for Admin code
- Add `ORDER_BY` clause in list commands
- Use job factory in tests to make them more stable (btw, make unspecified parameters in job factory random)
- Add queueing lock info in the shell

(ping @entwanne FYI)

ewjoachim:
Closes #223 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)


procrastinate/admin.py
```diff
@@ -47,39 +48,44 @@ async def list_jobs_async(
...
+            )
+        ]
 
     async def list_queues_async(
-        self, queue: str = None, task: str = None, status: str = None, lock: str = None,
+        self,
+        queue: Optional[str] = None,
+        task: Optional[str] = None,
+        status: Optional[str] = None,
+        lock: Optional[str] = None,
```
entwanne, line 88: Doesn't this one need a `queueing_lock` parameter as well?

procrastinate/shell.py
```diff
@@ -96,9 +97,9 @@ def do_retry(self, arg):
 
         Example: retry 2
         """
-        print_job(self.admin.set_job_status(arg, status="todo"))
+        print_job(self.admin.set_job_status(arg, status="todo"))  # type: ignore
```
entwanne, line 100: Why do you need to ignore typing here?

procrastinate/admin.py
```diff
@@ -101,29 +107,30 @@ async def list_queues_async(
...
+            )
+        ]
 
     async def list_tasks_async(
-        self, queue: str = None, task: str = None, status: str = None, lock: str = None,
+        self,
+        queue: Optional[str] = None,
+        task: Optional[str] = None,
+        status: Optional[str] = None,
+        lock: Optional[str] = None,
```
entwanne, line 133: Same here?

procrastinate/admin.py
ewjoachim: I'm not sure it makes sense... What information would you be searching if you listed queues with a queueing_lock filter ?

Happy to add it if you have a use case or if you think the lack of consistency in this aspect would be confusing.

procrastinate/shell.py
ewjoachim: mypy undestands `set_job_status_async` but not `set_job_status`.

https://github.com/peopledoc/procrastinate/issues/101

procrastinate/admin.py
ewjoachim: Same answer :)

procrastinate/admin.py
entwanne: No it was just for the consistency with `list_jobs`.

PR #227, opened 2020-05-29
[WIP] Attempt to fix procrastinate_fetch_job

### LINKED ISSUE #212: Race condition in locked tasks: wrong execution order
Re-activated by #210, this test produces surprising results:

https://travis-ci.org/github/peopledoc/procrastinate/jobs/687724543

It looks like jobs sometimes aren't ran in order, which contradicts one of our promises. We should really investigate this !

elemoine:
Closes #212 

This PR is a first attempt to fix procrastinate_fetch_job, which, I think, is currently subject to race conditions.

This is WIP for the moment, and the new query might be bad performance-wise.

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/sql/schema.sql
+++ b/procrastinate/sql/schema.sql
@@ -49,15 +49,16 @@ DECLARE
 	found_jobs procrastinate_jobs;
 BEGIN
 	WITH potential_job AS (
-		SELECT procrastinate_jobs.*
-			FROM procrastinate_jobs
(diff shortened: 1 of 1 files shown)


ewjoachim:
Does this mean you’ll lock every takable row with a select fir update each time you want to take a single one, until the end of the transaction ?

elemoine:
Yes. But that was just a try. That doesn't work. At this point I think I understand where the race condition is, but I don't know how to fix it.

PR #231, opened 2020-06-05
Close race condition in procrastinate_fetch_job

### LINKED ISSUE #212: Race condition in locked tasks: wrong execution order

elemoine:
This closes a race condition in the procrastinate_fetch_job plpgsql function, where jobs sharing the same lock can be run out of order.

With this commit jobs with the same lock are **always** executed in order, whatever their ETAs and queues.

In effect:
- if job A in queue 1 (id 1) and job B in queue 2 (id 2) have the same lock, and no workers process queue 1, then job B won't be executed, because job A must be executed first
- if job A is deferred with ETA 1 year, no other jobs with the same lock will be executed for 1 year

The lock name may change from "lock" to "serial lock" in the future.

Closes #212.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?) we have no SQL tests for the moment
- [ ] Documentation
  - [ ] (not applicable?) although it'd be good to mention that jobs with the same lock are now always executed in order
- [x] Had a good time contributing? that was a highly collaborative PR!

### DIFF
--- a/docs/discussions.rst
+++ b/docs/discussions.rst
@@ -80,10 +80,10 @@ About locks
 -----------
 
 Let's say we have a :term:`task` that writes a character at the end of a file after
-waiting for a random amount of time. This represents a real world problem where tasks
+waiting for a random amount of time. This represents a real world problem where jobs
 take an unforeseeable amount of time and share resources like a database.
 
-We launch 4 tasks respectively writing ``a``, ``b``, ``c`` and ``d``. We would expect
+We launch 4 jobs respectively writing ``a``, ``b``, ``c`` and ``d``. We would expect
 the file to contain ``abcd``, but it's not the case, for example maybe it's ``badc``.
 The jobs were taken from the queue in order, but because we have several workers, the
 jobs were launched in parallel and because their duration is random, the final result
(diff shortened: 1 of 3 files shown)


elemoine:
Making this a draft PR for the moment, as I'd like that #224 gets merged first, and that we cut a new release right after #224.

ewjoachim:
I unchecked documentation. I think we should have a dedicated page on serial locking. We should also explicitely mention that this is the current way of implementing a chain.

procrastinate/sql/migrations/delta_0.8.1_002_close_fetch_job_race_condition.sql
```diff
@@ -0,0 +1,46 @@
+DROP TABLE IF EXISTS procrastinate_job_locks;
```
ewjoachim, line 1: We'll be explicit in the changelog that with those migrations, the workers will need to be stopped when running the migration.
Of course, one can always write better migrations.

elemoine:
> I unchecked documentation. I think we should have a dedicated page on serial locking. We should also explicitely mention that this is the current way of implementing a chain.

We have https://procrastinate.readthedocs.io/en/stable/howto/locks.html already, and I think it's quite explicit.

tmartinfr:
@elemoine @ewjoachim I have a doubt, with the new query, are we changing the behaviour for scheduled tasks ? For example, if the task with the smallest id is scheduled in the future, will we block any execution of tasks ? When the previous query would have fetched the lowest-id task anyway ?

tmartinfr:
@elemoine We should remove this section of the doc : https://github.com/peopledoc/procrastinate/blob/master/docs/discussions.rst#the-procrastinate_job_locks-table, and maybe update the mention `Unavailable tasks, either locked, scheduled for the future or in a queue that the worker doesn't listen to, will be ignored.` here : https://github.com/peopledoc/procrastinate/blob/master/docs/discussions.rst#about-locks

elemoine:
> @elemoine We should remove this section of the doc : https://github.com/peopledoc/procrastinate/blob/master/docs/discussions.rst#the-procrastinate_job_locks-table,

:+1: 



> `Unavailable tasks, either locked, scheduled for the future or in a queue that the worker doesn't listen to, will be ignored.` here : https://github.com/peopledoc/procrastinate/blob/master/docs/discussions.rst#about-locks

I do not see the problem with that statement, but that may be me.

tmartinfr:
> > `Unavailable tasks, either locked, scheduled for the future or in a queue that the worker doesn't listen to, will be ignored.` here : https://github.com/peopledoc/procrastinate/blob/master/docs/discussions.rst#about-locks
> 
> I do not see the problem with that statement, but that may be me.

Not wrong, but it appeared not very clear for me. I could suggest something like : `When a worker requests a task, it will always receive the oldest available task. If this oldest task is unavailable, either locked, scheduled for the future or in a queue that the worker doesn't listen to, no other task will be selected.` (can be improved, and if you want to keep the existing version, it's fine for me)

docs/discussions.rst
```diff
@@ -115,9 +115,7 @@ their identifiers could be used (there's no hard limit on the length of a lock s
 but stay reasonable).
 
 A task can only take a single lock so there's no dead-lock scenario possible where two
-running tasks are waiting one another. That being said, if a worker dies with a lock, it
-will be up to you to free it. If the task fails but the worker survives though, the
-lock will be freed.
```
ewjoachim, line 120: It this outdated ?

ewjoachim: we could rephrasing it as "If a worker is killed without ending its job, following jobs with the same lock will not run until the interrupted job is either manually set to "failed" or "succeeded". If a job simply fails, following jobs with the same locks may run.

procrastinate/sql/migrations/delta_0.9.0_001_close_fetch_job_race_condition.sql
```diff
@@ -0,0 +1,48 @@
+DROP TABLE IF EXISTS procrastinate_job_locks;
```
ewjoachim, line 1: We'll be explicit in the changelog that with those migrations, the workers will need to be stopped when running the migration.
Of course, one can always write better migrations.

docs/discussions.rst
```diff
@@ -115,9 +115,7 @@ their identifiers could be used (there's no hard limit on the length of a lock s
...
+running tasks are waiting one another.
```
CorBott, line 118:
```suggestion
running tasks are waiting for one another.
```

procrastinate/sql/schema.sql
```diff
@@ -35,45 +35,46 @@ CREATE TABLE procrastinate_jobs (
 -- this prevents from having several jobs with the same queueing lock in the "todo" state
 CREATE UNIQUE INDEX procrastinate_jobs_queueing_lock_idx ON procrastinate_jobs (queueing_lock) WHERE status = 'todo';
 
+-- this prevents from having several jobs with the same lock in the "doing" state
+CREATE UNIQUE INDEX procrastinate_jobs_lock_idx ON procrastinate_jobs (lock) WHERE status = 'doing';
+
+CREATE INDEX ON procrastinate_jobs(queue_name);
```
k4nar, line 41: I don't see it in the migration, is it normal?

elemoine: Yes, it was just moved to here from a different location in the `schema.sql` file.

ewjoachim:
We wanted to rename the lock "serial lock" but this will change a LOT of code, so it will have its own PR.

Issue #175, opened 2020-03-27
Auto-deploy to PyPI is still broken


ewjoachim:
It looks like our .travis.yml syntax is still broken, due to a mix between matrix and jobs.

ewjoachim:
See https://travis-ci.org/github/peopledoc/procrastinate/builds/666683609/config?utm_medium=notification&utm_source=github_status :
![Screenshot from 2020-03-27 15-26-15](https://user-images.githubusercontent.com/1457576/77765995-59333480-703f-11ea-8710-666a5a2651df.png)

ewjoachim:
Closed by #179

ewjoachim:
FYI: https://config.travis-ci.com/explore !

ewjoachim:
Still-still broken :(

ewjoachim:
Still a problem with lightweight tags.

PR #234, opened 2020-06-05
Make setup.py work against lightweight tags

elemoine:
Refs #175 

This is to make "deploy to pypi" work for lightweight release tags.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ def ensure_version():
     # If running from the git repository
     try:
         version = (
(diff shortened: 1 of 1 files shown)


ewjoachim:
Hm, travis is on strike. I suggest we raise their salary and the re-push.

PR #258, opened 2020-06-17
Make listen/notify optional

### LINKED ISSUE #241: Comments regarding the "maxsize" pool argument
Currently we force "maxsize" to be at least 2:

https://github.com/peopledoc/procrastinate/blob/4a726b0d74e70026c29606d1c165ec55ce3c9c77/procrastinate/aiopg_connector.py#L128-L129

Two comments about this:

1. There are applications where the "defer" code base and the "worker" code base are separate. And one may want to use `maxsize=1` on the "defer" side
2. And `maxsize=1` on the "worker" side may even make sense, when one wants to avoid leaving open connections after the "fetch_job" and "finish_job" operations

This second use-case is of interest for my application. We use 2 worker servers, with 8 workers per server. So with maxsize=2 this means 32 open connections all the time. With maxsize=1 it would be mean 16 open connections.

ewjoachim:
Closes #241 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/docs/discussions.rst
+++ b/docs/discussions.rst
@@ -200,6 +200,8 @@ distinct sync workers and async workers. Of course, if your program is not that
 time-sensitive and you have sufficiently few blocking tasks, it's perfectly OK not to
 care.
 
+.. _discussions-pool-size:
+
 Mind the size of your PostgreSQL pool
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -207,6 +209,13 @@ You can size the PostgreSQL pool using the ``maxsize`` argument of `AiopgConnect
 Procrastinate will use use one connection to listen to server-side ``NOTIFY`` calls (see
 :ref:`discussion-general`). The rest of the pool is used for :term:`sub-workers
 <Sub-worker>`.
+
+.. warning::
+
+    Setting ``maxsize`` to ``0`` will actually disable the maximum. Setting it to
+    ``1`` will work, but the ``LISTEN/NOTIFY`` feature will be disabled.
+    Disabling this feature independently of your pool size is possible with
+    ``listen_notify=False``, see `howto/connections`.
 
 The relative sizing of your pool and your sub-workers all depends on the average length
 of your jobs, and especially compared to the time it takes to fetch jobs and register
--- a/docs/discussions.rst
+++ b/docs/discussions.rst
@@ -212,10 +212,12 @@ Procrastinate will use use one connection to listen to server-side ``NOTIFY`` ca
 
 .. warning::
 
-    Setting ``maxsize`` to ``0`` will actually disable the maximum. Setting it to
-    ``1`` will work, but the ``LISTEN/NOTIFY`` feature will be disabled.
-    Disabling this feature independently of your pool size is possible with
-    ``listen_notify=False``, see `howto/connections`.
+    Be careful not setting ``maxsize`` to ``0``: apart from disabling the maximum this
+    will likely trigger bugs__. Setting it to ``1`` will work, but the ``LISTEN/NOTIFY``
+    feature will be disabled. Disabling this feature independently of your pool size is
+    possible with ``listen_notify=False``, see `howto/connections`.
+
+    .. __: https://github.com/aio-libs/aiopg/issues/311
 
 The relative sizing of your pool and your sub-workers all depends on the average length
 of your jobs, and especially compared to the time it takes to fetch jobs and register
--- a/docs/howto/connections.rst
+++ b/docs/howto/connections.rst
@@ -11,7 +11,7 @@ Disabling the ``LISTEN/NOTIFY`` feature (see `discussion-general`) will use one 
 connection per worker::
 
     app = procrastinate.App(worker_defaults={"listen_notify": False})
-    # or when launchin a worker:
+    # or when launching a worker:
     app.run_worker(listen_notify=False)
 
 This works from the command line too:
--- a/docs/howto/connections.rst
+++ b/docs/howto/connections.rst
@@ -1,0 +1,24 @@
+Limit the number of openned connections
+=======================================
+
+By default, each worker using an `AiopgConnector` will open up to 10 parallel
+connections. You can control this value with the ``maxsize`` parameter (from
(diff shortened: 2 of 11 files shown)


docs/howto/connections.rst
```diff
@@ -0,0 +1,24 @@
...
+connections. You can control this value with the ``maxsize`` parameter (from
+`aiopg.create_pool()`__)(see `discussions-pool-size`)
+
+.. __: https://aiopg.readthedocs.io/en/stable/core.html#aiopg.create_pool
+
+Disabling the ``LISTEN/NOTIFY`` feature (see `discussion-general`) will use one less
+connection per worker::
+
+    app = procrastinate.App(worker_defaults={"listen_notify": False})
+    # or when launchin a worker:
```
elemoine, line 14:
```suggestion
    # or when launching a worker:
```

docs/discussions.rst
```diff
@@ -208,6 +210,13 @@ Procrastinate will use use one connection to listen to server-side ``NOTIFY`` ca
 :ref:`discussion-general`). The rest of the pool is used for :term:`sub-workers
 <Sub-worker>`.
 
+.. warning::
+
+    Setting ``maxsize`` to ``0`` will actually disable the maximum. Setting it to
```
elemoine, line 215: Setting `maxsize` to `0` doesn't really work in aiopg. See https://github.com/aio-libs/aiopg/issues/311. In particular it quickly leads to a `TypeError` when setting `minsize` to `0` as well:
```
>>> pool = await aiopg.create_pool(dsn="", minsize=0, maxsize=0)
>>> await pool._fill_free_pool(True)
  File "/usr/lib64/python3.8/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/lib64/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/home/elemoine/.virtualenvs/procrastinate/lib/python3.8/site-packages/aiopg/pool.py", line 208, in _fill_free_pool
    if override_min and self.size < self.maxsize:
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

And I'm pretty sure `maxsize` set to `0` will eventually cause an error even when `minsize` is not `0`.

So `maxsize=0` is a dangerous setting, and I am not sure we want to advertise it in the Procrastinate documentation.

procrastinate/aiopg_connector.py
```diff
@@ -87,9 +87,8 @@ def __init__(
             instead of standard cursor. There is no identified use case for changing
             this.
         maxsize : ``int``
-            Passed to aiopg. Cannot be lower than 2, otherwise worker won't be
-            functioning normally (one connection for listen/notify, one for executing
-            tasks).
+            Passed to aiopg. If value is 1, then listen/notify feature will be
+            deactivated.
```
elemoine, line 91:
```suggestion
            disabled.
```

procrastinate/aiopg_connector.py
```diff
@@ -221,10 +218,18 @@ def _make_dynamic_query(self, query: str, **identifiers: str) -> str:
...
     async def listen_notify(
         self, event: asyncio.Event, channels: Iterable[str]
-    ) -> NoReturn:
+    ) -> None:
         pool = await self._get_pool()
         # We need to acquire a dedicated connection, and use the listen
         # query
+        if pool.maxsize == 1:
+            logger.warning(
+                "Listen/Notify capabilities disabled because maximum pull size"
```
elemoine, line 227:
```suggestion
                "Listen/Notify capabilities disabled because maximum pool size"
```

procrastinate/aiopg_connector.py
```diff
@@ -221,10 +218,18 @@ def _make_dynamic_query(self, query: str, **identifiers: str) -> str:
...
+                "is set to 1",
+                extra={"action": "listen_pool_too_small"},
```
elemoine, line 229: Why `listen_pool_too_small` here? Do you mean `pool_too_small` instead?

procrastinate/app.py
```diff
@@ -249,6 +249,10 @@ async def run_worker_async(self, **kwargs) -> None:
             Raising this parameter can lower the rate of workers making queries to the
             database for requesting jobs.
             (defaults to 5.0)
+        listen_notify : ``bool``
+            If ``True``, worker will dedicate a connection from the pool to listening
+            for database event, notifying of newly available jobs. If ``False``, workers
```
elemoine, line 254:
```suggestion
            to database events, notifying of newly available jobs. If ``False``, workers
```

procrastinate/app.py
```diff
@@ -249,6 +249,10 @@ async def run_worker_async(self, **kwargs) -> None:
...
+            will just poll the database regularily (see ``timeout``). (defaults to True)
```
elemoine, line 255: s/regularily/regularly ?

elemoine: And you may want to use `periodically` here, which you use elsewhere. Up to you.

elemoine:
I've added a few comments, but it looks good to me.

procrastinate/aiopg_connector.py
ewjoachim: ok, but I want toconvey the idea that it's about listen/notify. An idea ?

ewjoachim:
Done !

procrastinate/aiopg_connector.py
elemoine: pool_too_small_for_listen_notify ?

procrastinate/aiopg_connector.py
```diff
@@ -221,10 +218,18 @@ def _make_dynamic_query(self, query: str, **identifiers: str) -> str:
...
-    ) -> NoReturn:
+    ) -> None:
         pool = await self._get_pool()
         # We need to acquire a dedicated connection, and use the listen
         # query
+        if pool.maxsize == 1:
+            logger.warning(
+                "Listen/Notify capabilities disabled because maximum pool size"
+                "is set to 1",
+                extra={"action": "listen_notify_disabled"},
```
elemoine, line 229: Works for me too!

PR #259, opened 2020-06-17
Retry on "server closed connection unexpectedly" errors

### LINKED ISSUE #255: Handle closed connections gracefully
On a defer operation, if the db connection obtained from the pool is actually closed (e.g. because of a call to `pg_terminate_backend` on the Postgres side), the defer operation will fail with this error:

```
  File ".../lib/python3.8/site-packages/aiopg/connection.py", line 106, in _ready                                                                        
    state = self._conn.poll()                                                                                                                                                                  
psycopg2.OperationalError: server closed the connection unexpectedly                                                                                                                           
        This probably means the server terminated abnormally                                                                                                                                   
        before or while processing the request.
```

The next defer operation will succeed though, probably because aiopg didn't add the (closed) connection that caused the exception back to the pool.

I am wondering if Procrastinate should detect these cases and raise a specific exception, to give the application an opportunity to handle this as it sees fit.

A similar problem may exist on the worker side as well.

elemoine:
Closes #255.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -20,8 +20,9 @@ CoroutineFunction = Callable[..., Coroutine]
 
 def wrap_exceptions(coro: CoroutineFunction) -> CoroutineFunction:
     """
-    Wrap psycopg2 and aiopg errors as connector exceptions
-    This decorator is expected to be used on coroutine functions only
+    Wrap psycopg2 and aiopg errors as connector exceptions.
+
+    This decorator is expected to be used on coroutine functions only.
     """
 
     @functools.wraps(coro)
@@ -32,6 +33,45 @@ def wrap_exceptions(coro: CoroutineFunction) -> CoroutineFunction:
             raise exceptions.UniqueViolation(constraint_name=exc.diag.constraint_name)
         except psycopg2.Error as exc:
             raise exceptions.ConnectorException from exc
+
+    # Attaching a custom attribute to ease testability and make the
+    # decorator more introspectable
+    wrapped._exceptions_wrapped = True  # type: ignore
+    return wrapped
+
+
+def wrap_query_exceptions(coro: CoroutineFunction) -> CoroutineFunction:
+    """
+    Detect aiopg OperationalError's with a "server closed the connection unexpectedly"
+    message and retry a number of times.
+
+    This is to handle the case where the database connection (obtained from the pool)
+    was actually closed by the server. In this case, aiopg raises an OperationalError
+    with a "server closed the connection unexpectedly" message (and no pgcode) when the
+    connection is used for issuing a query. What we do is retry when an OperationalError
+    is raised, and until max_tries is reached.
+
+    max_tries is set to the max pool size + 1 (maxsize + 1) to handle the case where all
+    the connections we have in the pool were closed by the server.
+    """
+
+    @functools.wraps(coro)
+    async def wrapped(*args, **kwargs):
+        final_exc = None
+        max_tries = (
+            args[0]._pool.maxsize + 1 if args and getattr(args[0], "_pool", None) else 1
+        )
+        for _ in range(max_tries):
+            try:
+                return await coro(*args, **kwargs)
+            except psycopg2.errors.OperationalError as exc:
+                if "server closed the connection unexpectedly" in str(exc):
+                    final_exc = exc
+                    continue
+                raise exc
+        raise exceptions.ConnectorException(
+            "Could not get a valid connection after {} tries".format(max_tries)
+        ) from final_exc
 
     # Attaching a custom attribute to ease testability and make the
     # decorator more introspectable
@@ -178,12 +218,14 @@ class AiopgConnector(connector.BaseAsyncConnector):
     # a pool or from a connection
 
     @wrap_exceptions
+    @wrap_query_exceptions
     async def execute_query_async(self, query: str, **arguments: Any) -> None:
         pool = await self._get_pool()
         with await pool.cursor() as cursor:
             await cursor.execute(query, self._wrap_json(arguments))
 
     @wrap_exceptions
+    @wrap_query_exceptions
     async def _execute_query_connection(
         self, query: str, connection: aiopg.Connection, **arguments: Any,
     ) -> None:
@@ -191,6 +233,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
             await cursor.execute(query, self._wrap_json(arguments))
 
     @wrap_exceptions
+    @wrap_query_exceptions
     async def execute_query_one_async(
         self, query: str, **arguments: Any
     ) -> Dict[str, Any]:
@@ -201,6 +244,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
             return await cursor.fetchone()
 
     @wrap_exceptions
+    @wrap_query_exceptions
     async def execute_query_all_async(
         self, query: str, **arguments: Any
     ) -> List[Dict[str, Any]]:
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -73,9 +73,6 @@ def wrap_query_exceptions(coro: CoroutineFunction) -> CoroutineFunction:
             "Could not get a valid connection after {} tries".format(max_tries)
         ) from final_exc
 
-    # Attaching a custom attribute to ease testability and make the
-    # decorator more introspectable
-    wrapped._exceptions_wrapped = True  # type: ignore
     return wrapped
 
 
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -58,9 +58,10 @@ def wrap_query_exceptions(coro: CoroutineFunction) -> CoroutineFunction:
     @functools.wraps(coro)
     async def wrapped(*args, **kwargs):
         final_exc = None
-        max_tries = (
-            args[0]._pool.maxsize + 1 if args and getattr(args[0], "_pool", None) else 1
-        )
+        try:
+            max_tries = args[0]._pool.maxsize + 1
+        except Exception:
+            max_tries = 1
         for _ in range(max_tries):
             try:
                 return await coro(*args, **kwargs)
--- a/procrastinate/psycopg2_connector.py
+++ b/procrastinate/psycopg2_connector.py
@@ -34,6 +34,45 @@ def wrap_exceptions(func: Callable) -> Callable:
     return wrapped
 
 
+def wrap_query_exceptions(func: Callable) -> Callable:
+    """
+    Detect psycopg2 InterfaceError's with a "connection already closed" message and
+    retry a number of times.
+
+    This is to handle the case where the database connection (obtained from the pool)
+    was actually closed by the server. In this case, pyscopg2 raises an InterfaceError
+    with a "connection already closed" message (and no pgcode) when the connection is
+    used for issuing a query. What we do is retry when an InterfaceError is raised, and
+    until max_tries is reached.
+
+    max_tries is set to the max pool size + 1 (maxconn + 1) to handle the case where
+    all the connections we have in the pool were closed by the server.
+    """
+
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        final_exc = None
+        max_tries = (
+            args[0]._pool.maxconn + 1 if args and getattr(args[0], "_pool", None) else 1
+        )
+        for _ in range(max_tries):
+            try:
+                return func(*args, **kwargs)
+            except psycopg2.errors.InterfaceError as exc:
+                if "connection already closed" in str(exc):
+                    final_exc = exc
+                    continue
+                raise exc
+        raise exceptions.ConnectorException(
+            "Could not get a valid connection after {} tries".format(max_tries)
+        ) from final_exc
+
+    # Attaching a custom attribute to ease testability and make the
+    # decorator more introspectable
+    wrapped._exceptions_wrapped = True  # type: ignore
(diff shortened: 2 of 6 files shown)


elemoine:
Now with tests.

procrastinate/aiopg_connector.py
```diff
@@ -20,18 +20,37 @@
...
     @functools.wraps(coro)
     async def wrapped(*args, **kwargs):
-        try:
-            return await coro(*args, **kwargs)
-        except psycopg2.errors.UniqueViolation as exc:
-            raise exceptions.UniqueViolation(constraint_name=exc.diag.constraint_name)
-        except psycopg2.Error as exc:
-            raise exceptions.ConnectorException from exc
+        cnt = 0
+        max_tries = (
```
ewjoachim, line 33: This deserves the hell of a comment :D
- Why are we doing this maxsize times ?
- Why are we using a while loop and not a classic for loop ?

Also, we could use an [`itertools.count()`](https://docs.python.org/3.8/library/itertools.html#itertools.count)

And finally, I think we mentionned we wanted to put this in a differentdecorator ? This would help wrap_exceptions not to become too big.

ewjoachim:
What does it do with a Psycopg2 sync connector

procrastinate/aiopg_connector.py
elemoine: > And finally, I think we mentionned we wanted to put this in a differentdecorator ? This would help wrap_exceptions not to become too big.

Yep, but this led to code duplication, and it was not so clear where to use `wrap_exceptions` and `wrap_query_exceptions`.

ewjoachim: Why was there code duplication ?

I was imagining the query methods be decorated by the 2 decorators, if this is unclear.

elemoine: Right, I hadn't realized that you imagined two decorations for the query functions. That should work too!

elemoine: I added a commit that adds a `wrap_query_exceptions` decorator. I don't find this better than what my first commit provides, but maybe you had something else in mind.

elemoine: > This deserves the hell of a comment :D

You're right. I'll add them.

elemoine:
> What does it do with a Psycopg2 sync connector ?

You're right here to. We need to make the same sort of changes to the Psycopg2 connector.

procrastinate/aiopg_connector.py
```diff
@@ -39,6 +40,43 @@ async def wrapped(*args, **kwargs):
...
+        final_exc = None
+        cnt = 0
+        max_tries = (
+            args[0]._pool.maxsize + 1 if args and getattr(args[0], "_pool", None) else 1
+        )
+        while cnt < max_tries:
+            try:
+                return await coro(*args, **kwargs)
+            except exceptions.ConnectorException as exc:
+                exc_cause = exc.__cause__
```
ewjoachim, line 62: put the decorators the other way around, it will be easier :)

elemoine: Yeah, I had figured this out :-)

elemoine:
@ewjoachim I think I took all your comments into account. But happy to have more comments if you have any :)

procrastinate/aiopg_connector.py
ewjoachim: > I don't find this better than what my first commit provides, but maybe you had something else in mind.

Well you have split a complex function into 2 less complex more scoped functions. I think it's overally better.

procrastinate/aiopg_connector.py
```diff
@@ -39,6 +40,45 @@ async def wrapped(*args, **kwargs):
...
+
+    max_tries is set to the max pool size + 1 (maxsize + 1) to handle the case where all
+    the connections we have in the pool were closed by the server.
+    """
+
+    @functools.wraps(coro)
+    async def wrapped(*args, **kwargs):
+        final_exc = None
+        max_tries = (
+            args[0]._pool.maxsize + 1 if args and getattr(args[0], "_pool", None) else 1
```
ewjoachim, line 62: I wonder if this would be easier to understand as:
```
try:
	max_tries = args[0]._pool.maxsize + 1
except Exception:
	max_tries = 1
```

procrastinate/psycopg2_connector.py
```diff
@@ -34,6 +34,45 @@ def wrapped(*args, **kwargs):
...
+                    final_exc = exc
+                    continue
+                raise exc
+        raise exceptions.ConnectorException(
+            "Could not get a valid connection after {} tries".format(max_tries)
+        ) from final_exc
+
+    # Attaching a custom attribute to ease testability and make the
+    # decorator more introspectable
+    wrapped._exceptions_wrapped = True  # type: ignore
```
ewjoachim, line 72: is this useful ? Escpecially since it overwrites the attribute that the other function wrote...

tests/unit/test_aiopg_connector.py
```diff
@@ -52,6 +52,55 @@ async def corofunc(a, b):
     assert await corofunc(1, 2) == (1, 2)
 
 
+@pytest.mark.asyncio
+async def test_wrap_query_exceptions_reached_max_tries(mocker):
+    @aiopg_connector.wrap_query_exceptions
+    async def corofunc(connector):
+        raise psycopg2.errors.OperationalError(
+            "server closed the connection unexpectedly"
+        )
```
ewjoachim, line 61: Can we count that we were called 6 times ?

procrastinate/psycopg2_connector.py
```diff
@@ -34,6 +34,45 @@ def wrapped(*args, **kwargs):
...
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        final_exc = None
+        max_tries = (
+            args[0]._pool.maxconn + 1 if args and getattr(args[0], "_pool", None) else 1
+        )
+        for _ in range(max_tries):
+            try:
+                return func(*args, **kwargs)
+            except psycopg2.errors.InterfaceError as exc:
```
ewjoachim, line 61: I thought we had a AdminShutdown exception ?

tests/unit/test_psycopg2_connector.py
```diff
@@ -14,15 +13,54 @@ def func():
...
+
+
+def test_wrap_query_exceptions_success(mocker):
+    cnt = 0
+
+    @psycopg2_connector.wrap_query_exceptions
+    def func(connector, a, b):
+        nonlocal cnt
+        if cnt < 2:
+            cnt += 1
```
ewjoachim, line 56: A more natural way to do this is to declare a list outside the function, and append things inside.

procrastinate/psycopg2_connector.py
elemoine: This is what we get:

```
  File "/home/elemoine/src/procrastinate/procrastinate/psycopg2_connector.py", line 168, in _connection
    yield connection
  File "/home/elemoine/src/procrastinate/procrastinate/psycopg2_connector.py", line 183, in execute_query_one
    cursor.execute(query, self._wrap_json(arguments))
  File "/home/elemoine/.virtualenvs/procrastinate/lib/python3.8/site-packages/psycopg2/extras.py", line 248, in execute
    return super(RealDictCursor, self).execute(query, vars)
psycopg2.errors.AdminShutdown: terminating connection due to administrator command
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.


During handling of the above exception, another exception occurred:

  File "/home/elemoine/src/procrastinate/procrastinate/psycopg2_connector.py", line 25, in wrapped
    return func(*args, **kwargs)
  File "/home/elemoine/src/procrastinate/procrastinate/psycopg2_connector.py", line 184, in execute_query_one
    return cursor.fetchone()
  File "/usr/lib64/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/elemoine/src/procrastinate/procrastinate/psycopg2_connector.py", line 168, in _connection
    yield connection
psycopg2.InterfaceError: connection already closed

  File "<stdin>", line 1, in <module>
  File "/home/elemoine/src/procrastinate/procrastinate/tasks.py", line 112, in defer
    return self.configure().defer(**task_kwargs)
  File "/home/elemoine/src/procrastinate/procrastinate/jobs.py", line 151, in defer
    id = self.job_store.defer_job(job=job)
  File "/home/elemoine/src/procrastinate/procrastinate/store.py", line 35, in defer_job
    result = self.connector.execute_query_one(
  File "/home/elemoine/src/procrastinate/procrastinate/psycopg2_connector.py", line 29, in wrapped
    raise exceptions.ConnectorException from exc
procrastinate.exceptions.ConnectorException: 
    Database error.
```

tests/unit/test_aiopg_connector.py
elemoine: Done.

tests/unit/test_psycopg2_connector.py
elemoine: Done.

procrastinate/psycopg2_connector.py
elemoine: Removed.

procrastinate/aiopg_connector.py
elemoine: I have no preference.

procrastinate/psycopg2_connector.py
elemoine: Ok, I think this hides something. Back to the black board…

elemoine: Now handling AdminShutdown exceptions rather than InterfaceError. This made me rewrite the `_connection` context manager function to avoid rolling back the connection or putting it back to the pool when an AdminShutdown error occurs. Rolling back the connection or putting it back to the pool is what makes psycopg2 raise an InterfaceError.

This was more complex than I initially expected.

procrastinate/psycopg2_connector.py
```diff
@@ -124,26 +155,42 @@ def _wrap_json(self, arguments: Dict[str, Any]):
...
-        finally:
+            except psycopg2.errors.AdminShutdown:
+                raise
+            except Exception:
+                connection.rollback()
+                raise
+            else:
+                connection.commit()
+        except psycopg2.errors.AdminShutdown:
+            raise
```
ewjoachim, line 173: I think those 2 lines do nothing. If we removed them, an AdminShutdown would also bubble up.

procrastinate/psycopg2_connector.py
```diff
@@ -124,26 +155,42 @@ def _wrap_json(self, arguments: Dict[str, Any]):
...
+            try:
                 yield connection
-        finally:
+            except psycopg2.errors.AdminShutdown:
+                raise
+            except Exception:
+                connection.rollback()
+                raise
+            else:
+                connection.commit()
```
ewjoachim, line 171: If we put the `putconn` here, then we don't need the outer `try` anymore, no ?

procrastinate/psycopg2_connector.py
```diff
@@ -124,26 +155,42 @@ def _wrap_json(self, arguments: Dict[str, Any]):
...
+        # rollback the connection or put the connection back to the pool as
+        # this will cause a psycopg2.InterfaceError exception
         try:
-            with self._pool.getconn() as connection:
+            connection = self._pool.getconn()
+            try:
                 yield connection
-        finally:
+            except psycopg2.errors.AdminShutdown:
+                raise
```
ewjoachim, line 166: So IIUC, the whole thing is that in case of AdminShutdown, we must avoid rolling back, right ?

ewjoachim:
We're getting there \o/ Thanks for your continued efforts.

procrastinate/aiopg_connector.py
ewjoachim: then I vote for the change :)

procrastinate/psycopg2_connector.py
elemoine: Yep, see the command above. And we also avoid putting the connection back to the pool.

ewjoachim: Ah but you don't do this for `except Exception` too, is it voluntary ?

How do you tell psycopg2 to stop keeping track of those connections ?

procrastinate/aiopg_connector.py
```diff
@@ -39,6 +40,43 @@ async def wrapped(*args, **kwargs):
...
+    message and retry a number of times.
+
+    This is to handle the case where the database connection (obtained from the pool)
+    was actually closed by the server. In this case, aiopg raises an OperationalError
+    with a "server closed the connection unexpectedly" message (and no pgcode) when the
+    connection is used for issuing a query. What we do is retry when an OperationalError
+    is raised, and until max_tries is reached.
+
+    max_tries is set to the max pool size + 1 (maxsize + 1) to handle the case where all
+    the connections we have in the pool were closed by the server.
```
ewjoachim, line 55:
```suggestion
    The number of retries is set to the pool maximum size plus one, to handle the
    case where the connections we have in the pool were closed all on the server side.
```

procrastinate/psycopg2_connector.py
elemoine: Yep, and I also need to call `putconn` after rollback.

Changed.

procrastinate/psycopg2_connector.py
elemoine: > Ah but you don't do this for `except Exception` too, is it voluntary ?

I now call `putconn` after rollback.

elemoine:
I am done updating the PR. Ready for final review.

PR #264, opened 2020-06-22
Add task aliases

### LINKED ISSUE #193: Don't lose tasks when changing their path
By default, the name of a task is the full dotted path of the decorated function.

If we have to refactor our code and move that function, the name will change and the jobs in queue won't be able to be executed after the deployment of the new code.

Could we add the possibility for a function to have multiple names (or aliases) in order to  implement safe code upgrades ?

tmartinfr:
Closes #193

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -1,6 +1,6 @@
 import functools
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, Optional, Set
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Set
 
 from procrastinate import admin
 from procrastinate import connector as connector_module
@@ -96,6 +96,7 @@ class App:
         *,
         queue: str = jobs.DEFAULT_QUEUE,
         name: Optional[str] = None,
+        aliases: Optional[List[str]] = None,
         retry: retry_module.RetryValue = False,
         pass_context: bool = False,
     ) -> Any:
@@ -129,6 +130,11 @@ class App:
             if the function is nested or dynamically defined, it is important to give
             it a unique name, and to make sure the module that defines this function
             is listed in the ``import_paths`` of the `procrastinate.App`.
+        aliases:
+            Additional names for the task.
+            The main use case is to gracefully rename tasks by moving the old
+            name to aliases (these tasks can have been scheduled in a distant
+            future, so the aliases can remain for a long time).
         retry :
             Details how to auto-retry the task if it fails. Can be:
 
@@ -151,6 +157,7 @@ class App:
                 app=self,
                 queue=queue,
                 name=name,
+                aliases=aliases,
                 retry=retry,
                 pass_context=pass_context,
             )
@@ -165,6 +172,8 @@ class App:
 
     def _register(self, task: "tasks.Task") -> None:
         self.tasks[task.name] = task
+        for alias in task.aliases:
+            self.tasks[alias] = task
         queue = task.queue
         if queue not in self.queues:
             logger.debug(
--- a/procrastinate/tasks.py
+++ b/procrastinate/tasks.py
@@ -1,6 +1,6 @@
 import datetime
 import logging
-from typing import Any, Callable, Dict, Optional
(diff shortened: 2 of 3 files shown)


procrastinate/tasks.py
```diff
@@ -89,6 +92,10 @@ def __init__(
         self.name: str = name if name else self.full_path
         self.pass_context = pass_context
 
+        if aliases is not None:
+            assert isinstance(aliases, list)
```
ewjoachim, line 96: Hm ?
- Why do you request an "Iterable" and not a list, if you want to make sure it's a list ?
- Why not accept any iterable and just make it a list ?

tmartinfr: The first point is a mistake, initially I wanted to accept any iterable

tmartinfr: For the second point, if we do that, then we should ensure the argument is not a string, right ?

ewjoachim: Yes, if we consider we can accept "strings or lists" then I'd say we should do:
```python
if isinstance(value, str):
    value = [value]   # transform a single string into a list of strings

value = list(value)  # transform any iterable into a list
```

tmartinfr: This approch is fine to me, but I've to tell that my initial choice of restricting only to list was driven by the fear someone could pass an iterable object which was not a list, tuple, or str, and it could lead to an unexpected behaviour. That's why I was prefering a defensive check. But maybe it's not really relevant, what is your opinion about this ?

ewjoachim: In pythonic code, I'd say it's expected that we could receive any iterable. If we need a list internally, it's our job to make a list from the iterable.

That being said, it could also be ok to explicitely request a list, but in this case, I would document it (through type hints, which will find their way to the documentation) and... I wouldn't necessarily check the type at runtime: we'll be using duck typing, if someone give us something that "walks and quacks like a list", then for all intended purposes it's a list, we have no reason to tell them it's wrong.
We're in a dynamic language, we could start doubting every argument type of every function. If we document that we receive a single type, for me we can trust the user to send the type.

But the best it to request an Iterable, and then make it a list. That way, no questions asked.

procrastinate/app.py
```diff
@@ -151,6 +157,7 @@ def _wrap(func: Callable[..., "tasks.Task"]):
                 app=self,
                 queue=queue,
                 name=name,
+                aliases=aliases,
```
elemoine, line 160: Do we really need to store the aliases in the Task?

ewjoachim: Good remark. This is most probably not needed.

tmartinfr: During a previous discussion, we thought it was a good idea to have a Task access to its aliases. But I have no problem to remove it if you think it's useless.

tmartinfr: Actually it's also what allows to register the task aliases in `_register`. How should we handle it if we don't store aliases in Task ?

elemoine: You can pass the aliases to the `_register` function, no?

tmartinfr: As discussed with @elemoine, my point is that it could be useful for later introspection and good for staying consistent with the fact we already store the task's name.
That's not a blocking point for him, so I'll keep like this :) Thanks @elemoine for your input.

procrastinate/tasks.py
tmartinfr: I'll keep the `List` approach as it's not blocking for you. Thanks for your input @ewjoachim.

procrastinate/app.py
elemoine: I also said that I am not a big fan with storing the aliases in the `Task` object for the sole purpose of passing them to the `_register` function. But it may be useful to have the aliases in the `Task` object for other purposes. Anyway, not a blocker for me.

ewjoachim:
We could have added an "howto". That can be done in a subsequent PR.

PR #270, opened 2020-06-26
WIP: added app context manager (open and close methods)

### LINKED ISSUE #265: App Context manager
So as seen with #262 , when possible, it would be better it the procrastinate application was explicitely opened and closed. Opening would mean start the connector (which would create the pool), closing means closing the connector, which would close the pool.

It would be nice if `enter` and `exit` methods were available out of the context manager too, because there are somecases where it's complicated to have one's code organized with pre-launch and post-stop actions in the same function.

thomasperrot:
Closes #265

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)
- [X] Had a good time contributing? :+1: 

### DIFF
--- a/procrastinate/aiopg_connector.py
+++ b/procrastinate/aiopg_connector.py
@@ -87,8 +87,11 @@ class AiopgConnector(connector.BaseAsyncConnector):
     ):
         """
         Create a PostgreSQL connector using aiopg. The connector uses an ``aiopg.Pool``,
-        which is either created automatically upon first use, or set through the
-        `AiopgConnector.set_pool` method.
+        which is either created by `App.open_async`.
+
+        The pool connection parameters can be provided in this method. Alternatively,
+        an already existing ``aiopg.Pool`` can be provided in the
+        `App.open_async`, via the ``pool`` parameter.
 
         All other arguments than ``json_dumps`` and ``json_loads`` are passed to
         :py:func:`aiopg.create_pool` (see aiopg documentation__), with default values
@@ -106,7 +109,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
             The JSON loads function to use for deserializing job arguments. Defaults
             to the function used by psycopg2. See the `psycopg2 doc`_. Unused if the
             pool is externally created and set into the connector through the
-            `AiopgConnector.set_pool` method.
+            :py:meth:`AiopgConnector.open_async` method.
         dsn : ``Optional[str]``
             Passed to aiopg. Default is "" instead of None, which means if no argument
             is passed, it will connect to localhost:5432 instead of a Unix-domain
@@ -166,12 +169,31 @@ class AiopgConnector(connector.BaseAsyncConnector):
         final_args.update(pool_args)
         return final_args
 
+    @property
+    def pool(self) -> aiopg.Pool:
+        if self._pool is None:  # Set by open_async
+            raise exceptions.AppNotOpen
+        return self._pool
+
+    async def open_async(self, pool: Optional[aiopg.Pool] = None) -> None:
+        if pool:
+            self._pool_externally_set = True
+            self._pool = pool
+        else:
+            self._pool = await self._create_pool(self._pool_args)
+
+    @staticmethod
+    @wrap_exceptions
+    async def _create_pool(pool_args: Dict[str, Any]) -> aiopg.Pool:
+        return await aiopg.create_pool(**pool_args)
+
     @wrap_exceptions
     async def close_async(self) -> None:
         """
         Close the pool and awaits all connections to be released.
         """
-        if self._pool:
+
+        if self._pool and not self._pool_externally_set:
             self._pool.close()
             await self._pool.wait_closed()
             self._pool = None
@@ -195,31 +217,6 @@ class AiopgConnector(connector.BaseAsyncConnector):
             for key, value in arguments.items()
         }
 
-    @staticmethod
-    @wrap_exceptions
-    async def _create_pool(pool_args: Dict[str, Any]) -> aiopg.Pool:
-        return await aiopg.create_pool(**pool_args)
-
-    def set_pool(self, pool: aiopg.Pool, *, _external: bool = True) -> None:
-        """
-        Set the connection pool. Raises an exception if the pool is already set.
-        """
-
-        if self._pool:
-            raise exceptions.PoolAlreadySet
-        self._pool_externally_set = _external
-        self._pool = pool
-
-    async def _get_pool(self) -> aiopg.Pool:
-        if self._pool:
-            return self._pool
-        if not self._lock:
-            self._lock = asyncio.Lock()
-        async with self._lock:
-            if not self._pool:
-                self.set_pool(await self._create_pool(self._pool_args), _external=False)
-        return self._pool
-
     # Pools and single connections do not exactly share their cursor API:
     # - connection.cursor() is an async context manager (async with)
     # - pool.cursor() is a coroutine returning a sync context manage (with await)
@@ -229,8 +226,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
     @wrap_exceptions
     @wrap_query_exceptions
     async def execute_query_async(self, query: str, **arguments: Any) -> None:
-        pool = await self._get_pool()
-        with await pool.cursor() as cursor:
+        with await self.pool.cursor() as cursor:
             await cursor.execute(query, self._wrap_json(arguments))
 
     @wrap_exceptions
@@ -246,8 +242,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
     async def execute_query_one_async(
         self, query: str, **arguments: Any
     ) -> Dict[str, Any]:
-        pool = await self._get_pool()
-        with await pool.cursor() as cursor:
+        with await self.pool.cursor() as cursor:
             await cursor.execute(query, self._wrap_json(arguments))
 
             return await cursor.fetchone()
@@ -257,8 +252,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
     async def execute_query_all_async(
         self, query: str, **arguments: Any
     ) -> List[Dict[str, Any]]:
-        pool = await self._get_pool()
-        with await pool.cursor() as cursor:
+        with await self.pool.cursor() as cursor:
             await cursor.execute(query, self._wrap_json(arguments))
 
             return await cursor.fetchall()
@@ -275,10 +269,9 @@ class AiopgConnector(connector.BaseAsyncConnector):
     async def listen_notify(
         self, event: asyncio.Event, channels: Iterable[str]
     ) -> None:
-        pool = await self._get_pool()
         # We need to acquire a dedicated connection, and use the listen
         # query
-        if pool.maxsize == 1:
+        if self.pool.maxsize == 1:
             logger.warning(
                 "Listen/Notify capabilities disabled because maximum pool size"
                 "is set to 1",
@@ -287,7 +280,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
             return
 
         while True:
-            async with pool.acquire() as connection:
+            async with self.pool.acquire() as connection:
                 for channel_name in channels:
                     await self._execute_query_connection(
                         connection=connection,
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -14,7 +14,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-@utils.add_sync_api
 class App:
     """
     The App is the main entry point for procrastinate integration.
@@ -270,3 +269,30 @@ class App:
     @property
     def admin(self) -> admin.Admin:
         return admin.Admin(connector=self.connector)
+
+    def open(self, pool: Optional[connector_module.Pool] = None) -> None:
+        self.connector.open(pool)
+
+    def close(self) -> None:
+        self.connector.close()
+
+    async def open_async(self, pool: Optional[connector_module.Pool] = None) -> None:
+        await self.connector.open_async(pool)
+
+    async def close_async(self) -> None:
+        await self.connector.close_async()
+
+    def __enter__(self, pool: Optional[connector_module.Pool] = None) -> None:
+        self.open(pool)
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.close()
+
+    async def __aenter__(self, pool: Optional[connector_module.Pool] = None) -> None:
+        await self.open_async(pool)
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        await self.close_async()
+
+
+utils.add_method_sync_api(cls=App, method_name="run_worker_async")
--- a/procrastinate/connector.py
+++ b/procrastinate/connector.py
@@ -6,9 +6,15 @@ from procrastinate import exceptions, utils
 QUEUEING_LOCK_CONSTRAINT = "procrastinate_jobs_queueing_lock_idx"
 
 
+Pool = Any
+
+
 class BaseConnector:
     json_dumps: Optional[Callable] = None
     json_loads: Optional[Callable] = None
+
+    def open(self, pool: Optional[Pool] = None) -> None:
+        raise NotImplementedError
 
     def close(self) -> None:
         raise NotImplementedError
@@ -21,6 +27,9 @@ class BaseConnector:
 
     def execute_query_all(self, query: str, **arguments: Any) -> List[Dict[str, Any]]:
         raise NotImplementedError
+
+    async def open_async(self, pool: Optional[Pool] = None) -> None:
+        raise exceptions.SyncConnectorConfigurationError
 
     async def close_async(self) -> None:
         raise exceptions.SyncConnectorConfigurationError
@@ -49,6 +58,9 @@ class BaseConnector:
 
 @utils.add_sync_api
 class BaseAsyncConnector(BaseConnector):
+    async def open_async(self, pool: Optional[Pool] = None) -> None:
+        raise NotImplementedError
+
     async def close_async(self) -> None:
         raise NotImplementedError
 
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -40,10 +40,14 @@ class JobRetry(ProcrastinateException):
         super().__init__()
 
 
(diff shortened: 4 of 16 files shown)


procrastinate/aiopg_connector.py
```diff
@@ -106,7 +109,7 @@ def __init__(
             The JSON loads function to use for deserializing job arguments. Defaults
             to the function used by psycopg2. See the `psycopg2 doc`_. Unused if the
             pool is externally created and set into the connector through the
-            `AiopgConnector.set_pool` method.
+            :py:meth:`AiopgConnector.open_async` method.
```
ewjoachim, line 112:
```suggestion
            `App.open_async` method.
```

procrastinate/aiopg_connector.py
```diff
@@ -87,8 +87,11 @@ def __init__(
     ):
         """
         Create a PostgreSQL connector using aiopg. The connector uses an ``aiopg.Pool``,
-        which is either created automatically upon first use, or set through the
-        `AiopgConnector.set_pool` method.
+        which is either created by `App.open_async`.
```
CorBott, line 90: Remove "either" if there is no option.... or something is missing

procrastinate/aiopg_connector.py
```diff
@@ -87,8 +87,11 @@ def __init__(
...
+
+        The pool connection parameters can be provided in this method. Alternatively,
```
elemoine, line 92: It's the constructor, I wouldn't say "in this method" here.

procrastinate/aiopg_connector.py
```diff
@@ -166,12 +169,31 @@ async def on_connect(connection):
         final_args.update(pool_args)
         return final_args
 
+    @property
+    def pool(self) -> aiopg.Pool:
+        if self._pool is None:  # Set by open_async
+            raise exceptions.AppNotOpen
```
elemoine, line 175: It feels strange to me to raise an `AppNotOpen` exception from `AiopgConnector`.

procrastinate/app.py
```diff
@@ -270,3 +269,30 @@ def health_check_runner(self) -> healthchecks.HealthCheckRunner:
...
+    def close(self) -> None:
+        self.connector.close()
+
+    async def open_async(self, pool: Optional[connector_module.Pool] = None) -> None:
+        await self.connector.open_async(pool)
+
+    async def close_async(self) -> None:
+        await self.connector.close_async()
+
+    def __enter__(self, pool: Optional[connector_module.Pool] = None) -> None:
```
elemoine, line 285: `__enter__` normally takes a single argument, `self`. Why having an additional argument, `pool`, here?

procrastinate/connector.py
```diff
@@ -6,10 +6,16 @@
 QUEUEING_LOCK_CONSTRAINT = "procrastinate_jobs_queueing_lock_idx"
 
 
+Pool = Any
```
elemoine, line 9: I think this deserves a comment.

procrastinate/exceptions.py
```diff
@@ -40,10 +40,14 @@ def __init__(self, scheduled_at: datetime.datetime):
...
+class AppNotOpen(ProcrastinateException):
     """
-    connector.set_pool() was called but the pool already had a set.
-    Changing the pool of a connector is not permitted.
+    App was not open. Procrastinate App needs to be opened using:
+
+    - ``App.open``,
+    - ``App.open_async``,
+    - ``with App():``,
+    - ``async with App():``,
```
elemoine, line 50:
```suggestion
    - ``async with App():``.
```

elemoine:
As I understand it these are the various use-cases covered with new API:

1. AiopgConnector in async mode with a pool created internally
   ```py
   app = procrastinate.App(connector=procrastinate.AiopgConnector())
   async with app:
       # do things…
   ```
   And you can use `await app.open_async()` and `await app.close_async()` if using the context manager is not an option.

2. AiopgConnector in async mode with a pool created externally
   ```py
   app = procrastinate.App(connector=procrastinate.AiopgConnector())
   async with aiopg.create_pool() as pool:
       await app.connector.open_async(pool)
       # do things…
   ```
   And you can use `await aiopg.create_pool()` and `pool.close(); await pool.wait_closed()` if using the context manager is not an option.

3. AiopgConnector in sync mode with a pool created internally
   ```py
   app = procrastinate.App(connector=procrastinate.AiopgConnector())
   with app:
       # do things
   ```
   And you can use `app.open()` and `app.close()` if using the context manager is not an option.

4. AiopgConnector in sync mode with a pool created externally -> NOT COVERED (which makes sense)


5. Psycopg2Connector with a pool created internally
   ```py
   app = procrastinate.App(connector=procrastinate.Psycopg2Connector())
   with app:
       # do things…
   ```

6. Psycopg2Connector with a pool created externally
    ```py
   app = procrastinate.App(connector=procrastinate.Psycopg2Connector())
   pool = psycopg2.pool.ThreadedConnectionPool()
   app.open(pool)
   # do things…
   pool.closeall()
   ```

So far I don't see the need for `__enter__` and `__aenter__` dunders that take an extra argument, `pool`.

procrastinate/aiopg_connector.py
elemoine: I wouldn't mention `App.open_sync` in the `AiopgConnector` docstrings. I'd just say that the pool is created internally, or set into the connector, by calling `open_async`.

thomasperrot: Updated

procrastinate/aiopg_connector.py
thomasperrot: I changed for `The pool connection parameters can be provided here.`

procrastinate/app.py
thomasperrot: True, I noticed that this morning running unittests. This is pretty anoying, as we wanted our API to give the user the ability to specifiy the pool when entering the context (in `.enter` method for instance). We actually cannot do that. Maybe we should not allow the user specify the pool when entering the context. 

I will remove the user ability to provide the pool when entering context.

procrastinate/connector.py
thomasperrot: Following the previous comment, if we cannot specify the pool when entering context, this line is not needed anymore.

procrastinate/exceptions.py
thomasperrot: Updated.

procrastinate/app.py
elemoine: > This is pretty anoying, as we wanted our API to give the user the ability to specifiy the pool when entering the context (in `.enter` method for instance).

Why did you want that?

When the pool is created externally it makes no sense to use a Procrastinate context manager. You'd use that instead:

```py
app = procrastinate.App(connector=procrastinate.AiopgConnector())
async with aiopg.create_pool() as pool:
    await app.open_async(pool)
    …
```

No?

ewjoachim:
`connector.open_async` is supposedly not part of the external API.

That being said, you're right, `__enter__` and `__aenter__` do not take an additional argument. When using `with` and `async with`, these magic methods are called without argument (besides self).

procrastinate/app.py
ewjoachim: This is interesting, we should discuss it !

procrastinate/aiopg_connector.py
ewjoachim: I understand but I think it's the best solution. Raising something more connector-specific means we would have to expect and reraise it in the app and it feels awfully complex.

elemoine: Ok I understand, and it kinda makes sense :-)

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+The app must be opened at the beginning of the program to initiate the connection to
+the database, and eventually closed to properly terminate it.
+
+The app implements several ways to do this.
+
+Explicitly open and close the app
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The app can be opened using the ``open`` method (for sync case) or ``open_async``
+method (for async case). It eventually needs to be close, using the ``close`` method
```
elemoine, line 32:
```suggestion
method (for async case). It eventually needs to be closed, using the ``close`` method
```

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+
+Or with a context manager::
+
+   async with procrastinate.App(connector=connector).open_async() as app:
+       pass
+
+Open the app, and let the garbage collector terminate the connection
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The app does not necessarily needs to be closed at the end of your program. Indeed, the
```
elemoine, line 92:
```suggestion
The app does not necessarily need to be closed at the end of your program. Indeed, the
```

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+connections to the database will be automatically terminated when the database
+connectors are garbage collected. Thus, your program can work with only the following
```
elemoine, line 94:
```suggestion
connector is garbage collected. Thus, your program can work with only the following
```

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+statements, as we did in the quickstart::
```
elemoine, line 95: In the `README.rst` file `app.close()` is called, which might not be necessary either.

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+
+   app = procrastinate.App(connector=connector)
+   app.open()
+
+Note that the ``close`` method is not called in he above example.
```
elemoine, line 100:
```suggestion
Note that the ``close`` method is not called in the above example.
```

docs/quickstart.rst
```diff
@@ -47,6 +48,9 @@ The application will be the entry point for both:
 - Launching the worker that will consume the jobs created from those tasks.
 - Applying Database schema
 
+The ``App.open`` method is called to initiate the connection to the underlying
+database. It will be automatically terminated upon garbage collection.
```
elemoine, line 52: It is not completely correct to say that `App.open` initiates the connection to the underlying database. `App.open` just creates the connection pool.

docs/howto/app.rst
thomasperrot: It is not necessary indeed, but I think it adds clarity. If `app.close()` was not mentioned in the README.rst, the user might wonder why we do not close the connection.

docs/quickstart.rst
thomasperrot: True, and I made the same confusion the the app.rst file. I correct that in both :+1:

docs/howto/app.rst
thomasperrot: I remove `app.close()` to keep this example as light as possible.

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+For async case::
+
+    async with app.open_async():
+        pass
+
+
+Create and open the app in a single line
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both ``open`` and ``open_async`` return the current app when they are called or
```
elemoine, line 68: Strictly speaking `open_async` returns an awaitable, which `__await__` method returns the application. Do we want to be more accurate here in the documentation?

docs/howto/app.rst
elemoine: :+1:

docs/quickstart.rst
elemoine: :+1:

docs/howto/app.rst
```diff
@@ -16,3 +16,92 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+
+The app does not necessarily need to be closed at the end of your program. Indeed, the
+connections to the database will be automatically terminated when the database
+connector is garbage collected. Thus, your program can work with only the following
+statements, as we did in the readme and the quickstart::
+
+   app = procrastinate.App(connector=connector)
+   app.open()
+
+Note that the ``close`` method is not called in the above example.
```
elemoine, line 100: I think this sentence is superfluous.

elemoine:
I added two new comments, but this looks really good to me.

docs/howto/app.rst
thomasperrot: You are right. I tried to keep this as simple as possible. Is the following better ?
```suggestion
Both ``open`` and ``open_async`` return an ``AwaitableContext`` object that returns the current app when it is entered, or awaited
```
It is more accurate, but I am afraid it improves complexity too much.

docs/howto/app.rst
thomasperrot: Totally agree ! I change that :ok_hand:

tests/unit/test_aiopg_connector.py
```diff
@@ -68,18 +68,21 @@ async def corofunc(connector):
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize("pool_max_size", [5, "5"])
 @pytest.mark.parametrize(
     "exception_class", [Exception, psycopg2.errors.OperationalError]
 )
```
ewjoachim, line 74: I'm not sure I understand why you're doing that.

Do we really need to test the cartesian product of the pool sizes and the exception classes here ?

docs/howto/app.rst
ewjoachim: Yep, I think it's counter-productive to mention the AwaitableContext here. I'm comfortable with the original sentence.

docs/howto/app.rst
```diff
@@ -16,3 +16,90 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+
+For async case::
+
+    await app.open_async()
+    ...
+    await app.close_async()
+
+
+Open the app and automatically close it with a context manager
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ewjoachim, line 49:
```suggestion
Open and close the app with a context manager
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/howto/app.rst
```diff
@@ -16,3 +16,90 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+
+Or with a context manager::
+
+   async with procrastinate.App(connector=connector).open_async() as app:
+       pass
+
+Open the app, and let the garbage collector terminate the connection
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The app does not necessarily need to be closed at the end of your program. Indeed, the
```
ewjoachim, line 92:
```suggestion
The app does not necessarily need to be closed at the end of your program. The
```

docs/howto/app.rst
```diff
@@ -16,3 +16,90 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+connections to the database will be automatically terminated when the database
+connector is garbage collected. Thus, your program can work with only the following
+statements, as we did in the readme and the quickstart::
+
+   app = procrastinate.App(connector=connector)
+   app.open()
+
```
ewjoachim, line 99:
```suggestion

But remember: "explicit is better than implicit".
```

docs/howto/app.rst
```diff
@@ -16,3 +16,90 @@ In the first two cases, it's important to specify the location to your tasks usi
         connector=connector,
         import_paths=["dotted.path.to", "all.the.modules", "that.define.tasks"]
     )
+
```
ewjoachim, line 19: The title of the page is not very relevant anymore.

docs/howto/app.rst
```diff
@@ -16,3 +16,90 @@ In the first two cases, it's important to specify the location to your tasks usi
...
+Create and open the app in a single line
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both ``open`` and ``open_async`` return the current app when they are called or
+entered, which allows you to instantiate and open your app in a single line.
+
+For sync case::
+
+   app = procrastinate.App(connector=connector).open()
```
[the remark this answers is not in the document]
thomasperrot: Ok, I delete the three others

docs/howto/app.rst
thomasperrot: Do you mean changing 

> Find a correct place for your app

?

tests/unit/test_aiopg_connector.py
thomasperrot: Without this line, the test never pass in the following exception https://github.com/peopledoc/procrastinate/blob/master/procrastinate/aiopg_connector.py#L64.

You are right, we do not need to test the full cartesian product. Maybe we can reach that exception in an other test.

docs/howto/app.rst
ewjoachim: yes

tests/unit/test_aiopg_connector.py
thomasperrot: I put it in a more relevant test, see upgrade. I let you close this discussion if you are ok with the upgrade :)

ewjoachim: Yep, when possible, I prefer to have:
- the shortest tests
- tests that test the least amount of code
- the least amount of tests

Of course, there's a balance that needs to be struck between those 3, but adding 2 complex tests to cover a single missing line seems improvable :)

thomasperrot: Improved in an other test using parametrize. It is much cleaner.

docs/howto/app.rst
thomasperrot: I changed a few files in the doc to improve it. The open/close explanation has now its own dedicated file : `open_connection.rst`, which is the third step in the Howto : 
image

If this is good, we might be ready to merge :+1:

thomasperrot: I think it deserves its own file, as it not directly related to connector creation, nor App creation, and must be done before defining tasks.

PR #272, opened 2020-07-03
Add py.typed to package

### LINKED ISSUE #249: Add a py.typed (PEP-561)
https://www.python.org/dev/peps/pep-0561/

We're providing static typing, we should definitely expose it !

SBillion:
Closes #249 

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,8 @@ classifiers =
     License :: OSI Approved :: MIT License
 
 [options]
-zip_safe = True
+ # mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages
+zip_safe = False
 include_package_data = True
 packages = find:
 install_requires =
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,6 +4,7 @@ prune procrastinate_demo
 
 include LICENSE.rst
 include VERSION.txt
+include procrastinate/py.typed
 
 exclude .editorconfig
 exclude .readthedocs.yml


SBillion:
Had to set zip_safe to False or mypy won't find the installed package
https://mypy.readthedocs.io/en/latest/installed_packages.html?highlight=pep#making-pep-561-compatible-packages

setup.cfg
```diff
@@ -72,6 +72,9 @@ include =
 console_scripts =
     procrastinate = procrastinate.cli:main
 
+[options.package_data]
+procrastinate = py.typed
```
ewjoachim, line 76: Why do we need that with the MANIFEST.in ?

setup.cfg
```diff
@@ -17,7 +17,7 @@ classifiers =
     License :: OSI Approved :: MIT License
 
 [options]
-zip_safe = True
+zip_safe = False
```
ewjoachim, line 20: Would be nice to mention https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages

setup.cfg
SBillion: Yes you're right. I added it to MANIFEST.in after and forgot to remove here. Done

setup.cfg
SBillion: Done

PR #279, opened 2020-07-17
Use python-dateutil to parse date

EliotBerriot:
### Successful PR Checklist:

- [x] Tests
- [x] Documentation

## Rationale

`procrastinate` currently relies on `pendulum` to manipulate and parse dates, which is sometimes difficult to install (see https://github.com/sdispater/pendulum/issues/457 for instance).

As the needs of `procrastinate` in terms of date parsing and manipulation are quite light, and `python-dateutil` is already an indirect dependency of `procrastinate` (not sure which dependency relies on it though), I think it makes sense to replace `pendulum` with a couple of functions and calls to `python-dateutil`.

This PR does exactly that, and as far as I can tell coverage remains stable and tests are passing.

Let me know what you think about it :)

### DIFF
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -1,12 +1,12 @@
 import asyncio
 import contextlib
+import datetime
 import json
 import logging
 import os
 from typing import Any, Callable, Dict, Optional
 
 import click
-import pendulum
 
 import procrastinate
 from procrastinate import connector, exceptions, jobs, shell, types, utils, worker
@@ -283,20 +283,16 @@ def load_json_args(json_args: str, json_loads: Callable) -> types.JSONDict:
     return args
 
 
-# return type should not be Optional[str], but there's a typing bug in
-# Pendulum 2.1.0 (https://github.com/sdispater/pendulum/issues/450)
-def get_schedule_at(at: Optional[str]) -> Optional[pendulum.DateTime]:
+def get_schedule_at(at: Optional[str]) -> Optional[datetime.datetime]:
     if at is None:
         return None
 
     try:
-        datetime = pendulum.parse(at)
-        if not isinstance(datetime, pendulum.DateTime):
-            raise ValueError
-    except (pendulum.exceptions.ParserError, ValueError):
+        dt = utils.parse_datetime(at)
+    except (ValueError):
         raise click.BadOptionUsage("--at", f"Cannot parse datetime {at}")
 
-    return datetime
+    return dt
 
 
 def get_schedule_in(in_: Optional[int]) -> Optional[Dict[str, int]]:
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -289,7 +289,7 @@ def get_schedule_at(at: Optional[str]) -> Optional[datetime.datetime]:
 
     try:
         dt = utils.parse_datetime(at)
-    except (ValueError):
+    except ValueError:
         raise click.BadOptionUsage("--at", f"Cannot parse datetime {at}")
 
     return dt
--- a/procrastinate/utils.py
+++ b/procrastinate/utils.py
@@ -239,7 +239,7 @@ def task_context(awaitable: Awaitable, name: str):
 
 
 def utcnow() -> datetime.datetime:
-    return datetime.datetime.now(datetime.timezone.utc)
+    return datetime.datetime.now(tz=datetime.timezone.utc)
 
 
 def aware_datetime(
--- a/procrastinate/utils.py
+++ b/procrastinate/utils.py
@@ -242,26 +242,6 @@ def utcnow() -> datetime.datetime:
     return datetime.datetime.now(tz=datetime.timezone.utc)
 
 
-def aware_datetime(
-    year: int,
-    month: int,
-    day: int,
-    hour: int = 0,
-    minute: int = 0,
-    second: int = 0,
-    microsecond: int = 0,
-    tz_offset: Optional[int] = None,
-) -> datetime.datetime:
-    tzinfo = (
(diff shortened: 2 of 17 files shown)


EliotBerriot:
Fixed a linting issue

tests/unit/test_utils.py
```diff
@@ -280,3 +281,47 @@ async def foo():
 
     exc_logs = [r for r in caplog.records if r.action == "foo_error" and r.exc_info]
     assert len(exc_logs) == 1
+
+
+def test_utcnow(mocker):
+    dt = mocker.patch("datetime.datetime")
+    assert utils.utcnow() == dt.now.return_value
+    dt.now.assert_called_once_with(datetime.timezone.utc)
```
ewjoachim, line 289:
```suggestion
    dt.now.assert_called_once_with(tz=datetime.timezone.utc)
```

tests/unit/test_utils.py
```diff
@@ -280,3 +281,47 @@ async def foo():
...
+def test_aware_datetime(offset, expected):
+    assert utils.aware_datetime(2020, 1, 5, 0, 0, tz_offset=offset) == expected
+
+
+@pytest.mark.parametrize(
+    "input, expected",
+    [
+        ("2020-01-05", datetime.datetime(2020, 1, 5, 0, 0)),
+        (
+            "1984-06-02T19:05:57.720Z",
```
ewjoachim, line 319: Does it work with ` ` instead of `T` too ? And with `+01:00` ?

procrastinate/cli.py
```diff
@@ -283,20 +283,16 @@ def load_json_args(json_args: str, json_loads: Callable) -> types.JSONDict:
...
     if at is None:
         return None
 
     try:
-        datetime = pendulum.parse(at)
-        if not isinstance(datetime, pendulum.DateTime):
-            raise ValueError
-    except (pendulum.exceptions.ParserError, ValueError):
+        dt = utils.parse_datetime(at)
+    except (ValueError):
```
ewjoachim, line 292:
```suggestion
    except ValueError:
```

procrastinate/cli.py
```diff
@@ -283,20 +283,16 @@ def load_json_args(json_args: str, json_loads: Callable) -> types.JSONDict:
     return args
 
 
-# return type should not be Optional[str], but there's a typing bug in
-# Pendulum 2.1.0 (https://github.com/sdispater/pendulum/issues/450)
```
ewjoachim, line 287: Oh, that comment was stale !

procrastinate/utils.py
```diff
@@ -233,3 +236,27 @@ async def wrapper():
         yield task
     finally:
         task.cancel()
+
+
+def utcnow() -> datetime.datetime:
+    return datetime.datetime.now(datetime.timezone.utc)
+
+
+def aware_datetime(*args, tz_offset: Optional[int] = None) -> datetime.datetime:
```
ewjoachim: If I'm not mistaken, this function is only used in tests. Would you mind defining it as a test fixture ? (and maybe remove its own tests)

ewjoachim:
Note for self: this is theoretically a breaking change because we parse fewer formats than before. That being said, I really don't think this is problematic at this stage, but we'll be open about that.

ewjoachim:
Just this point left and we're good to merge ! https://github.com/peopledoc/procrastinate/pull/279#discussion_r456514076

procrastinate/utils.py
EliotBerriot: It's only used in tests indeed, however, in at least one occurence, it's called at the module level in a test file, meaning I cannot make it only a fixture. I've made it a function and moved it into conftest :)

EliotBerriot:
> Note for self: this is theoretically a breaking change because we parse fewer formats than before. That being said, I really don't think this is problematic at this stage, but we'll be open about that.

@ewjoachim I've added more tests, python-dateutil is able to parse other formats than just ISO so it may not be that big a regression after all :)

tests/unit/test_utils.py
EliotBerriot: It should, I've added more tests to showcase that ;)

ewjoachim:
Interesting !

procrastinate/utils.py
```diff
@@ -245,9 +245,14 @@ def utcnow() -> datetime.datetime:
...
+            dt = dt.replace(tzinfo=datetime.timezone.utc)
+        return dt
     except ValueError:
         pass
     # this parser is quite forgiving, and will attempt to return
     # a value in most circumstances, so we use it as last option
-    return dateutil.parser.parse(raw)
+    dt = dateutil.parser.parse(raw)
+    dt = dt.replace(tzinfo=datetime.timezone.utc)
+    return dt
```
ewjoachim, line 258: Can `dateutil.parser.parse`` return something else than whatever `dateutil.parser.isoparse` returns in cases where ``dateutil.parser.isoparse`` does ?

Just trying to understand in what case this would behave differently if we removed the whole try/except block.

ewjoachim: (I'll merge but we can continue discussing it. Best case scenario, this will allow for a follow-up PR simplifying the code)

EliotBerriot: Both functions should return a `datetime.datetime` object. However, the second one seem more unpredictable when given exotic formats, hence the first attempt with the stricter one. Removing the `isoparse` call break the test with the `+01:00` offset.

ewjoachim: Ok, perfect then. As long as we have the proof that removing the first call changes anything, I'm fine keeping both. Thanks for your thourough answer

PR #283, opened 2020-07-18
Django contrib app

### LINKED ISSUE #280: Best way to integrate with django migration system
This isn't a bug or feature request but rather a discussion to understand the best way to integrate `procrastinate` within a Django project, in particular, at the SQL level, as Django as its own migration system.

My first attempt was to use raw SQL and execute each `sql/migrations/delta*` delta file, like this:

```python
# Generated by Django 3.0.8 on 2020-07-18 08:39
import os
from django.db import connection
from django.db import migrations
from django.db import transaction


def apply_sql(apps, schema_editor):
    import procrastinate
    sql_path = os.path.join(
        os.path.dirname(procrastinate.__file__),
        'sql',
        'migrations',
    )
     
    migrations = [
        "baseline-0.5.0.sql",
        "delta_0.5.0_001_drop_started_at_column.sql",
        "delta_0.5.0_002_drop_started_at_column.sql",
        "delta_0.5.0_003_drop_procrastinate_version_table.sql",
        "delta_0.6.0_001_fix_procrastinate_fetch_job.sql",
        "delta_0.7.1_001_fix_trigger_status_events_insert.sql",
        "delta_0.8.1_001_add_queueing_lock_column.sql",
        "delta_0.10.0_001_close_fetch_job_race_condition.sql",
        "delta_0.10.0_002_add_defer_job_function.sql",
        "delta_0.11.0_003_add_procrastinate_periodic_defers.sql",
        "delta_0.12.0_001_add_foreign_key_index.sql",
    ]

    with connection.cursor() as cursor:
        with transaction.atomic():
            for name in migrations:
                full_path = os.path.join(sql_path, name)
                with open(full_path, 'rb') as f:
                    print('Applying {}'.format(full_path))
                    sql = f.read().decode()
                    cursor.execute(sql)


def rewind(*args, **kwargs):
    pass


class Migration(migrations.Migration):

    dependencies = [
        ('common', '0008_auto_20200701_1317'),
    ]

    operations = [
        migrations.RunPython(apply_sql, rewind)
    ]
```

However, this crashes, with the following output:

```
python manage.py migrate common 0009

Operations to perform:
  Target specific migration: 0009_procrastinate, from common
Running migrations:
  Applying common.0009_procrastinate...Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/baseline-0.5.0.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.5.0_001_drop_started_at_column.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.5.0_002_drop_started_at_column.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.5.0_003_drop_procrastinate_version_table.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.6.0_001_fix_procrastinate_fetch_job.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.7.1_001_fix_trigger_status_events_insert.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.8.1_001_add_queueing_lock_column.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.10.0_001_close_fetch_job_race_condition.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.10.0_002_add_defer_job_function.sql
Applying /venv/lib/python3.7/site-packages/procrastinate/sql/migrations/delta_0.11.0_003_add_procrastinate_periodic_defers.sql
  File "/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.SyntaxError: syntax error at or near ";"
LINE 9: DROP FUNCTION IF EXISTS procrastinate_defer_job;
                                                       ^


  File "manage.py", line 27, in <module>
    execute_from_command_line(sys.argv)
  File "/venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/venv/lib/python3.7/site-packages/django/core/management/base.py", line 328, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/venv/lib/python3.7/site-packages/django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "/venv/lib/python3.7/site-packages/django/core/management/base.py", line 83, in wrapped
    res = handle_func(*args, **kwargs)
  File "/venv/lib/python3.7/site-packages/django/core/management/commands/migrate.py", line 233, in handle
    fake_initial=fake_initial,
  File "/venv/lib/python3.7/site-packages/django/db/migrations/executor.py", line 117, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/venv/lib/python3.7/site-packages/django/db/migrations/executor.py", line 147, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  File "/venv/lib/python3.7/site-packages/django/db/migrations/executor.py", line 245, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/venv/lib/python3.7/site-packages/django/db/migrations/migration.py", line 124, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/venv/lib/python3.7/site-packages/django/db/migrations/operations/special.py", line 190, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/app/funkwhale_api/common/migrations/0009_procrastinate.py", line 37, in apply_sql
    cursor.execute(sql)
  File "/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 100, in execute
    return super().execute(sql, params)
  File "/venv/lib/python3.7/site-packages/cacheops/transaction.py", line 93, in execute
    result = self._no_monkey.execute(self, sql, params)
  File "/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 68, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 77, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
  File "/venv/lib/python3.7/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql)
django.db.utils.ProgrammingError: syntax error at or near ";"
LINE 9: DROP FUNCTION IF EXISTS procrastinate_defer_job;
```

Execution of `delta_0.11.0_003_add_procrastinate_periodic_defers.sql` fails with a syntax error, which I don't understand unfortunately. If you have some ideas, I can definitely try these, and possibly document Django integration when I have a working setup :)

EliotBerriot:
Closes #280 

### Successful PR Checklist:

- [x] Tests
- [ ] Documentation

I'll add documentation if the PR is considered okay, but basically:

1. Add `"procrastinate.contrib.django",` to `INSTALLED_APPS`
2. Run `python manage.py migrate`

Let me know if you have any question!


tests/acceptance/test_django.py
```diff
@@ -0,0 +1,26 @@
...
+from django.db import connection
+
+
+def test_django_migrations_run_properly(db):
+    # At this point, with the db fixture, we have all migrations applied
+    with connection.cursor() as cursor:
+        cursor.execute("SELECT * FROM procrastinate_jobs")
+
+
+def test_no_missing_migration():
```
EliotBerriot, line 12: This is a safeguard test to ensure we never add a `.sql` migration without its `.py` counterpart

procrastinate/contrib/django/utils.py
```diff
@@ -0,0 +1,9 @@
+import pathlib
+
+
+def get_sql(name: str) -> str:
+    sql_path = migration_path = (
+        (pathlib.Path(__file__).parent.parent.parent) / "sql" / "migrations"
+    )
+    with open(migration_path / name, "rb") as f:
+        return f.read().decode()
```
ewjoachim: There's already a similar function in schema.py. Do you think:

- it would make sense to put this function there ?
- it would make sense to use `importlib_resources.read_text` ?

(My opinion is yes to both, but at this stage it's more of a discussion than a suggestion)

tests/acceptance/test_django.py
```diff
@@ -0,0 +1,26 @@
+import pathlib
+
+from django.db import connection
+
+
+def test_django_migrations_run_properly(db):
```
ewjoachim, line 6: I know the `db` fixture comes from pytest-django, but I'm afraid this could be really unclear. Would you mind adding a very explicit alias:

```python
# In this file of conftest.py
def django_db(db):
	# db is provided by pytest-django
    return db
```

tests/acceptance/test_django.py
```diff
@@ -0,0 +1,26 @@
...
+    ]
+    sql_migrations_path = procrastinate_dir / "sql" / "migrations"
+    sql_migrations = list(sql_migrations_path.iterdir())
+
+    assert len(sql_migrations) > 0
+    assert len(django_migrations) > 0
+
+    assert len(sql_migrations) == len(
+        django_migrations
+    ), "Some django migrations are missing"
```
ewjoachim, line 26: We have `tests/migrations/`, which is here to test that we didn't forget to write a sql migration. I think this should leave next to that test.

tests/unit/test_django_utils.py
```diff
@@ -0,0 +1,18 @@
...
+    migration_path = (
+        (pathlib.Path(__file__).parent.parent.parent)
+        / "procrastinate"
+        / "sql"
+        / "migrations"
+    )
+    with open(migration_path / migration_name, "rb") as f:
+        expected = f.read().decode()
+
+    assert utils.get_sql(migration_name) == expected
```
ewjoachim, line 18: This feels like we're testing the code by reimplementing the function twice with the same code. This means we could have the same bug in both sides. We know that this file will not change. What about:

```python
def test_get_sql():
    migration_name = "baseline-0.5.0.sql"
    migration = utils.get_sql(migration_name)

	assert migration.startswith(
		"CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;"
	)
	assert len(migration.splitlines()) == 187
```

(I prefer tests to compare dynamic results to static, human-verified things)

setup.cfg
```diff
@@ -105,6 +105,9 @@ no_implicit_optional = True
 [mypy-setuptools.*,psycopg2.*,importlib_metadata.*,importlib_resources.*,aiopg.*,croniter.*]
 ignore_missing_imports = True
 
+[mypy-procrastinate.contrib.django.migrations.*,procrastinate.contrib.django.apps]
```
ewjoachim, line 108: I don't understand why not just `django.*` ?

setup.cfg
```diff
@@ -37,10 +37,10 @@ install_requires =
 dev =
     tox
     black
-    django>=3,<=4
     isort>=5.0.0
 
 test =
+    django>=3,<=4
```
ewjoachim, line 43: What about a new extra ? This way, `pip install procrastinate[django]` would ensure your django version is consistent with other deps ?

Also: `django>=3,<=4` can be written `django~=3.0`

Finally... I'd rather not explicitely remove Django 4 if we have no reason to believe it will fail.

Finally: the current LTS is 2.2, we should be supporting it, I think.

So in the end, I'd say: `django>=2.2`

procrastinate/contrib/django/utils.py
```diff
@@ -1,7 +1,73 @@
...
+
+
+def sort_nicely(l):
+    """
+    Sort the given list in the way that humans expect.
+    """
+    l.sort(key=alphanum_key)
+
+
+SQL_PATH = (pathlib.Path(__file__).parent.parent.parent) / "sql" / "migrations"
```
ewjoachim, line 28:
```suggestion
SQL_PATH = (pathlib.Path(__file__).parents[2]) / "sql" / "migrations"
```

setup.py
```diff
@@ -41,5 +43,27 @@ def ensure_version():
...
+def generate_django_migrations():
+    from procrastinate.contrib.django.utils import generate_migrations
+
+    generate_migrations()
+
+
+class InstallWrapper(install):
+    def run(self):
+        generate_django_migrations()
+        # Run the standard PyPi copy
```
ewjoachim, line 55:
```suggestion
        # Run the standard PyPI copy
```

setup.py
```diff
@@ -41,5 +43,27 @@ def ensure_version():
         os.remove(version_filename)
 
 
+def generate_django_migrations():
+    from procrastinate.contrib.django.utils import generate_migrations
```
ewjoachim, line 47: Hm, this will import `procrastinate.__init__`, `procrastinate.contrib.__init__` and `procrastinate.contrib.django.__init__`. From an empty venv, this will very likely fail with `ImportError`. I think the generation code needs to live in the `setup.py` directly (or we need a top-level script that will be distributed alongside Procrastinate)

setup.py
```diff
@@ -41,5 +43,27 @@ def ensure_version():
...
+        generate_django_migrations()
+        # Run the standard PyPi copy
+        develop.run(self)
+
+
 with ensure_version() as version:
-    setuptools.setup(version=version)
+    setuptools.setup(
+        version=version, cmdclass={"install": InstallWrapper, "develop": DevelopWrapper}
+    )
```
ewjoachim, line 69: I'm wondering: why not just call generate_django_migrations() at the top level of setup.py ? What are the cases where setup.py runs and we don't want to generate migrations ?

ewjoachim:
Splendid! I think your proposal goes in the right direction.

I'll let you be the judge for the following: I don't want to delay the whole thing just because of my silly suggestions, so if you think it makes sense, we could split that into 2 PRs: one without automated generation (or with automated generation but not triggered by setup.py) and one with fully automated generation in setup.py, which seems to be not as simple. You decide, really, I'm fine with both.

ewjoachim:
Hey :) Just want to make sure everything is all right regarding this contribution. You're welcome to take all the time you need to iterate on the different parts, but if you run into a blocker or have a part you don't want to invest your time into, it's very ok not to. Just make sure to let me know if/how I can help.

See you around!

EliotBerriot:
Hey @ewjoachim ! 

I've lacked the spoons lately, but I do want to get back at it and complete it, probably without automation :) 

I hope you're okay, take care

ewjoachim:
Hey :)

Here's a handful of spoons :spoon::heavy_multiplication_x::100:
(for anyone wondering what those two strange individuals are talking about, have a look at [this](https://en.wikipedia.org/wiki/Spoon_theory))

Take your time, it's ok. Wishing you all the best and 10% more on top of that.

See you around :)

EliotBerriot:
So I reverted the dynamic migration generation and integrated your suggestions:

- [django] extra
- Moving functions / tests in different modules for consistency

However, the `get_sql` call to `read_text` fails with:

```
FileNotFoundError: 'baseline-0.5.0.sql' resource not found in 'procrastinate.sql.migrations'
```

I'm not familiar at all with `importlib_resources` so maybe you'll have an idea @ewjoachim ?

tests/unit/test_schema.py
```diff
@@ -15,3 +17,13 @@ def test_apply_schema(app, connector):
...
+
+
+def test_get_sql(app):
+    migration_name = "baseline-0.5.0.sql"
+    migration = app.schema_manager.get_sql(migration_name)
+
+    assert migration.startswith(
+        "CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;"
+    )
+    assert len(migration.splitlines()) == 187
```
ewjoachim: Yay, that seems perfect !

ewjoachim:
Hey there :) Sorry for the response time, I'm going to have a look. Thanks for taking the time!

ewjoachim:
First approximation: I think it's because read_text only works with python modules, so the migrations folder needs an empty `__init__.py` file. When I add it it seems I have a test error further down the line, so this could be it!

procrastinate/contrib/django/migrations/0001_initial.py
```diff
@@ -0,0 +1,12 @@
+from django.db import migrations
+
+from ..utils import get_sql
```
ewjoachim, line 3:
```suggestion
from procrastinate.schema import get_sql
```

setup.cfg
```diff
@@ -64,6 +65,9 @@ docs =
 docs_spelling =
     sphinxcontrib-spelling
 
+django = 
+    django>=2.2
```
ewjoachim, line 69: Can you add the django extra in `requirements.txt` too please ?

setup.cfg
```diff
@@ -96,13 +100,16 @@ testpaths =
     tests/integration
     tests/acceptance
     tests/migration
-
+DJANGO_SETTINGS_MODULE = tests.acceptance.django_settings
```
ewjoachim, line 103: Can we leave a blank line before mypy ? Thanks :)

tests/acceptance/django_settings.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+SECRET_KEY = "test"
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.postgresql_psycopg2",
+        "NAME": os.environ.get("PGDATABASE_DJANGO", "procrastinate"),
+        "USER": os.environ.get("PGUSER", "postgres"),
+        "PASSWORD": os.environ.get("PGPASSWORD", "password"),
+        "HOST": os.environ.get("PGHOST", "localhost"),
```
ewjoachim, line 10: As far as I can tell, only the NAME is mandatory. Other variables will already default to the libpq env vars if not present. We could keep ENGINE and NAME.

Also, is there a specific reason for using `PGDATABASE_DJANGO` instead of `PGDATABASE` and could we make it default to `PGDATABASE` if `PGDATABASE_DJANGO` is not provided ?

tests/migration/test_migration.py
```diff
@@ -76,3 +83,26 @@ def test_migration(schema_database, migrations_database, pum):
...
+
+
+def test_no_missing_django_migration():
+    procrastinate_dir = pathlib.Path(__file__).parent.parent.parent / "procrastinate"
+    django_migrations_path = procrastinate_dir / "contrib" / "django" / "migrations"
+    django_migrations = [
+        e for e in list(django_migrations_path.iterdir()) if e.name.startswith("0")
+    ]
+    sql_migrations_path = procrastinate_dir / "sql" / "migrations"
+    sql_migrations = list(sql_migrations_path.iterdir())
```
ewjoachim, line 101:
```suggestion
    sql_migrations = [e for e in sql_migrations_path.iterdir() if e.suffix == ".sql"]
```

tests/migration/test_migration.py
```diff
@@ -76,3 +83,26 @@ def test_migration(schema_database, migrations_database, pum):
...
+    # At this point, with the db fixture, we have all migrations applied
+    with connection.cursor() as cursor:
+        cursor.execute("SELECT * FROM procrastinate_jobs")
+
+
+def test_no_missing_django_migration():
+    procrastinate_dir = pathlib.Path(__file__).parent.parent.parent / "procrastinate"
+    django_migrations_path = procrastinate_dir / "contrib" / "django" / "migrations"
+    django_migrations = [
+        e for e in list(django_migrations_path.iterdir()) if e.name.startswith("0")
```
ewjoachim, line 98:
```suggestion
        e for e in django_migrations_path.iterdir() if e.name.startswith("0")
```

procrastinate/contrib/django/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class ProcrastinateConfig(AppConfig):
+    name = "procrastinate.contrib.django"
+    label = "procrastinate_django"
```
ewjoachim, line 6: I think "procrastinate" could be enough in this case. Also, this would make the verbose name (which defaults on `label.title()`) cleaner.

Note: I'm not 100% sure.

EliotBerriot:
I've integrated your suggestions @ewjoachim, and the test suite seems to pass \o/

ewjoachim (approved):
Aaaaand it's a wrap :clapper: 

Congratulations :)

If you want to continue contributing, let me know, I'm sure we can find interesting avenues :) Feel free to have a look at the issues anyway !

PR #288, opened 2020-08-07
Make test_periodic_deferrer more robust

### LINKED ISSUE #276: acceptance "test_periodic_deferrer" is flaky
See the error in https://travis-ci.org/github/peopledoc/procrastinate/jobs/706522770
This does not reproduce easily.

I'm not sure whether this is a real subtle bug in periodic tasks or a test problem. We should analyze this and take appropriate measures.

elemoine:
Closes #276 

See https://github.com/peopledoc/procrastinate/issues/276#issuecomment-670538989.

There may be a way to fix this by changing the code instead of the test, but the `croniter` API doesn't make that easy. So fixing the test for now.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/tests/acceptance/test_nominal.py
+++ b/tests/acceptance/test_nominal.py
@@ -154,10 +154,16 @@ def test_queueing_lock(defer, running_worker):
 
 def test_periodic_deferrer(worker):
     # We're launching a worker that executes a periodic task every second, and
-    # letting it run for 2.5 s. It should execute the task 3 times, and print to stdout:
-    # 0 <timestamp>
-    # 1 <timestamp + 1>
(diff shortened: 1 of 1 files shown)


elemoine:
> I'm willing to accept this for now but trying to understand if this means we should fix something or document something in the larger scheme of things :D

I went through the same thinking process :-)

ewjoachim:
Added another PR. I think we can merge this one.

elemoine:
Closing this, as we're leaning toward a better solution.

PR #290, opened 2020-08-14
Add github actions

### LINKED ISSUE #287: Switch to GitHub Actions
### Rationale

Reasons to leave Travis:
- The whole travis.org / travis.com is, as far as I can tell, quite a mess
- The back and forth when I was trying to get the config working quite turned me away

Reasons to choose GitHub Actions:
- Supports Windows (as far as I can tell) and it appears we may have our first Windows users so maybe it's better if the CI can test Windows.
- We already have release drafter
- As far as I can tell, if we make a working pipeline, we might be able to factor it in other repos.

Things that are not a reason but might count still:
- Seems fun and I'd like to learn more about GH Actions.

Things to look for:
- The new CI should be simple to maintain and understand.
- In particular, the deployment is really simple, so I wish it can stay as simple.

Alternatives:
- CircleCI ? 

### What needs to be done:

On PR push and/or branch push, run either on the branch HEAD or the merge commit:
- tox: all python versions
- check-lint
- docs
- Maybe docs-spelling BUT only if it's not hard to run, if it doesn't change the current results and if it doesn't keep from merging (I don't want this to be a barrier to contribution)

On PR merge: release drafter (already present)

On tag push: release and twine upload

thomasperrot:
Closes #287

**N.B**: lint, tests and docs stage are run in parallel to fasten the CI. The upload to PyPi is performed only when a tag is pushed, and when the abovemention stages have succeeded.

Already done: 
- [X] lint checking
- [X] run unittests in all python versions
- [X] build coverage and upload it
- [X] build doc
- [x] upload doc
- [x] build package
- [ ] upload on pypi (implemented, but not tested yet)

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [X] Documentation


ewjoachim:
Hello! Thanks for the contribution! I'll try to have it reviewed soon :)

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
```
ewjoachim, line 9: Does this mean only PRs targetting master run the CI ?
1. [Doc is not clear](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) on what branch means for PRs
2. And why would we want to limit CI to PRs on master ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+jobs:
+  lint:
+    name: Check lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
ewjoachim, line 20: If we just specify `3`, what version is used ? I'd be inclined to do that if the version used is recent enough, so we don't need to remember to change this.

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
```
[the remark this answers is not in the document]
ewjoachim: Is there a way to bust the cache if there's a problem ? If not, can we add a "CACHE_VERSION" env var in the cache key that we may manually update if we need to force a refresh ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        run: |
+          set -xe
+          python -VV
+          python -m pip install --upgrade tox
```
ewjoachim, line 33: so this would upgrade tox but not potential dependencies if they're already installed (we'd need the `eager` update strategy).

Why the upgrade ? If tox is already installed, by whom, at which version ? Would there be a way to avoid installing tox altogether ? Or if we install it, just pip install because we know we're in a controlled clean env ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+        run: |
+          set -xe
+          python -VV
+          python -m pip install --upgrade coverage
+      - name: Download coverages artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: coverage-reports
+      - name: Combine coverage artifacts
+        run: python -m coverage combine .coverage-*
```
ewjoachim, line 107: If we combine the coverage runs, codecov looses a bit of information. I think itis already able to combine reports, but it's also able to say what coverage came from which report, which it won't in this configuration. Do we need to do this ? Wouldn't it be easier to upload coverage just after running the tests ?

To be honest though, I could accept if we lose codecov as long as we get a report on the coverage, but we've been bitten by this (nice and free !) service a few times.

CONTRIBUTING.rst
```diff
@@ -423,9 +423,9 @@ relevant edits to the changelog, (see ``TODO``) including listing the migrations
 for the release. Click on Release, that's it, the rest is automated.
 
 When creating the release, GitHub will save the release info and create a tag with
-the provided version. The new tag will be seen by Travis, which will then create a
-wheel (using the tag as version number, thanks to our ``setup.py``), and push it
-to PyPI (using the new API tokens). That tag should also trigger a ReadTheDocs
+the provided version. The new tag will be seen by Github Actions, which will then
+create a wheel (using the tag as version number, thanks to our ``setup.py``), and push
+it to PyPI (using the new API tokens). That tag should also trigger a ReadTheDocs
```
ewjoachim, line 428: I don't see the part where we do that ...?

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+        ports:
+          - 5432:5432
+
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
ewjoachim, line 58: Can we add a comment to make sure this stays synced with tox ? (and the other way around)

.github/workflows/main.yml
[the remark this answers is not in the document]
thomasperrot: > well, requirements.txt does not change a lot, dependencies are in setup.cfg

You are right, I forgot that. If `setup.cfg` does not change too much, maybe we can use its md5 in the cache key

> Is there a way to bust the cache if there's a problem ?

According to the documentation, their is no functionnality that allow us to do that. See https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

> can we add a "CACHE_VERSION" env var in the cache key that we may manually update if we need to force a refresh

We can not do that using env variable, but we can do this using file contents. For instance we can have a file named `CACHE_VERSION` that contains a string that is changed everytime the cache needs to be updated. The key would be : `${{ runner.os }}-pip-${{ hashFiles('requirements.txt')-${{ hashFiles('CACHE_VERSION') }}`. I am not a huge fan of this method. Maybe we could remove the cache implementation from the CI ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+        with:
+          python-version: 3.8
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
```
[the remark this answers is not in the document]
thomasperrot: It allows us to use other cache versions if no cache version match the current one. For instance if the requirements have changed, we can still load the previous cache, which is faster than downloading again the hole dependancy tree. See https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

.github/workflows/main.yml
```diff
@@ -0,0 +1,131 @@
...
+      - name: Install dependencies
+        run: |
+          set -xe
```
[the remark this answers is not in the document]
thomasperrot: Duly noted, I will change that.

.github/workflows/main.yml
thomasperrot: > Does this mean only PRs targetting master run the CI ?

Absolutely

> And why would we want to limit CI to PRs on master ?

¯\_(ツ)_/¯ I assumed it without real reason. We can remove this.

.github/workflows/main.yml
thomasperrot: > If we just specify 3, what version is used ?

The correct syntax is `3.x`, which according to the documentation uses the « latest minor release of Python 3 ». So currently it is the `3.8.5`. It will use `3.9` when `3.9` becomes stable if I understand correctly.

> I'd be inclined to do that if the version used is recent enough, so we don't need to remember to change this.

I agree, I will change that.

.github/workflows/main.yml
thomasperrot: Yes ! Additionaly, I will try to find a way to avoid using `tox-gh-actions`, and to set the matching between tox version with pythons versions in the `main.yml` file. I do not like to put some CI logic in the `tox.ini` file.

CONTRIBUTING.rst
thomasperrot: It is not implemented yet. This PR is currently in draft (but I wanted the CI to run, so I set is as open). I added a check list of what remains to be done in the PR description.

.github/workflows/main.yml
thomasperrot: Because we check the python version in some parts of the code (`signals.py` if I remember correctly), running the coverage on the unitests of a single python version (python 3.8 for instance) is not sufficient to achieve 100% coverage. Additionaly, in the current CI implementation, the unittests are run in parallel on different workers, so we have to upload the `.coverage` files to a single worker once all the tests have completed. I could not find a better way to upload the coverage than combining the `.coverage` with `coverage combine`. If you know a better way, I will gladly replace what I have done ! :+1:

.github/workflows/main.yml
thomasperrot: > so this would upgrade tox but not potential dependencies if they're already installed

yes

> Why the upgrade ?

No need actually

> If tox is already installed, by whom, at which version ?

The environment is supposed to be clean. `tox` should not be installed yet.

> Would there be a way to avoid installing tox altogether ?

I do not think so. How would you check linting if tox is not installed ?

> just pip install because we know we're in a controlled clean env ?

Yep, this is what I think is the best thing to do :+1:

.github/workflows/main.yml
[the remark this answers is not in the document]
ewjoachim: Ok, but the line with `pip-` seems not to do anything. I think the doc hints that it should be the branch name there, so you'd get "same commit or same branch or anything"

.github/workflows/main.yml
ewjoachim: I'd rather the CI runs on every PR and on protected branches (which is just master)

.github/workflows/main.yml
ewjoachim: Hm, I think my question lacked clarity. Let's chat about this in our native language :p

CONTRIBUTING.rst
ewjoachim: CI doesn't run on Draft PRs ? That's not ideal... Is it something we can control ?

.github/workflows/main.yml
ewjoachim: isn't the TOX_ENV env var trick that was used in Travis applicable here too ?

.github/workflows/main.yml
ewjoachim: >> Would there be a way to avoid installing tox altogether ?
> 
> I do not think so. How would you check linting if tox is not installed ?

Nah :) I mean "if somehow tow was already present in our environment, maybe we could avoid touching it and let whoever made it available manage it" but you answered: tow is not installed and the env is clean. So let's `pip install` all the things !

CONTRIBUTING.rst
ewjoachim: Seems it does, yay :tada:

.github/workflows/main.yml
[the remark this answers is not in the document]
ewjoachim: Hm, given what it is we're caching (the pip wheel cache), I think it's safe to accept that we can't bust it. Worst case scenario, we'll add a manual `-2` in the key if we ever run into a problem.

I'm surprised (and a bit disappointed) that we can't use env var though.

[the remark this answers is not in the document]
ewjoachim: Hm, why doesn't it work ? https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables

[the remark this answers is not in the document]
thomasperrot: We can set environment variables, but they are hardcoded in the CI. There is no way to change their value manually when running the CI (such as in Gitlab CI for exemple). We can however change their value in the main.yml file, which will be versionned. Actually it is not worst than using a `CACHE_VERSION` file, which would also have been versionned.

.github/workflows/main.yml
ewjoachim: conclusion: each stage will upload its own coverage data to codecov, and we'll have less code here :)

CorBott (approved):
Nothing to report!

.github/workflows/main.yml
```diff
@@ -0,0 +1,128 @@
...
+          python -VV
+          pip install tox coverage
+      - name: Run tox targets for ${{ matrix.python-version }}
+        run: tox -e ${{ matrix.tox-env }}
+        env:
+          PGDATABASE: procrastinate
+          PGHOST: localhost
+          PGUSER: postgres
+          PGPASSWORD: password
+          PYTEST_ADDOPTS: --cov-report xml
```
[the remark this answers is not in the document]
thomasperrot: The default is SQLite format, which is not compatible with codecov. In our travis CI, the `codecov` command automatically convert the `.coverage` file (in SQLite format) to XML before uploading it. We have to do it manually here.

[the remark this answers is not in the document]
ewjoachim: I thought we already had this option in https://github.com/peopledoc/procrastinate/blob/b1adb93795aae1d881b6ce60b6106ce80ff724f0/setup.cfg#L98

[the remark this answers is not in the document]
ewjoachim: But... How come this currently works in travis Oo ?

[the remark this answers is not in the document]
thomasperrot: I prefered add this in the CI rather than in the setup.cfg, as the xml file is only needed in the CI :)

ewjoachim (approved):

PR #296, opened 2020-08-20
Mixed things around period tasks, locks and such

### LINKED ISSUE #289: Deferring a periodic job to multiple queues does not work as expected
Let's say I want per worker server periodic tasks. Naturally I'd be tempted to use per-server queues:

```py
@app.periodic(cron="*/5 * * * *")
@app.task(queue=host_settings.queue_name)
def my_task(timestamp):
    …
```

But, as I understand it, the `procrastinate_defer_periodic_job` function will [refuse](https://github.com/peopledoc/procrastinate/blob/dce801709585f4f8f54676359672965e0f3a9880/procrastinate/sql/schema.sql#L100-L108) to defer a periodic job if a periodic job with the same name and same defer timestamp was already deferred, even if the periodic job is not deferred to the same queue as the one already deferred.

Shouldn't the `procrastinate_periodic_defers` table include a `queue_name` column? And shouldn't the unique constraint be `UNIQUE (task_name, queue_name, defer_timestamp)` rather than `UNIQUE (task_name, defer_timestamp)`?

### LINKED ISSUE #292: PeriodicDeferrer should continue to work even when all workers are busy
Imagine a setup with:
- A single worker
- Only sync tasks
- Most tasks are short, but 10% of the tasks take 30 minutes (so it's likely that ~100% of the worker time is spent on those tasks)
- There's a periodic task that should run every minute.

Given the only worker is almost always busy with the long task, the execution of the periodic job will have to wait and that's normal. What not normal is that the deferring of those tasks also has to wait, and moreover, when it happens, only the last missed task will be deferred. We expect that all 30 jobs would be deferred in between long tasks, each with their own unique timestamp.

So I'm not saying the PeriodicDeferrer should run in another thread, but when it hasn't run because of long tasks, it should consider ALL missed periods, not just the latest one.

There's an argument to be made on whether the MAX_DELAY (10 minutes) should be respected here or not. I think not, because it's not about running tasks for the current period at startup, here, it's deferring tasks that absolutely should have been deferred if internally, procrastinate wasn't running on an async event loop while still allowing for long tasks that will block the loop.

Alternatively, sync tasks could be ran asynchronously in a separate thread, but this raises a LOT of thread safety potential issues, but it could be a whole different way of solving the issue, a way that might make things much easier.

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor

### LINKED ISSUE #295: Default queue is ignored in `app.configure_task`
`app.configure_task()` (as opposed to `task.configure()`) doesn't try to load the task. If the task doesn't exist, then fine. If it does though, the task might be specifying a default queue that we would be ignoring, and that's not cool.

ewjoachim:
Ok, I've just thrown away code together for discussion. No tests, no docs, it's just for wrapping our heads around the several linked problems.

Fixes #295: app.configure_task uses tasks defaults when they exist
Related to #292: tasks accept default lock and queuing lock
Related to #292: Periodic jobs use task default lock and queueing lock
Closes #289: same periodic tasks on different queues can coexist
Fixes #292: periodic deferrer doesn't miss tasks anymore 

Of course, we can extract each of these subjects in its own PR.

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/store.py
+++ b/procrastinate/store.py
@@ -1,9 +1,12 @@
 import asyncio
 import datetime
+import logging
 import uuid
 from typing import Any, Dict, Iterable, Optional
 
 from procrastinate import connector, exceptions, jobs, sql
+
+logger = logging.getLogger(__name__)
 
 
 def get_channel_for_queues(queues: Optional[Iterable[str]] = None) -> Iterable[str]:
@@ -26,7 +29,7 @@ class JobStore:
                 **self._defer_job_query_kwargs(job=job)
             )
         except exceptions.UniqueViolation as exc:
-            self._raise_already_enqueued(exc=exc, job=job)
+            self._raise_already_enqueued(exc=exc, queueing_lock=job.queueing_lock)
 
         return result["id"]
 
@@ -36,7 +39,7 @@ class JobStore:
                 **self._defer_job_query_kwargs(job=job)
             )
         except exceptions.UniqueViolation as exc:
-            self._raise_already_enqueued(exc=exc, job=job)
+            self._raise_already_enqueued(exc=exc, queueing_lock=job.queueing_lock)
 
         return result["id"]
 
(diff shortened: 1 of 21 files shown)


procrastinate/store.py
```diff
@@ -64,6 +64,8 @@ async def defer_periodic_job(self, task, defer_timestamp) -> Optional[int]:
         result = await self.connector.execute_query_one_async(
             query=sql.queries["defer_periodic_job"],
             queue=task.queue,
+            lock=task.lock,
+            queueing_lock=task.queueing_lock,
             task_name=task.name,
             defer_timestamp=defer_timestamp,
         )
```
ewjoachim, line 71: the queueing lock may raise here. Debug log & ignore.

ewjoachim: Maybe reuse defer_job_async

elemoine (approved):
Reviewed it again, and it looks great :+1:

PR #297, opened 2020-08-22
Django auto-generate migrations

ewjoachim:
It's still needed to generate the Django migrations when writing a SQL migration but now, it can be done automatically with `bin/generate-django-migrations`.

Note: it would be hard to write tests for this script, so I haven't. We'll see if we run into a problem. The script isn't included in the package.

Also, I've thought about generating migrations at release time, but it would require a lot of code in setup.py ( :( ) and it would be a mess to avoid importing resources not installed yet. If we only did wheels, this would be easier, but we're doing sdists too and it's a mess. Similarily, I explored the idea of generating the migrations just in time in `manage.py migrate` but that would require the executing user to have write permissions on the venv and we're absolutely not sure this will be true for our users (let alone ourselves) so I didn't explore further.

A more advanced solution could be to generate migrations using
```console
$ DJANGO_SETTINGS_MODULE=tests.acceptance.django_settings django-admin makemigrations procrastinate
```
This would require redefining `makemigration` to use a custom `MigrationAutodetector` subclass which would add our own operations to the mix, the benefit being that we would use Django's own machinery to generate the migration files and the dependencies. I'm not sure of other benefits, but it sounds fun :)

Also, I'm adding some documentation regarding Django.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)


docs/howto/django.rst
```diff
@@ -0,0 +1,16 @@
+Use Procrastinate in a Django application
+=========================================
+
+Add Procrastinate in your ``INSTALLED_APPS``::
```
ewjoachim, line 4: We need to mention to install with `pip install 'procrastinate[django]'`

ewjoachim:
I think #298 is a better approach.

PR #300, opened 2020-08-25
Trigger CI when publishing a new release

thomasperrot:
### Successful PR Checklist:

- [X] Tests (on my own repository)
- [ ] Documentation
  - [X] (not applicable?)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,8 +5,10 @@ on:
     branches:
       - master
     tags:
-      - '*'
   pull_request:
+  release:
+    types:
+      - published
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -5,8 +5,10 @@ on:
     branches:
       - master
     tags:
-      - '*'
```
elemoine, line 8: Do you want/need to keep the `tags` key?

thomasperrot: Not sure about that. If we always use the releases to deploy new version, tags are not needed. Besides, the last step would contain 
```
if: github.event_name == 'release'
```
instead of 
```
if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
```

ewjoachim: no, we might release with a manual tag if something broke

elemoine: I'd say yes, we always use releases, as our workflow is to create a release from a draft release on github.

ewjoachim: That being said, the "tags" argument didn't work :( so, not sure about that. I don't know.

ewjoachim (approved):

PR #305, opened 2020-08-28
Fix black version

### LINKED ISSUE #304: Do not use black>=20.8b0 for the moment
The last versions of black, `20.8b0` and `20.8b1` contain a bug that breaks our linting. See https://github.com/psf/black/issues/1649. We need to use the most recent version BELOW 20.8b0 for the moment.

thomasperrot:
Closes #304 

### Successful PR Checklist:

- [ ] Tests
  - [X] (not applicable?)
- [ ] Documentation
  - [X] (not applicable?)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
 [options.extras_require]
 dev =
     tox
-    black
+    black==19.10b0
     isort>=5.0.0
 
 test =
@@ -49,7 +49,7 @@ test =
     pum
 
 lint =
-    black
+    black==19.10b0
     flake8
     isort
     mypy


ewjoachim (commented):
I’m ok temporarily excluding a release, I’d rather avoid pinning because this could stay in place longer than necessary.

If this is not a bug but Black’s formatting rules, what would be needed to fix our codebase ?

ewjoachim (changes requested):
(see above)

elemoine:
My initial plan was to use `black!=20.8b1`.

There's more information in this Black issue: https://github.com/psf/black/issues/1649. The comment https://github.com/psf/black/issues/1649#issuecomment-683327600 suggests to remove the trailing comas, which might indeed be a good option for us.

ewjoachim:
Are you ok to close in favor of #306 ?

ewjoachim:
Hm, I've been a little fast and I've merged already but feel free to continue the discussion if anything feels unresolved :)

PR #298, opened 2020-08-22
Auto-generating Django migrations using Django's own machinery

ewjoachim:
Up for discussion! I think I prefer the idea of having Django more autonomous, but:
- Django's code is not very modular, so I can have minimal code
- This means the corresponding code is not very readable
- And I have to reimplement a few details that are already implemented in Django's own code but I couldn't reuse.
- That being said, the code here is more testable than the script it replaces (well... differently testable. The script was not easily testable because it was not part of the python package. That version is more easily testable in this way, but we'll need more integration tests, vs unit tests)

It's kind of funny because Django here does a lot of very smart things, but they have so many unrelated code in the same functions, that there's always one little detail that makes many things not reusable. That being said, this code in django was most probably not supposed to be hacked, so I expected it to be hard.

Hmm. I'm wondering about grouping all the missing SQL migrations in a single django migrationwith multiple operations. This would simplify the code a lot !

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)
- [ ] Had a good time contributing? < I've known better
- [ ] (Maintainers: add PR labels)

PR #308, opened 2020-09-04
Use migra instead of pum for the db migration tests

elemoine:
With this PR we switch from `pum` to `migra` for the db migration tests.

`migra` provides for a better API, making our db migration tests look better. And `migra` provides the SQL DDL code to use to go from the source database to the target database, which can help developers write migrations for their changes.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)


elemoine:
There are pros and cons to switching from pum to migra for our database migration tests.

**pros**

When migra detects differences between the two databases it provides the SQL DDL code to use to go from the source database to the target database, which can help developers write migrations for their changes.

Here's an example of output when the `test_migration.py::test_migration` test fails:

```
>       assert not m.statements                                                                                                                                                                
E       assert not ['alter table "public"."procrastinate_jobs" add column "queueing_lock" text;', "CREATE UNIQUE INDEX procrastinate_jobs..._lock_idx ON public.procrastinate_jobs USING btree 
(queueing_lock) WHERE (status = 'todo'::procrastinate_job_status);"]                                                                                                                           
E        +  where ['alter table "public"."procrastinate_jobs" add column "queueing_lock" text;', "CREATE UNIQUE INDEX procrastinate_jobs..._lock_idx ON public.procrastinate_jobs USING btree (
queueing_lock) WHERE (status = 'todo'::procrastinate_job_status);"] = <migra.migra.Migration object at 0x7fd6bd3f32b0>.statements
```

Compare it pum:

```
Check...DIFFERENCES FOUND                                                                                                                                                                      
columns:                                                                                                                                                                                       
- '- (''public'', ''procrastinate_jobs'', ''queueing_lock'', None, ''YES'', ''text'',                                                                                                          
  None, None, None, None)'                                                                                                                                                                     
constraints: []                                                                                                                                                                                
functions: []                                                                                                                                                                                  
indexes:                                                                                                                                                                                       
- '- (''procrastinate_jobs'', ''procrastinate_jobs_queueing_lock_idx'', ''queueing_lock'')'                                                                                                    
rules: []                                                                                                                                                                                      
sequences: []                                                                                                                                                                                  
tables: []                                                                                                                                                                                     
triggers: []                                                                                                                                                                                   
views: [] 
````

The output of pum is not nice, and it's not of the most useful either.

**cons**

The part that I like the least with migra is that it pulls the dependencies sqlbag, schemainspect and sqlalchemy. I actually like sqlbag and sqlalchemy, but we don't use them in Procrastinate and it feels a bit strange to use them in our tests.

**Comment by ewjoachim** [comment]
I'm ok having tests new dependencies as long as it's only test.

**Review by ewjoachim — COMMENTED** [review]
Preliminary :+1:, waiting for the docs

**Comment by k4nar** [comment]
Maybe we could include `migra` as an optional dep, and only install it in a dedicated env in our tests? That way we make sure that the rest of the code works without sqlalchemy & sqlbag.

**Comment by k4nar** [comment]
That's not what I meant, sorry :) . I meant as a non mandatory requirement for the dev / tests. The related tests would be skipped if the dependency is not present, and they would be ran in a dedicated env in CI.

**Comment by elemoine** [comment]
> That's not what I meant, sorry :) . I meant as a non mandatory requirement for the dev / tests. The related tests would be skipped if the dependency is not present, and they would be ran in a dedicated env in CI.

I should have figured out that you knew "migra" was a dev extra require :wink: 

Going back to your idea, yes, that's an option. But not sure it's worth the effort and extra complexity. And developers would not understand if the tests passed on their local dev environments, but not on the CI.

**Comment by ewjoachim** [comment]
I think our tox setup makes it not so complicated.

Are sqlalchemy or sqlbag known to have side effects ?

**Comment by elemoine** [comment]
I removed the pum howto, and replaced it with a migrations howto.

This is ready for review.

File: tests/migration/test_migration.py [hunk 1]
```diff
@@ -68,22 +66,31 @@ def django_db(db):
...
+        )
+        migrations_db_session = stack.enter_context(
+            S(f"postgresql:///{migrations_database}", poolclass=NullPool)
+        )
+        m = Migration(schema_db_session, migrations_db_session)
+        m.set_safety(False)
+        m.add_all_changes()
+
+    if m.statements:
+        print(m.sql)
```
ewjoachim, line 91:
```suggestion
    print(m.sql)
```

elemoine: Agreed.

docs/howto/migrations.rst
```diff
@@ -0,0 +1,43 @@
...
+whose versions are greater than or equal to 0.11.0, and lower than 0.15.0 (0.11.0
+≤ version < 0.15.0). And you will apply them in version order, and, for a version, in
+serial number order. For example, you will apply the following migration scripts, in
+that order:
+
+1. ``delta_0.11.0_001_add_column.sql``
+2. ``delta_0.11.0_002_add_index.sql``
+3. ``delta_0.12.0_001_drop_index.sql``
+4. ``delta_0.14.0_001_add_table.sql``
+5. ``delta_0.14.0_002_drop_table.sql``
```
ewjoachim, line 43: We should warn that this is not always lexicographic order because 0.10.0 is lexicographically before 0.9.0.

ewjoachim: There should also be a word on Django migrations.

ewjoachim (approved):
a few details in the doc. Otherwise, :+1:

docs/howto/migrations.rst
elemoine: > There should also be a word on Django migrations.

Actually, I'd like to add it when #298 is merged.

elemoine: > We should warn that this is not always lexicographic order because 0.10.0 is lexicographically before 0.9.0.

I added a note for this.

docs/howto/migrations.rst
```diff
@@ -0,0 +1,51 @@
...
+3. ``delta_0.11.0_001_xxxxx.sql``
+4. ``delta_0.11.0_002_xxxxx.sql``
+5. ``delta_0.12.0_001_xxxxx.sql``
+6. ``delta_0.14.0_001_xxxxx.sql``
+7. ``delta_0.14.0_002_xxxxx.sql``
+
+.. note::
+
+    The ordering of versions is not based on the typical lexicographical order that
+    Python uses for comparing strings. For example, version 0.9.0 is lower than 0.10.0.
```
ewjoachim, line 50: It's not just python though. `psql *` would fail too for the same reasons.

elemoine: I changed the wording to (hopefully) make things clearer.

ewjoachim (approved):

ewjoachim (approved):

docs/howto/migrations.rst
elemoine: With #298 now merged I added a note to mention the Django migration scripts, and a link to our Django howto.

PR #315, opened 2020-09-17
Use postgres:12 image for faster tests

elemoine:
This PR suggests to use the postgres:12 image rather than postgres:10 for much much faster tests!

|                            | postgres:10 | postgres:11     | postgres:12 | postgres:12 w/ fsync=off
|-----------------------|----------------|-------------------|-----------------|---------------------------------
| unit tests             | 2.5s            | 2.5s                | 2.5s              | 2.5s
| integration tests   | 35s             | 30s                 | 9s                 | 6s
| acceptance tests | 22s             | 21s                 | 18s               | 17s
| migration tests    | 4s               | 4s                   | 3s                 | 3s
| all the tests         | 1m             | 54s                  | 28s                | 25s

We save 3 additional seconds by setting `fsync` to `off`. 

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,6 @@ jobs:
           --health-timeout 5s
         ports:
           - 5432:5432
-        command: ["-c", "fsync=off"]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,6 +64,7 @@ jobs:
           --health-timeout 5s
         ports:
           - 5432:5432
+        command: ["-c", "fsync=off"]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:10
+        image: postgres:12
         env:
           POSTGRES_DB: procrastinate
           POSTGRES_PASSWORD: password
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,3 +21,4 @@ services:
     environment:
         POSTGRES_DB: procrastinate
         POSTGRES_PASSWORD: password
+    command: ["-c", "fsync=off"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - postgres
     user: ${UID:-0}:${GID:-0}
   postgres:
-    image: postgres:10
+    image: postgres:12
     ports: ["5432:5432"]
     environment:
         POSTGRES_DB: procrastinate


.github/workflows/main.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:10
+        image: postgres:12.4
```
ewjoachim, line 56: 12.4 is 12 and latest. Why 12.4 rather than 12 or latest ?

.github/workflows/main.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           --health-timeout 5s
         ports:
           - 5432:5432
+        command: ["-c", "fsync=off"]
```
ewjoachim, line 67: Ah ? I thought you said this didn't change anything ?

ewjoachim (approved):
We can discuss the remaining details but :+1: already.

.github/workflows/main.yml
ewjoachim: Ah, just saw the table.

.github/workflows/main.yml
elemoine: Yeah, 12, 12.4 and latest are tags to the same image. Using `postgres:12` is what makes more sense I think. I'll change it if you agree.

.github/workflows/main.yml
elemoine: It's not a huge win, but it looks like we do save a few seconds for the entire test suite.

.github/workflows/main.yml
ewjoachim: I think I'd even agree with "latest" but 12 is sufficiently safe :)

elemoine:
@anayrat has requested that I test with postgres:11 as well. I've added a column with the postgres:11 figures. They are more or less the same as postgres:10.

.github/workflows/main.yml
elemoine: I assumed that you'd agree and change it to `postgres:12` :-)

anayrat:
Nice, I just wonder which changes in 12 give such improvement :)

ewjoachim:
Strangely, we only gained 15s on the CI apparently.

elemoine:
@anayrat to be sure I did a quick test outside Procrastinate. The test involves a sequence of database creations/deletions:


```bash
#!/bin/bash
n=1000
while [[ $n -gt 0 ]]; do
    createdb --template=procrastinate_template procrastinate_test && dropdb procrastinate_test
    n=$((n-1))
done
```

(`procrastinate_template` is an empty database with the Procrastinate schema applied.)

With postgres:10:

```bash
$ time bash test.sh 

real    7m3.034s
user    0m57.130s
sys     0m11.143s
```

With postgres:12:

```bash
$ time bash test.sh                                                                                                          
                                                                                                                                                                                               
real    2m0.332s                                                                                                                                                                               
user    0m46.823s                                                                                                                                                                              
sys     0m9.245s
```

Issue #242, opened 2020-06-09
Use  setuptools_scm in setup.py


ewjoachim:
Our setup.py makes exactly what setuptools_scm provides, so we could get rid of some code for a part that is completely process boilerplate.

I think it could be nice to switch to it.

https://pypi.org/project/setuptools-scm/

ewjoachim:
[Dunamai](https://github.com/mtkennerly/dunamai) seems nice too. Even nice maybe.

ewjoachim:
Closed by #319

PR #318, opened 2020-09-18
Use dunamai to get version from Git

elemoine:
Refs https://github.com/peopledoc/procrastinate/issues/242

Use dunamai to get version from Git.

The main issue with this is that we need to have dunamai installed prior to being able to do `pip install -r requirements.txt` (or `pip install -e .`).

Closes #<ticket number>

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,7 +112,7 @@ jobs:
         run: |
           set -xeu
           python -VV
-          pip install setuptools wheel
+          pip install setuptools wheel dunamai
       - name: Build the wheel
         run: python setup.py sdist bdist_wheel
       - name: Publish to PyPI
--- a/dev-env
+++ b/dev-env
@@ -22,6 +22,7 @@ echo ""
 
 if ! ls .venv/lib/python*/site-packages/procrastinate.egg-link 1>/dev/null 2>&1; then
     echo "Installing package for development"
+    pip install dunamai
     pip install -r requirements.txt
 fi
 
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 import contextlib
 import os
 import pathlib
-import subprocess
 
 import setuptools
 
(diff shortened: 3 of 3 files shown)


ewjoachim:
We can use `setup_requires`, but I seem to recall this has implications.
Would it be easier if we had switched to poetry ?

elemoine:
> We can use `setup_requires`, but I seem to recall this has implications.

Me too, but I can't remember the details.

elemoine:
Setuptools uses `easy_install` to fullfill the dependencies specified with `setup_requires`. Meaning that pip has no way to control how these dependencies are locaced. For example an `index-url` option set in `pip.conf` won't be taken into account. I'd rather stay away from `setup_requires` :smile: 

Also, in our case we import `dunamai` before the `setup` function is called, so `setup_requires` won't help us.

So either we require and document that `dunamai` be installed prior to executing `setup.py`, or we just consider that our own `git describe --tags` way of getting the version is good enough :-) Another alternative would be to use `versioneer`.

ewjoachim:
Not a fan of versionneer that leaves some code in the repo for which we don't have ownership :/

elemoine:
> Not a fan of versionneer that leaves some code in the repo for which we don't have ownership :/

I know you don't like it :-) But it just works!

ewjoachim:
Putting this in the `pyproject.toml` maybe ?
```
[build-system]
requires = ["dunami"]
build-backend = "setuptools.build_meta"
```

ewjoachim:
And compared to setuptools-scm ? (which advocates installation trhough pyproject.toml https://pypi.org/project/setuptools-scm/)

mgu:
maybe it would be useful to add `"setuptools >= 40.6.0"` to the `requires`

ewjoachim:
superseded by #319

PR #319, opened 2020-09-22
Use setuptools-scm to get migration from Git

mgu:
Refs #242

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)


.github/workflows/main.yml
```diff
@@ -112,7 +112,7 @@ jobs:
         run: |
           set -xeu
           python -VV
-          pip install setuptools wheel
+          pip install setuptools wheel "setuptools-scm[toml]"
```
ewjoachim, line 115: Is it needed ? I thought the point of putting that in pyproject.toml was that it was THE right place to put it.

Note: this applies to the whole line :)

dev-env
```diff
@@ -22,6 +22,7 @@ echo ""
 
 if ! ls .venv/lib/python*/site-packages/procrastinate.egg-link 1>/dev/null 2>&1; then
     echo "Installing package for development"
+    pip install "setuptools-scm[toml]"
```
ewjoachim, line 25: idem ?

.github/workflows/main.yml
mgu: In my tests I had to add it, but I don't know exactly why the build-system was not used when building the sdist/bdist_wheel

ewjoachim:
@mgu If I understand it correctly, there may be a link between pep517-enabled (pyproject.toml) projects and `pip install --editable`.

[This issue is closed](https://github.com/pypa/pip/issues/6314) so I'm not sure what the problem is.

ewjoachim:
After a long and fruitful discussion, we're adding [python-build](https://python-build.readthedocs.io/en/stable/index.html) to the mix.

setup.py
```diff
@@ -1,45 +1,3 @@
...
-        with open(version_filename, "w") as f:
-            f.write(version)
-        yield version
-    finally:
-        os.remove(version_filename)
-
-
-with ensure_version() as version:
-    setuptools.setup(version=version)
+setup()
```
ewjoachim, line 3:
```suggestion
setup(use_scm_version=True)
```

elemoine:
FWIW it looks great! :smile:

elemoine (approved):

elemoine:
Do not forget about the PR labels.

ewjoachim (approved):
Like my [FedEx driver](https://xkcd.com/1428/) says, move fast & break things
(if things go wrong, we'll iterate)

elemoine:
> Like my [FedEx driver](https://xkcd.com/1428/) says, move fast & break things
> (if things go wrong, we'll iterate)

So long that you're not an aircraft pilot… :-)

PR #322, opened 2020-09-25
Report test workflow success in Check API

### LINKED ISSUE #321: Separate test and release workflows
Currently, tests and release are all performed in the same workflow with Github Actions. This results in executing test workflow twice when releasing (once for merging into master, once for releasing), which in not necessary. We must ensure the release workflow does not execute tests beforehand.

thomasperrot:
Closes #321 

What we want to do:
- create two new checks: `push-success` and `pull_request-success`, that are created when their workflow succeed
- enable PR merge only when the above checks are successfull (need to configure that in the repository settings)
- enable deploying only when `push-success` on the last master HEAD is succussfull

### Successful PR Checklist:

- [ ] Tests
  - [X] (not applicable?)
- [ ] Documentation
  - [X] (not applicable?)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,15 +92,12 @@ jobs:
         with:
           fail_ci_if_error: true
 
-  report-success:
-    name: Report success in Github Checks
+  report-status:
+    name: ${{ github.event_name }}-success
     runs-on: ubuntu-latest
     needs:
       - check-lint
       - tests
     steps:
-      - uses: LouisBrunner/checks-action@v0.1.0
-        with:
-          name: Test
-          token: ${{ secrets.GITHUB_TOKEN }}
-          conclusion: ${{ job.status }}
+      - name: Report success
+        run: echo 'Success !'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
           fail_ci_if_error: true
 
   report-status:
-    name: ${{ github.event_name }}-success
+    name: success
     runs-on: ubuntu-latest
     needs:
       - check-lint
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,7 +18,7 @@ jobs:
         id: wait-for-ci
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          checkName: push-success
+          checkName: success
       - name: Exit if CI did not succeed
         if: steps.wait-for-ci.outputs.conclusion != 'success'
         run: exit 1
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,41 @@
+name: Deploy
+
+on:
+  push:
+    tags:
+      - '*'
+  release:
+    types:
+      - published
+
+jobs:
+  deploy:
+    name: Publish package to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for tests to succeed
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        id: wait-for-ci
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: push-success
+      - name: Exit if CI did not succeed
+        if: steps.wait-for-ci.outputs.conclusion != 'success'
+        run: exit 1
+      - uses: actions/checkout@v2
+      - name: Setup Python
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -95,28 +92,15 @@ jobs:
...
-          pip install -U pip build
-      - name: Build the wheel
-        run: python -m build --sdist --wheel
-      - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@master
+      - uses: LouisBrunner/checks-action@v0.1.0
         with:
-          user: __token__
-          password: ${{ secrets.PYPI_TOKEN }}
+          name: Test
```
ewjoachim, line 104: I'd tend to call this "CI", like the workflow, to convey that it also ran linters & docs

ewjoachim:
Is this PR WIP ?

elemoine:
> Closes #322

Self reference?

thomasperrot:
Yes it is WIP, please do not review it now :speak_no_evil:

ewjoachim:
Is it still a draft ?

ewjoachim (approved):

ewjoachim:
Can we purposefully make a test break just to be sure it reports correctly ?

thomasperrot:
@ewjoachim I tried to break tests in an other repository, which I use to test Github Actions. The commit that breaks tests results in the following actions : https://github.com/thomasperrot/weather/actions/runs/284146480. 

On the plus side, the CI properly fails, and the last job (which creates a check-run) does not execute. On the down side, the last job name is not properly reported when the CI fails, because its name is not evaluated. It is a minor inconvenient, which occurs only when the CI fails.

.github/workflows/ci.yml
```diff
@@ -95,28 +92,12 @@ jobs:
         with:
           fail_ci_if_error: true
 
-  deploy:
-    name: Publish package to PyPI
+  report-status:
+    name: ${{ github.event_name }}-success
```
ewjoachim, line 96: why not just use `success` here ?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+  deploy:
+    name: Publish package to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for tests to succeed
+        uses: fountainhead/action-wait-for-check@v1.0.0
+        id: wait-for-ci
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: push-success
```
ewjoachim, line 21: (and success here too)

ewjoachim (commented):
I'm not seeing a check that the tag is on master. If I push a tag on a branch, will it get released?

.github/workflows/ci.yml
thomasperrot: I want to differenciate the success for the `pull_request`event, and the one for the `push` event. They need different name if we want to make them mandatory before merging the PR.

.github/workflows/deploy.yml
thomasperrot: Deploy only runs when publishing a new release. It must check the CI correctly passed on the HEAD of master. The CI on master run because of a `push` event when merge in PR, thus the associated CI has `push-success` name.

thomasperrot:
Yes, it will get released. Do we need to trigger the deploy workflow when pushing a new tag ? If not, I suggest we remove tag support in deploy workflow, and trigger it only when releasing.

.github/workflows/ci.yml
thomasperrot: EDIT: I just tested what would happen if we use just a single check for the CI workflow. Turns out we end up with only one check in the branch settings : 
image
And when activating only that check, we have 2 mandatory checks in the CI workflow (one for push event, one for pull_request event).
image
I just check, and the PR can not be merged if one of them is missing. So I think we are good with `success` check, as you suggest. I pushed a new commit to fix it.

.github/workflows/deploy.yml
thomasperrot: EDIT: I changed it for `success`

ewjoachim:
I'm thinking about it and I think it's fine. We can create a GitHub release or a git tag on any commit. Given external people can put tags on their PR but it's tags on their fork, not ours, I think we're good.

(If you can try and fork the project, push a PR from your fork with a tag and check what happens, I think I'll be a bit reassured)

ewjoachim (approved):

PR #327, opened 2020-10-02
Fix the AppNotOpen exception message

### LINKED ISSUE #313: The AppNotOpen exception message is incorrect
The message of the `AppNotOpen` exception is

```
    App was not open. Procrastinate App needs to be opened using:                                                                                                                              
                                                                                                                                                                                               
    - ``App.open``,                                                                                                                                                                            
    - ``App.open_async``,                                                                                                                                                                      
    - ``with App():``,                                                                                                                                                                         
    - ``async with App():``.
```

That's not correct, as `async with App()` does not work. `async with App().open_async()` is to be used instead.

elemoine:
Closes #313 

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -44,10 +44,10 @@ class AppNotOpen(ProcrastinateException):
     """
     App was not open. Procrastinate App needs to be opened using:
 
-    - ``App.open``,
-    - ``App.open_async``,
-    - ``with App():``,
-    - ``async with App():``.
+    - ``app.open()``,
+    - ``await app.open_async()``,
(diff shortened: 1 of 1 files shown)


ewjoachim (approved):
I see you've decided to go with `app` and not `App()`. I'm fine with that, and by no means I'd request a change, but any explanation as to why? I'm guessing that it's to avoid giving people the impression they need to instantiate an app for that ? but at the expense of making it slightly less obvious what type `app` is ? Honestly I'm very fine with that.

PR #329, opened 2020-10-05
Fix the procrastinate_defer_periodic_job SQL function

### LINKED ISSUE #328: Postgres DeadlockDetected error
My team has moved to using "periodic tasks", and we got this Postgres error:

```
DeadlockDetected

deadlock detected
DETAIL:  Process 57350 waits for ShareLock on transaction 528284087; blocked by process 48792.
Process 48792 waits for ShareLock on transaction 528284086; blocked by process 57350.
HINT:  See server log for query details.
CONTEXT:  while deleting tuple (2,2) in relation "procrastinate_periodic_defers"
SQL statement "DELETE
        FROM procrastinate_periodic_defers
        WHERE
            _job_id IS NOT NULL
            AND procrastinate_periodic_defers.task_name = _task_name
        ...
```

I don't have more for the moment, but I'll provide more details as soon as I can.

elemoine:
Concurrent DELETE's may result in deadlocks because the concurrent transactions may not acquire the share locks on the to-be-deleted rows in a consistent order.

This problem can be worked out by using USING with a SELECT ORDER BY FOR UPDATE guaranteeing that share locks are acquired in a consistent order.

Relevant links:

* https://github.com/Kinto/kinto/issues/1308
* https://github.com/Kinto/kinto/pull/1312
* https://stackoverflow.com/questions/45621347/how-can-two-delete-queries-deadlock-in-postgres
* https://stackoverflow.com/questions/51972202/are-the-rows-locked-in-order-in-a-select-order-by-for-update-statement

Closes #328 

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?) We have no SQL tests for the moment.
- [ ] Documentation
  - [x] (not applicable?)
- [x] Had a good time contributing? So much :smile:


procrastinate/sql/migrations/delta_0.15.2_001_fix_procrastinate_defer_periodic_job.sql
```diff
@@ -0,0 +1,52 @@
+CREATE OR REPLACE FUNCTION procrastinate_defer_periodic_job(
```
ewjoachim, line 1: I think it's ok because we don't change the function signature but neithertheless, I'd be reassured if we could delete the function, as a rule. `CREATE OR REPLACE FUNCTION` doesn't work in some cases and even if I'm 90% we're not in one of those cases, I4d rather not take useless risks.

ewjoachim (approved):

procrastinate/sql/migrations/delta_0.15.2_001_fix_procrastinate_defer_periodic_job.sql
elemoine: Agree. I was thinking about the same.

ewjoachim (approved):

PR #336, opened 2020-10-16
Split procrastinate_finish_job into two functions

### LINKED ISSUE #332: Split the procrastinate_finish_job SQL function into two functions
The `procrastinate_finish_job` SQL function looks like this currently:

https://github.com/peopledoc/procrastinate/blob/f831565b00d67afa7a4291d734c8fe85074a360c/procrastinate/sql/schema.sql#L175-L185

`scheduled_at` only makes sense for retries.

So @ewjoachim and I suggest that we split this function into two functions:

```sql
CREATE FUNCTION procrastinate_finish_job(job_id integer, end_status procrastinate_job_status) RETURNS void
    LANGUAGE plpgsql
    AS $$
BEGIN
    UPDATE procrastinate_jobs
    SET status = end_status,
        attempts = attempts + 1
    WHERE id = job_id;
END;
$$;

CREATE FUNCTION procrastinate_retry_job(job_id integer, retry_at timestamp with time zone) RETURNS void
    LANGUAGE plpgsql
    AS $$
BEGIN
    UPDATE procrastinate_jobs
    SET status = 'todo',
        attempts = attempts + 1,
        scheduled_at = retry_at
    WHERE id = job_id;
END;
$$;
```

And call the appropriate function from the worker code.

elemoine:
Closes #332 

This PR splits the `procrastinate_finish_job` SQL functions into two functions: `procrastinate_finish_job` and `procrastinate_retry_job`.

To be discussed:

For backward compatibility reasons I've kept `procrastinate_finish_job` as-is, and added `procrastinate_finish_job_1`. In this way one can apply the schema migration (the `delta_0.16.0_001` SQL file in this PR) before upgrading the Procrastinate code.

I think that the schema of Procrastinate version N should work for any previous version (< N) of Procrastinate. This is to make it easy to upgrade Procrastinate, and in a "blue/green" way. And I think we should do that until 1.0. In 1.0 we would remove all the backward compatibility stuff from the schema. For example, we would rename `procrastinate_finish_job_1` to `procrastinate_finish_job`, and remove the old `procrastinate_finish_job` function. And we would apply the same backward compatibility principle within the 1.x.y series (until 2.0.0).

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/procrastinate/sql/queries.sql
+++ b/procrastinate/sql/queries.sql
@@ -47,8 +47,12 @@ WHERE id IN (
 )
 
 -- finish_job --
--- Stop a job, free the lock and record the relevant events
-SELECT procrastinate_finish_job(%(job_id)s, %(status)s, %(scheduled_at)s);
+-- Finish a job, changing it from "doing" to "succeeded" or "failed"
+SELECT procrastinate_finish_job(%(job_id)s, %(status)s);
+
+-- retry_job --
+-- Retry a job, changing it from "doing" to "todo"
+SELECT procrastinate_retry_job(%(job_id)s, %(retry_at)s);
 
 -- listen_queue --
 -- In this one, the argument is an identifier, shoud not be escaped the same way
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -152,25 +152,26 @@ class Worker:
             extra=context.log_extra(action="loaded_job_info"),
         )
 
-        status = jobs.Status.FAILED
-        next_attempt_scheduled_at = None
+        status, retry_at = None, None
         try:
             await self.run_job(job=job, worker_id=worker_id)
             status = jobs.Status.SUCCEEDED
         except exceptions.JobRetry as e:
-            status = jobs.Status.TODO
-            next_attempt_scheduled_at = e.scheduled_at
+            retry_at = e.scheduled_at
         except exceptions.JobError:
-            pass
+            status = jobs.Status.FAILED
         except exceptions.TaskNotFound as exc:
+            status = jobs.Status.FAILED
             self.logger.exception(
                 f"Task was not found: {exc}",
                 extra=context.log_extra(action="task_not_found", exception=str(exc)),
             )
         finally:
-            await self.job_manager.finish_job(
-                job=job, status=status, scheduled_at=next_attempt_scheduled_at
-            )
+            if retry_at:
+                await self.job_manager.retry_job(job=job, retry_at=retry_at)
+            else:
+                assert status is not None
+                await self.job_manager.finish_job(job=job, status=status)
 
             self.logger.debug(
                 f"Acknowledged job completion {job.call_string}",
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -333,6 +333,13 @@ care of removing the old version of the function:
 In this way, you provide the new SQL code, the compatibility layer, and the migration
 for the removal of the compatibility layer.
 
+.. note::
+
+    The migration scripts that remove the SQL compatibility code are to be added to the
+    ``future_migrations`` directory instead of the ``migrations`` directory. And it will
+    be the responsibility of Procrastinate maintainers to move them to the
+    ``migrations`` directory after the next major release.
+
 Migration tests
 ^^^^^^^^^^^^^^^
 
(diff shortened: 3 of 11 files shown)


procrastinate/worker.py
```diff
@@ -152,25 +152,26 @@ async def process_job(self, job: jobs.Job, worker_id: int = 0) -> None:
...
+            status = jobs.Status.FAILED
             self.logger.exception(
                 f"Task was not found: {exc}",
                 extra=context.log_extra(action="task_not_found", exception=str(exc)),
             )
         finally:
-            await self.job_manager.finish_job(
-                job=job, status=status, scheduled_at=next_attempt_scheduled_at
-            )
+            if retry_at:
```
AlexisCothenet, line 170: juste une question svp :)  : est-ce que cette gestion faite en python pourrait pas être faite plutôt dans le schema ?
et même que la procédure stockée en db fasse elle même un pg cron pour faire un notify à la date voulue ?

ewjoachim: Hey :) This is a public project, I know you know I speak French, but it's easier for other people if we try and stick to English ;)

As of today, determining if a task should be retried or not is done in Python, because it is an operation that depends on the Python outcome of running the task. Could you tell us more on what you're suggesting ? Also, I'm not really sure it belongs to this PR, but maybe I misunderstood your comment.

AlexisCothenet: Hello. yes sorry ;) just wondering if the logic to retry of not could be handled in the 
_procrastinate_finish_job_   stored procedure. So that, depending if _next_scheduled_at_ is null or not , the procedure would call _procrastinate_finish_job_1_ or _procrastinate_retry_job_ .
Moreother, inside the _procrastinate_retry_job_, we could imagine that a notify (call to procrastinate_notify_queue()) could be done at the right _next_scheduled_at_  thanks to a pg cron ?

elemoine: > just wondering if the logic to retry of not could be handled in the procrastinate_finish_job stored procedure. So that, depending if next_scheduled_at is null or not , the procedure would call procrastinate_finish_job_1 or procrastinate_retry_job

Well, that's basically how it's implemented currently (before this PR). Except that the the status should always be passed, even when `scheduled_at` is specified. See https://github.com/peopledoc/procrastinate/blob/ddd2d19794227b4178084d59f56f3f38f09873c8/procrastinate/sql/schema.sql#L175-L185.

And we thought that it would make for a cleaner and more explicit API if we had two separate functions. We want to distinguish the action of "finishing a job", which involves changing its status to either "succeeded" or "failed", from the action of "retrying a job", which involves resetting its status to "todo", with a scheduling in the future.

> Moreother, inside the _procrastinate_retry_job_, we could imagine that a notify (call to procrastinate_notify_queue()) could be done at the right _next_scheduled_at_ thanks to a pg cron ?

Are you referring to [`pg_cron`](https://github.com/citusdata/pg_cron)?

AlexisCothenet: Hello. thanks for your answer. yes, i'm refering to pg_cron.
What i m' suggesting is something like : 

```
CREATE FUNCTION procrastinate_finish_job (job_id integer, end_status procrastinate_job_status, next_scheduled_at timestamp with time zone) RETURNS void
BEGIN
   IF (next_scheduled_at IS NULL) 
         CALL procrastinate_finish_job_1(job_id, end_status)
  ELSE
         CALL procrastinate_retry_job(job_id, next_scheduled_at)
END
```

so that to minimize the code that the client has to do. The client only has to call _procrastinate_finish_job_ 
Remark : if guess we want to retry only on on failed status ?
Remark2: calling only the retry procedure with status='todo', you won't have anymore the _procrastinate_events_ failed status triggered by [ref](https://github.com/peopledoc/procrastinate/blob/ddd2d19794227b4178084d59f56f3f38f09873c8/procrastinate/sql/schema.sql#L260)

elemoine:
This is now ready for review!

ewjoachim:
Can you add a note in https://github.com/peopledoc/procrastinate/blob/ele_finish-job/CONTRIBUTING.rst#L512-L525 ? Thanks :)

procrastinate/sql/queries.sql
```diff
@@ -48,8 +48,17 @@ WHERE id IN (
 
 -- finish_job --
 -- Stop a job, free the lock and record the relevant events
+-- Old version, kept for backward compatibility reasons
```
ewjoachim, line 51:
```suggestion
-- Deprecated, remove after 1.0.0
```

ewjoachim: No... It's actually not necessary to keep it.

queries.sql is at the version of the code, not the DB. We'll never need that.

elemoine: Oh yes. Good catch!

elemoine:
> Can you add a note in
> 
> https://github.com/peopledoc/procrastinate/blob/001647469c7326daea1e35d78886a197469225fc/CONTRIBUTING.rst#L512-L525

Done. I hope this is sufficient :-)

ewjoachim (approved):

PR #342, opened 2020-11-06
Document new rules for database migrations

### LINKED ISSUE #341: New rules for migrations
See #100 for our first iteration.

## Why

We need to provide a way to run migrations while procrastinate runs, which means we need each each version of the schema to be compatible with 2 at least versions of the code (and likewise, each version of the code to be compatible with at least 2 versions of the schema).

## How

- By setting clear rules on what version of the schema should work with what version of the code
- By testing that it's the case
- By documenting the upgrade strategy

We can be quite limiting at start (e.g. only upgrading versions 1 by 1 is supported)

Let's flesh out proposals below.

elemoine:
Closes #341.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -248,6 +248,89 @@ that the ``migrations`` directory already includes a migration script whose seri
 number is ``001`` for that release number. In that case, if you need to add a migration
 script, its name will start with ``delta_1.0.1_002_``.
 
+Backward-compatibility
+^^^^^^^^^^^^^^^^^^^^^^
+
+As a Procrastinate developer, the changes that you make to the Procrastinate database
+schema must be compatible with the Python code of previous Procrastinate versions.
+
+For example, let's say that the current Procrastinate database schema includes an SQL
+function
+
+.. code-block:: sql
+
+    procrastinate_func(arg1 integer, arg2 text, arg3 timestamp)
+
+that you want to change to
+
+.. code-block:: sql
+
+    procrastinate_func(arg1 integer, arg2 text)
+
+The straightforward way to do that would be to edit the ``schema.sql`` file and just
+replace the old function by the new one, and add a migration script that removes the old
+function and adds the new one:
+
+.. code-block:: sql
+
+    DROP FUNCTION procrastinate_func(integer, text, timestamp);
+    CREATE FUNCTION procrastinate_func(arg1 integer, arg2 text)
+    RETURNS INT
+    ...
+
+But if you do that you will break the Procrastinate Python code that uses the old
+version of the ``procrastinate_func`` function. The direct consequence of that is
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -250,6 +250,89 @@ running the following command and committing the results:
...
+
+    procrastinate_func(arg1 integer, arg2 text, arg3 timestamp)
+
+that you want to change to
+
+.. code-block:: sql
+
+    procrastinate_func(arg1 integer, arg2 text)
+
+The naive way to do that is to edit the ``schema.sql`` file and just replace the old
```
ewjoachim, line 272:
```suggestion
The straightforward way to do that would be to edit the ``schema.sql`` file and just replace the old
```
(that might make a line a bit too long, so you might need to balance the lines)

docs/howto/migrations.rst
```diff
@@ -56,3 +56,18 @@ that order:
     order. For example, ``delta_0.9.0_001_xxxxx.sql`` is to be applied before
     ```delta_0.10.0_001_xxxxx.sql`` (because version 0.9.0 is lower than
     0.10.0).
+
+If you want to upgrade from one Procrastinate major version to another, and if you care
+about service continuity, you'll need to make intermediate upgrades. For example, let's
+say you want to upgrade from Procrastinate 1.6.0 to Procrastinate 3.2.0, here are the
+steps you will need to follow:
```
ewjoachim, line 63:
```suggestion
If you want to upgrade from one Procrastinate major version to another, the easiest way is to follow those steps:

- Shut down the services that use procrastinate: both services that defer tasks and workers
- Apply all the SQL migration scripts (1.6.0 ≤ version < 3.2.0)
- Upgrade your code to the new Procrastinate version
- Start all the services.

This, as you've noticed, only works if you're able to stop your services. If you care
about service continuity, you'll need to make intermediate upgrades. For example, let's
say you want to upgrade from Procrastinate 1.6.0 to Procrastinate 3.2.0, here are the
steps you will need to follow:
```

ewjoachim (approved):
Very good!

It doesn't have to be in this PR, but I realize that we miss clear instruction how to run the migrations (hands on, not theoretically).

elemoine:
> It doesn't have to be in this PR, but I realize that we miss clear instruction how to run the migrations (hands on, not theoretically).

We just have this:

> The migration scripts are pure-SQL scripts, meaning that they may be applied to the
database using any PostgreSQL client, including ``psql`` and ``PGAdmin``.

ewjoachim:
Yes, let's add a snippet :)

k4nar (commented):
LGTM, thanks!

PR #349, opened 2020-11-13
Remove Admin and move its methods to JobManager

### LINKED ISSUE #333: Admin and JobManager are ambiguous
Currently, the application has both an `admin` and a `job_manager`. But those two objects are very much the same: they provide functions for managing jobs in the database. They convey the same objective.

After discussion, @ewjoachim and I suggest to remove `Admin` and keep `JobManager` only. And, in `JobManager`, mark the "admin" functions as public in the API documentation.

ewjoachim:
Closes #333

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)


tests/conftest.py
```diff
@@ -159,11 +159,6 @@ def job_manager(app):
     return app.job_manager
 
 
-@pytest.fixture
-def serial():
-    return itertools.count(1)
-
-
```
elemoine, line 166: Why did you remove this? The `job_factory` fixture still uses it…

tests/conftest.py
```diff
@@ -190,6 +185,16 @@ def factory(**kwargs):
...
+@pytest.fixture
+def deferred_job_factory(job_manager, job_factory):
+    async def factory(*, manager=job_manager, **kwargs):
+        job = job_factory(**kwargs)
+        job_id = await manager.defer_job_async(job)
+        return job.evolve(id=job_id)
+
+    return factory
+
+
```
elemoine, line 197: I also don't see the reason for this new fixture. It's not used currently.

procrastinate/manager.py
```diff
@@ -151,12 +234,212 @@ async def retry_job(
...
+                "jobs_count": row["jobs_count"],
+                "todo": row["stats"].get("todo", 0),
+                "doing": row["stats"].get("doing", 0),
+                "succeeded": row["stats"].get("succeeded", 0),
+                "failed": row["stats"].get("failed", 0),
+            }
+            for row in await self.connector.execute_query_all_async(
+                query=sql.queries["list_tasks"],
+                queue_name=queue,
+                name=task,
```
elemoine, line 412: I should be `task_name=task` here.

tests/conftest.py
ewjoachim: I wonder if it wouldn't make more sense to make an undefered job, and then let the manager defer it. The InMemory backend already maintains a counter.

(Anyway, WIP :D )

elemoine:
I added a WIP3 commit (a wommit in our new jargon :) that fixes the tests.

tests/conftest.py
elemoine: I see. But with this the object provided by the new `job_factory` fixture (that you'd named `deferred_job_factory`) would a coroutine function. So that would imply doing `job = await job_factory(…)` instead of `job = job_factory(…)` everywhere we use that fixture. That's a lot of changes. I'd suggest that, if we want to do that, we do it with a separate PR.

elemoine:
I fixed the tests, and improved the doc strings making sure that the generated API reference looks good.

I think this is ready for merging.

ewjoachim (commented):
:+1: (cannot approve my own PR)

elemoine (approved):

PR #357, opened 2020-11-30
Add the ability to personalize log format by environment variable

ignaciocabeza:
This PR provides the ability to set log format with an environment variable `PROCRASTINATE_LOG_FORMAT` or use the default format: `%(asctime)s %(levelname)s %(name)s:%(lineno)s %(message)s`

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -21,7 +21,7 @@ CONTEXT_SETTINGS = {
     "auto_envvar_prefix": ENV_PREFIX,
 }
 
-LOG_FORMAT = os.environ.setdefault(f"{ENV_PREFIX}_LOG_FORMAT", "%(asctime)s %(levelname)s %(name)s:%(lineno)s %(message)s")
+LOG_FORMAT = os.environ.setdefault(f"{ENV_PREFIX}_LOG_FORMAT", logging.BASIC_FORMAT)
 
 
 def get_log_level(verbosity: int) -> int:
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -21,6 +21,8 @@ CONTEXT_SETTINGS = {
     "auto_envvar_prefix": ENV_PREFIX,
 }
 
+LOG_FORMAT = os.environ.setdefault(f"{ENV_PREFIX}_LOG_FORMAT", "%(asctime)s %(levelname)s %(name)s:%(lineno)s %(message)s")
+
 
 def get_log_level(verbosity: int) -> int:
     """
@@ -38,7 +40,7 @@ def click_set_verbosity(ctx: click.Context, param: click.Parameter, value: int) 
 def set_verbosity(verbosity: int) -> None:
     level = get_log_level(verbosity=verbosity)
     logging.basicConfig(level=level)
-    level_name = logging.getLevelName(level)
+    level_name = logging.getLevelName(level, format=LOG_FORMAT)
     logger.debug(
         f"Log level set to {level_name}",
         extra={"action": "set_log_level", "value": level_name},
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -39,8 +39,8 @@ def click_set_verbosity(ctx: click.Context, param: click.Parameter, value: int) 
 
 def set_verbosity(verbosity: int) -> None:
     level = get_log_level(verbosity=verbosity)
-    logging.basicConfig(level=level)
-    level_name = logging.getLevelName(level, format=LOG_FORMAT)
+    logging.basicConfig(level=level, format=LOG_FORMAT)
+    level_name = logging.getLevelName(level)
     logger.debug(
         f"Log level set to {level_name}",
         extra={"action": "set_log_level", "value": level_name},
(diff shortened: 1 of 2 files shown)


ewjoachim:
Hello :) And thanks for your contribution !

I like the idea, but if we did it as a command-line argument, thanks to the Click magic, it would be available both as CLI arg & env var. Would that be OK for you ? I can make the change if you want :)

Also, it could be worth adding a small test like this one:
https://github.com/peopledoc/procrastinate/blob/99c2b5cbaa2b1e33852478f32101486e9bfe3f13/tests/unit/test_cli.py#L18-L28

Finally: it seems the CI didn't start. I think I know why, I'll try and fix it.

procrastinate/cli.py
```diff
@@ -38,7 +40,7 @@ def click_set_verbosity(ctx: click.Context, param: click.Parameter, value: int)
 def set_verbosity(verbosity: int) -> None:
     level = get_log_level(verbosity=verbosity)
     logging.basicConfig(level=level)
-    level_name = logging.getLevelName(level)
+    level_name = logging.getLevelName(level, format=LOG_FORMAT)
```
ewjoachim, line 43: [`getLevelName`](https://docs.python.org/3/library/logging.html#logging.getLevelName) has no argument `format`. Did you mean to add this to basicConfig instead ?

ignaciocabeza: Yes! 🤦‍♂️ Good catch. I copy/paste only the arguments from other project. I'll fixit later

ewjoachim:
If you rebase this branch or merge master in it, the CI should run. I might be able to do it if you want.

ignaciocabeza:
> Hello :) And thanks for your contribution !
> 
> I like the idea, but if we did it as a command-line flag, thanks to the Click magic, it would be available both as CLI option & env var. Would that be OK for you ? I can make the change if you want :)
> 
> Also, it could be worth adding a small test like this one (or more probably fixing this one which probably fails as of now):
> https://github.com/peopledoc/procrastinate/blob/99c2b5cbaa2b1e33852478f32101486e9bfe3f13/tests/unit/test_cli.py#L18-L28
> 
> Documentation might help make this feature findable, though we'll have automated doc if we add it as a click option.
> 
> Finally: it seems the CI didn't start. I think I know why, I'll try and fix it.

Hi, i think using a new CLI option will require a little bit of refactor & testing around `set_verbosity` fn. I'm happy if you want to do it, I believe it would take more time if I do it. 

Offtopic: any reason on why 0.17 is not in pypi? 

BTW, thanks for the library. It was a little bit hard to implement in our project but now it's working good!

ewjoachim:
> BTW, thanks for the library. It was a little bit hard to implement in our project but now it's working good!

Oh, do you want to share the pain points and what would have made it easier for you to integrate the lib ?

ewjoachim:
> Hi, i think using a new CLI option will require a little bit of refactor & testing around set_verbosity fn. I'm happy if you want to do it, I believe it would take more time if I do it.

Ok then, we can take it from here !

ewjoachim:
> Offtopic: any reason on why 0.17 is not in pypi?

Oh no, you're right, there's a problem in our deployment pipeline !

ewjoachim (approved):

ewjoachim:
Congratulations on your first successful PR in procrastinate ! Please feel free to keep going :)

ewjoachim:
I just uploaded 0.17.0 manually to PyPI, and there's a PR (#360 ) to fix the pipeline.
https://pypi.org/project/procrastinate/0.17.0/

ignaciocabeza:
> > BTW, thanks for the library. It was a little bit hard to implement in our project but now it's working good!
> 
> Oh, do you want to share the pain points and what would have made it easier for you to integrate the lib ?

I reckon that some of the problems were caused because our architecture and sorry if i'm not very specific about the pain points.

The first problem was the integration with Django. Use django models inside a task is not possible, so we had to write raw SQL inside our task and create test for that raw SQL. Also not a big fan of writing sql in 2020. 
In addition, it was not possible to use modules with django `models` imports inside, so we had to refactor a good amount of code.

Testing was tricky/challenging. For example, we had to mock/override procrastinate `execute_query_async`  and also refactor our mocked external services. Maybe @frinzekt can add more info on this.

In conclusion, I think it would be good to make procrastinate more django friendly but I know that is not depending only on procrastinate. 

Documentation is also flaky on testing and django:
- https://procrastinate.readthedocs.io/en/stable/howto/testing.html?highlight=test
- https://procrastinate.readthedocs.io/en/stable/howto/django.html?highlight=django#use-procrastinate-in-a-django-application
- https://procrastinate.readthedocs.io/en/stable/reference.html?highlight=django#django

I would like to see documentation on how to test without a In Memory connector.

Besides the development problems, procrastinate is working good and steady.

elemoine:
@ignaciocabeza thanks a lot for taking the time to provide feedback.

PR #360, opened 2020-12-01
Fix deployment pipeline

ewjoachim:
The deployment pipeline was accidentally broken in #322, here's the fix.

### Successful PR Checklist:

- [ ] Tests < manually tested
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,12 +28,9 @@ jobs:
         with:
           python-version: 3.x
       - name: Install dependencies
-        run: |
-          set -xeu
-          python -VV
-          pip install setuptools wheel
+        run: pip install build
       - name: Build the wheel
-        run: python setup.py sdist bdist_wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -29,11 +29,9 @@ jobs:
           python-version: 3.x
       - name: Install dependencies
         run: |
-          set -xeu
-          python -VV
-          pip install setuptools wheel
+          pip install build
```
thomasperrot, line 32: You can write that in a single line : `run: pip install build`

thomasperrot (approved):

.github/workflows/deploy.yml
ewjoachim: true

elemoine (approved):
Thanks for taking care of that @ewjoachim!

PR #362, opened 2020-12-01
Fix dev-env

ewjoachim:
Tiny tiny fix: when using an external venv, we need to check that procrastinate is installed on that venv.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)


elemoine (approved):
Good catch!

thomasperrot (approved):

PR #368, opened 2020-12-13
Fix README async example, make both `sync` and `async` example more complete

### LINKED ISSUE #364: `README` outdated
Hey guys,

love the idea, been testing it out a little for use at work. I tried to follow the examples from the `README.md`, as they showed programmatic usage in comparison to the CLI based approach I was able to find in your documentation.

Sadly, the examples do not work (out of the box):
```py
# Somewhere in your program, run a worker asynchronously
worker = procrastinate.Worker(
    app=app,
    queues=["sums"]
)
```

Fails, since `Worker` is not exported by `procrastinate`.

This:
```py
await worker.run_async()
```

Fails, since `AttributeError: 'Worker' object has no attribute 'run_async'`.

This is not urgent to me, I've been able to work around it, but I think for a quick try out it would be nice if `README` examples work (:

Thanks for the library!

BracketJohn:
Closes #364

### Successful PR Checklist:
- [ ] Tests
  - [x] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

I've made the following changes/fixes:
1. Fixed the `async` README example, that was still relying on non-existent `procrastinate.Worker`
2. Extended both examples to:
  - include necessary imports
  - include DB connection boilerplate: Easier for newbies (like me) to approach and understand how they can connect to their DB
  - use context managers instead: Generally regarded as good practice - will automatically tear down `app.open` in this case

Overall, both the `sync` and `async` examples can no be copy pasted into any script/repl and should run from top to bottom.

Looking forward to your review.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,7 +66,7 @@ Here's an example:
         sum.defer(a=3, b=5)
 
         # Somewhere in your program (actually, often a different
-        # program than the one defering jobs for execution), run a worker
+        # program than the one deferring jobs for execution), run a worker
         app.run_worker(queues=["sums"])
 
 The worker will run the job, which will create a text file
--- a/README.rst
+++ b/README.rst
@@ -47,24 +47,32 @@ Here's an example
 .. code-block:: python
 
     # mycode.py
+    import procrastinate
 
     # Make an app in your code
-    app = procrastinate.App(connector=procrastinate.AiopgConnector())
+    app = procrastinate.App(connector=procrastinate.AiopgConnector(host="localhost", user="postgres", password="password"))
 
     # Open the connection to the database
-    app.open()
+    with app.open():
 
-    # Then define tasks
-    @app.task(queue="sums")
-    def sum(a, b):
-        with open("myfile", "w") as f:
-            f.write(str(a + b))
+        # Automatically apply the `procrastinate` schema to the database
+        try:
+            app.schema_manager.apply_schema()
+        except procrastinate.exceptions.ConnectorException:
+            # Schema has already been applied
+            pass
 
-    # Launch a job
-    sum.defer(a=3, b=5)
+        # Then define tasks
+        @app.task(queue="sums")
+        def sum(a, b):
+            with open("myfile", "w") as f:
+                f.write(str(a + b))
 
-    # Somewhere in your program, run a worker
-    app.run_worker(queues=["sums"])
+        # Launch a job
+        sum.defer(a=3, b=5)
+
+        # Somewhere in your program, run a worker
+        app.run_worker(queues=["sums"])
 
 The worker will run the job, which will create a text file
 named ``myfile`` with the result of the sum ``3 + 5`` (that's ``8``).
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -47,24 +47,32 @@ Here's an example
...
     # mycode.py
+    import procrastinate
 
     # Make an app in your code
-    app = procrastinate.App(connector=procrastinate.AiopgConnector())
+    app = procrastinate.App(connector=procrastinate.AiopgConnector(host="localhost", user="postgres", password="password"))
 
     # Open the connection to the database
-    app.open()
+    with app.open():
```
ewjoachim, line 56: I think we wanted to keep the sync example as simple as possible, and the context manager is not mandatory in sync mode. But I'm totally OK leaving that if you think this helps understanding better how it works.

README.rst
```diff
@@ -47,24 +47,32 @@ Here's an example
...
-    @app.task(queue="sums")
-    def sum(a, b):
-        with open("myfile", "w") as f:
-            f.write(str(a + b))
+        # Automatically apply the `procrastinate` schema to the database
+        try:
+            app.schema_manager.apply_schema()
+        except procrastinate.exceptions.ConnectorException:
+            # Schema has already been applied
+            pass
```
ewjoachim, line 63: I understand the idea to have a single fully functional example, but if we go this way, I'd be much more comfortable to make sure the schema fails early when applied on a database that already has it.

I have a feeling we'll end up making things easier for everyone and implementing a very rudimentary migration engine :D

README.rst
BracketJohn: I suggest to leave it, as context managers have been "mainstream" for a long time, thus this code should still be pretty straight forward to understand. And it has the added benefit that all context managers have: Controlled resource setup and tear down.

README.rst
BracketJohn: I understand. What would you suggest instead? The following options come to mind:
* not setting up the schema at all
* raising in the `except` block
* setting up the schema with a `try/except` and adding a `NOTE: ` comment above it, that tells the reader that this will fail the second time around and should/must be only done once

Of course I'd be more than happy to get another suggestion from your side.

README.rst
```diff
@@ -47,24 +47,32 @@ Here's an example
 .. code-block:: python
 
     # mycode.py
+    import procrastinate
 
     # Make an app in your code
-    app = procrastinate.App(connector=procrastinate.AiopgConnector())
+    app = procrastinate.App(connector=procrastinate.AiopgConnector(host="localhost", user="postgres", password="password"))
```
elemoine, line 53: In my opinion the goal of the code snippets in the README is to show what Procrastinate is and the various constructs it provides, but they're not necessarily codes you can copy & paste and execute. We have the Quickstart for that. So I'd rather keep the README snippets simple and to the point, avoiding things like passing database parameters to the `AiopgConnector` constructor.

README.rst
elemoine: I am fine with using `with app.open()` in the snippets, but I think the `sum` task function should be declared outside the context.

README.rst
elemoine: As I said in a [previous comment](https://github.com/peopledoc/procrastinate/pull/368/files#r542528982) I'd like to keep the README snippets simple and to the point. So I would not add the setting up of the schema here in the code.

README.rst
```diff
@@ -47,24 +47,32 @@ Here's an example
...
+        def sum(a, b):
+            with open("myfile", "w") as f:
+                f.write(str(a + b))
+
+        # Launch a job
+        sum.defer(a=3, b=5)
 
-    # Somewhere in your program, run a worker
-    app.run_worker(queues=["sums"])
+        # Somewhere in your program, run a worker
```
elemoine, line 74: I'd be tempted to change this comment to this one:

```
# Somewhere in your program (actually, often a different program than the one launching jobs), run a worker
```

BracketJohn: Added 👍

BracketJohn:
Addressed all of your comments @elemoine, @ewjoachim - looking forward to the next round (:

README.rst
```diff
@@ -42,29 +42,32 @@ In other words, from your main code, you call specific functions (tasks) in a
...
+    with app.open():
+        # Launch a job
+        # (make sure that the `procrastinate` schema was applied
+        #  _once_ before)
+        sum.defer(a=3, b=5)
 
-    # Somewhere in your program, run a worker
-    app.run_worker(queues=["sums"])
+        # Somewhere in your program (actually, often a different
+        # program than the one defering jobs for execution), run a worker
```
elemoine, line 69:
```suggestion
        # program than the one deferring jobs for execution), run a worker
```

elemoine (approved):
I added another comment, but otherwise looks good to me. Thanks a lot for that.

BracketJohn:
Sure thing 👍

elemoine:
You have this typo elsewhere in your changes. I'll merge this and fix it later. Thanks again for your contribution.

PR #376, opened 2020-12-31
Add async apply schema

### LINKED ISSUE #367: Can't apply the schema in an `async with app` block
As far as I understood, one needs to call `app.schema_manager.apply_schema()` from within the code at least once in order to initialize the database correctly.

In the `sync` variant, this works well. In the `async` variant, it does not work when using the context manager.

This works:
```py
await app.open_async()

app.schema_manager.apply_schema()

await app.close_async()
```

This does not work:
```py
await app.open_async()

await app.schema_manager.apply_schema()

# --> RuntimeError: This event loop is already running

await app.close_async()
```

And neither of the following works:
```py
async with app.open_async():
    app.schema_manager.apply_schema()

# --> RuntimeError: This event loop is already running

async with app.open_async():
    await app.schema_manager.apply_schema()

# --> RuntimeError: This event loop is already running

# ... after some time passes

#  RuntimeWarning: coroutine 'AiopgConnector.execute_query_async' was never awaited
#  self.mouse_handlers[x, y] = handler
#  RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

My first guess would be, that the problem comes from the `schema_manager` being purely sync, see https://github.com/peopledoc/procrastinate/blob/a8a7a50e5097026d924bb86cdf9838cdb3f83245/procrastinate/schema.py#L25.

A workaround that I'm currently using is to instantiate the schema `sync` and then switch to `async`.

BracketJohn:
Closes #367

### Successful PR Checklist:
- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/procrastinate/schema.py
+++ b/procrastinate/schema.py
@@ -23,3 +23,7 @@ class SchemaManager:
     def apply_schema(self) -> None:
         queries = self.get_schema()
         self.connector.execute_query(query=queries)
+
+    async def apply_schema_async(self) -> None:
+        queries = self.get_schema()
+        await self.connector.execute_query_async(query=queries)
--- a/tests/unit/test_schema.py
+++ b/tests/unit/test_schema.py
@@ -1,5 +1,7 @@
 from collections import defaultdict
 
+import pytest
+
 
 def test_get_schema(app):
     assert app.schema_manager.get_schema().startswith("-- Procrastinate Schema")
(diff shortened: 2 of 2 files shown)


elemoine (approved):

mgu (approved):

ewjoachim (approved):
Yay, that a nice one :)

Sorry for the delay, but also, happy new year, and congratulations for a successful PR :) Do you need it released right away ?

BracketJohn:
> Sorry for the delay, but also, happy new year, and congratulations for a successful PR :)

No worries! To you too and thanks - I'm glad I could contribute (:

> Do you need it released right away ?

That would be nice, this and #374 are really nice and will allow us to remove some workarounds (:

ewjoachim:
Here you are :)

https://github.com/peopledoc/procrastinate/releases/tag/0.18.0
https://pypi.org/project/procrastinate/0.18.0/

BracketJohn:
Thanks! (:

PR #379, opened 2021-01-12
add todo check for queueing lock, add test for todo check

### LINKED ISSUE #378: `InMemoryConnector` and `AiopgConnector` _still_ handle `queueing_lock` differently
Follow up to https://github.com/peopledoc/procrastinate/issues/372 - we're only discovering these peu à peu as we're using the library.

### Bug description

For the database connector the following works:

1. enqueue job in queue `A` with queueing lock `B`
2. run worker on queue `A`, so that the job enters state `SUCCEEDED`
3. enqueue job in queue `A` with queueing lock `B`
4. ...

For the `InMemoryConnector`, this does not work:
1. enqueue job in queue `A` with queueing lock `B`
2. run worker on queue `A`, so that the job enters state `SUCCEEDED`
3. enqueue job in queue `A` with queueing lock `B`

-> Raises `AlreadyEnqueued`

### Bug source

The Database queueing lock is implemented using a unique index on `queueing_lock` with the additional check that `status = 'todo'` ([schema here](https://github.com/peopledoc/procrastinate/blob/master/procrastinate/sql/schema.sql#L57))

The Memory queueing lock implementation has a "unique index via `any` check" _without_ the additional check that `status = 'todo'` ([implementation here](https://github.com/peopledoc/procrastinate/blob/master/procrastinate/testing.py#L97))

### Fix

To fix this, the job `status` also needs to be checked, see #379

BracketJohn:
Closes #378

### Successful PR Checklist:
- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/tests/unit/test_testing.py
+++ b/tests/unit/test_testing.py
@@ -2,7 +2,7 @@ import asyncio
 
 import pytest
 
-from procrastinate import utils, exceptions
+from procrastinate import exceptions, utils
 
 from .. import conftest
 
--- a/tests/unit/test_testing.py
+++ b/tests/unit/test_testing.py
@@ -2,7 +2,7 @@ import asyncio
 
 import pytest
 
-from procrastinate import utils
+from procrastinate import utils, exceptions
 
 from .. import conftest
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_testing.py
```diff
@@ -104,6 +104,37 @@ def test_defer_job_one_multiple_times(connector):
...
+        "queue": "default",
+    }
+
+    # 1. Defer job with queueing-lock
+    job_row = connector.defer_job_one(**job_data)
+    assert len(connector.jobs) == 1
+
+    # 2. Defering a second time should fail, as first one
+    #    still in state `todo`
+    with pytest.raises(exceptions.UniqueViolation):
```
BracketJohn, line 125: I'm unsure whether this is the right `raises` condition here, as the db connector directly raises `AlreadyEnqueued`. Any insights on that?

elemoine: The [aiopg connector](https://github.com/peopledoc/procrastinate/blob/80576fc11f89b50743af11ac8d2c6dce91f3a3b0/procrastinate/aiopg_connector.py#L33) and [psycopg2 connector](https://github.com/peopledoc/procrastinate/blob/80576fc11f89b50743af11ac8d2c6dce91f3a3b0/procrastinate/psycopg2_connector.py#L27) actually raise `UniqueViolation`. So the memory connector should do the same.

`UniqueViolation` is caught by the [manager](https://github.com/peopledoc/procrastinate/blob/3d8d8ad67e555853f5939d70abd43029d435194f/procrastinate/manager.py#L41-L42), which, as a result, is responsible for raising `AlreadyEnqueued`.

Hope this answers your question.

BracketJohn: I see - yes, it does, thanks

ewjoachim (approved):
This looks very nice. Thanks for the contribution. @elemoine I'll let you merge if it's ok for you, and  I believe we can issue a hotfix release :)

ewjoachim:
https://pypi.org/project/procrastinate/0.18.1/

PR #354, opened 2020-11-26
Delete finished jobs

mxd4:
Closes #344

This PR adds a new parameter when creating a worker, allowing the deletion of jobs on completion,
either for successful jobs or all jobs.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)
- [x] Had a good time contributing?

PR #383, opened 2021-02-03
Fix finish_job compatibility issue

### LINKED ISSUE #381: Ambiguous PSQL function when `delete_job` undefined
**Summary**

When `delete_job` of the current `procrastinate_finish_job` function is `None`, the current `procrastinate_finish_job`, [defined here](https://github.com/peopledoc/procrastinate/blob/master/procrastinate/sql/schema.sql#L203), becomes ambiguous with the old(est) one [defined here](https://github.com/peopledoc/procrastinate/blob/master/procrastinate/sql/schema.sql#L177).

**Trace**

The resulting bug looks like:
```sh
  File "aiopg/connection.py", line 106, in _ready
    state = self._conn.poll()
psycopg2.errors.AmbiguousFunction: function procrastinate_finish_job(integer, unknown, unknown) is not unique
LINE 1: SELECT procrastinate_finish_job(4, 'succeeded', NULL);
               ^
HINT:  Could not choose a best candidate function. You might need to add explicit type casts.

[...]
 File "/procrastinate/worker.py", line 171, in process_job
    await self.job_manager.finish_job(
 File "/procrastinate/manager.py", line 135, in finish_job
    await self.connector.execute_query_async(
 File "/procrastinate/aiopg_connector.py", line 35, in wrapped
    raise exceptions.ConnectorException from exc
procrastinate.exceptions.ConnectorException: 
    Database error.
```

Sadly, I don't have a minimal example for reproduction at the moment. The error occured when starting one worker after a job was deferred.

A possible solution would be enforcing a boolean value of the `delete_job` param of  `finish_job`.

elemoine:
This fixes an SQL compatibility issue introduced by #354.

I think this should lead to a 0.18.2 bugfix release, and upgrading to 0.18.0 and 0.18.1 should be discouraged.

Closes #381

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/procrastinate/sql/schema.sql
+++ b/procrastinate/sql/schema.sql
@@ -200,7 +200,9 @@ END;
 $$;
 
 -- procrastinate_finish_job
-CREATE FUNCTION procrastinate_finish_job(job_id integer, end_status procrastinate_job_status, delete_job boolean) RETURNS void
+-- the next_scheduled_at argument is kept for compatibility reasons, it is to be
+-- removed after 1.0.0 is released
+CREATE FUNCTION procrastinate_finish_job(job_id integer, end_status procrastinate_job_status, next_scheduled_at timestamp with time zone, delete_job boolean) RETURNS void
     LANGUAGE plpgsql
 AS $$
 DECLARE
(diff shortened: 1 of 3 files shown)


procrastinate/sql/schema.sql
```diff
@@ -200,7 +200,9 @@ END;
 $$;
 
 -- procrastinate_finish_job
-CREATE FUNCTION procrastinate_finish_job(job_id integer, end_status procrastinate_job_status, delete_job boolean) RETURNS void
+-- the next_scheduled_at argument is kept for compatibility reasons, it is to be
+-- removed after 1.0.0 is released
```
ewjoachim, line 204:
```suggestion
-- remove after 1.0.0 is released
```

elemoine: no "removed" is correct here :-)

ewjoachim (approved):

PR #392, opened 2021-03-26
Make it possible to define additional context on the worker

### LINKED ISSUE #390: Access app.connector from a worker
I am not sure this is an appropriate place for questions like mine, as I haven't found a good place for procrastinate related questions - I apologize if it is not & please let me know, so I will move it to an appropriate place.

I want my workers to write results from completed jobs to a psql database - the same database initiated for the App connector. Is there a way for workers to access to app.connector from within a task function?

I have tried my best to search for sample code in the repository and the document and haven't found any. I also tried to pass app or an aiopg.Pool object to my task functions, but I am getting a "TypeError: Object of type Pool is not JSON serializable".

Related to this, what is the best way to pass an object like an aiopg.Pool that is to be shared across jobs and close it before all workers exit? Besides the connection pool, I hope to share aihttp.ClientSession across jobs and then close it before all workers exit. Is it possible?

At this point, I resort to pass the dsn to each job and have each worker create its own psql connection and aihttp.ClientSession. It works but is less than optimal. I like how the arq package (an async job queues with redis) accepts a startup & shutdown function to initiate & dispose resources that are to be shared across jobs (https://arq-docs.helpmanual.io/#startup-shutdown-coroutines). Is this something possible with procrastinate?

Thanks for creating this powerful package!

elemoine:
Closes #390 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/docs/howto/context.rst
+++ b/docs/howto/context.rst
@@ -1,0 +1,40 @@
+Get more context for task execution
+-----------------------------------
+
+When declaring a task, you can request for more context to be passed to the task
+function::
+
+    @app.task(..., pass_context=True)
+    def mytask(context: procrastinate.JobContext, ...):
+        ...
+
+This serves multiple purposes. The first one is for introspection, logs, etc. The
+`JobContext` object contains all sort of useful information about the worker that
+executes the job.
+
+The other useful feature is that you can pass arbitrary context elements using
+`App.run_worker` (or `App.run_worker_async`) and its ``additional_context`` argument. In
+this case the context the task function receives will have an ``additional_context``
+attribute corresponding to the elements that were passed::
+
+    @app.task(pass_context=True)
+    def mytask(context: procrastinate.JobContext):
+        http_session = context.additional_context["http_session"]
+        return await http_session.get("www.example.com")
+
+    async with AsyncSession() as http_session:
+        await app.run_worker_async(additional_context={"http_session": http_session})
+        ...
+
+It may not be a good practice to use this ``additional_context`` object to share data
+from tasks to tasks. In order to keep the least surprising behavior, Procrastinate will
+try to keep modifications of this dictionary in one task from being visible by other
+tasks: tasks receive a shallow copy of this dict instead of the dict itself.
+
+That being said, the values kept in this dict are not processed by Procrastinate. Any
+task mutating a value inside this dict will impact what all the concurrent and following
+tasks will read.
+
+Note that if you start a worker, providing it an ``additional_context`` dict, and then
+modify the dict, the dict the tasks will receive will also be a shallow copy of the dict
+at the time the worker started running.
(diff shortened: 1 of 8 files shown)


elemoine:
Now with docs!

This is ready for review.

docs/howto/context.rst
```diff
@@ -0,0 +1,27 @@
...
+attribute corresponding to the elements that were passed::
+
+    @app.task(pass_context=True)
+    def mytask(context: procrastinate.JobContext):
+        http_session = context.additional_context["http_session"]
+        return await http_session.get("www.example.com")
+
+    async with AsyncSession() as http_session:
+        await app.run_worker_async(additional_context={"http_session": http_session})
+        ...
```
ewjoachim, line 27:
```suggestion
        ...
        
It may not be a good practice to use this ``additional_context`` object to share data from tasks to tasks. In order to keep the least surprising behavior, Procrastinate will try to keep modifications of this dictionary in one task to be visible by other tasks: tasks receive a shallow copy of this dict instead of the dict itself. That being said, the values kept in this dict are not processed by Procrastinate, so any task mutating a value inside this dict will impact what all the concurrent and following tasks will read.

Note that if you start a worker, providing it an ``additional_context`` dict, and then modify the dict, the dict the tasks will receive will be a shallow copy of the dict at the time the worker started running.

If you need a communication channel between tasks and with the rest of your program, you're free to use your own objects for that.
```

docs/spelling_wordlist.txt
```diff
@@ -13,6 +13,7 @@ coroutine
 coroutines
 cron
 crons
+datetime
```
ewjoachim, line 16: this should probably be backticked instead

elemoine: I'm not sure where it is, and I'll search harder :)

ewjoachim: Remove it from the wordlist and re-run the spellcheck, the error will pinpoint the location of the faulty string :)

elemoine: Ok, I found where the problem. It's in the [Job dataclass](https://github.com/peopledoc/procrastinate/blob/97cb7547c2ba1a77f9edf6c18887855c2871e1c6/procrastinate/jobs.py#L75-L76), whose `scheduled` property is automatically documented. I see no other way to fix this than adding `datetime` to the spelling word list.

docs/howto/context.rst
elemoine: Updated with your additional paragraphs.

elemoine:
@ewjoachim this is ready. I left `datetime` in `spelling_wordlist.txt` as I couldn't find any other way to fix the spelling error.

ewjoachim (approved):

PR #399, opened 2021-05-21
Attributes on the decorated task function shouldn't leak on the task

### LINKED ISSUE #394: Task objects created by app.task may not have the expected attributes
In one of our applications we have our own `task` decorator, which looks like this:

```py
def task(f, lock=None, pass_context=False):
    f.lock = lock
    f.pass_context = pass_context
    registry[f.__qualname__] = f
    return f
```

and that we use like this:

```py
@task(lock="mylock", pass_context=False)
def mytaskfunc(…):
    …
```

And we have a setup function that basically does this:

```py
def setup_tasks():
    app = procrastinate.App(…)
    for name, task in registry.items():
        wrapped = task_wrapper(name, task)
        app.task(name=name, pass_context=True)(wrapped)
```

We expect that the `procrastinate.tasks.Task` objects created by `app.task` always have their `pass_context` set to `True`. But that's not the case: if the function object wrapped by `app.task` has `pass_context` set to `False` then the corresponding `procrastinate.tasks.Task` object will have `pass_context` set to `False` as well.

This is because Procrastinate [uses](https://github.com/peopledoc/procrastinate/blob/de781491bf59c317f1b57af02fe6a64dabb05b03/procrastinate/app.py#L187) [`functools.update_wrapper`](https://docs.python.org/3/library/functools.html#functools.update_wrapper) with its default `updated` value, making the wrapper function be updated with all the attributes of the wrapped function.

Using `functools.update_wrapper(task, func, updated=())` should fix the problem.

ewjoachim:
Fixes #394

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/tests/unit/test_app.py
+++ b/tests/unit/test_app.py
@@ -53,7 +53,7 @@ def test_app_task_aliases(app, mocker):
     assert app.tasks["d"] is wrapped
 
 
-def test_app_task_implicit(app, mocker):
+def test_app_task_implicit(app):
     @app.task
     def wrapped():
         return "foo"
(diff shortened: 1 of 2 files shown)


elemoine (approved):
Looks great! And you perfectly understood the problem I reported. Thanks.

tests/unit/test_app.py
```diff
@@ -67,6 +67,19 @@ def wrapped():
...
+    # This is a weird one. It's a regression test. At some point, we noted that,
+    # due to the slightly wrong usage of update_wrapper, the attributes on the
+    # decorated function were copied on the task. This led to surprising
+    # behaviour. This test is just here to make sure it doesn't happen again.
+    def wrapped():
+        return "foo"
+
+    wrapped.pass_context = True
+    task = app.task(wrapped)
+    assert task.pass_context is False
```
ewjoachim, line 80: (let it be noted that I did check and this test didn't pass before the fix)

Issue #400, opened 2021-05-21
Periodic tasks have a NULL lock


ewjoachim:
Periodic jobs for which a lock hasn't been defined on the task have a NULL lock in the database, whereas other jobs without a task are assigned a random UUID as lock.

NULL is not implemented as a special value that keep tasks from being locked, 2 tasks with a lock value NULL will still block one another.

Consequently, it's currently not possible to use periodic tasks and not have each task block subsequent tasks, but worse: all periodic jobs block one another, even across different tasks / queues.

In order to solve this, I suggest we add a UUID default & a NOT NULL constraint (modulo the backwards compatibily DB rules). Also, currently we explicitely create those tasks with the query containing `lock=NULL` so the default value won't be used. We may need triggers.

@elemoine an opinion?

PR #401, opened 2021-05-21
Add list_locks

### LINKED ISSUE #395: Add list_locks task to CLI
The CLI currently has the ability to list the current tasks, jobs, and queues, and allows to filter each of these by the relevant locks, but does not provide a way to list the current (or historical) locks.  Adding the ability to list these would simplify daily admin of projects using Procrastinate

ewjoachim:
Closes #395 
Because of #400 , it's not the cleanest PR ever, but the PR that will handle #400 will fix this.

Also, an ORM would definitely help.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/manager.py
+++ b/procrastinate/manager.py
@@ -422,9 +422,60 @@ class JobManager:
             )
         ]
 
+    async def list_locks_async(
+        self,
+        queue: Optional[str] = None,
+        task: Optional[str] = None,
+        status: Optional[str] = None,
+        lock: Optional[str] = None,
+    ) -> Iterable[Dict[str, Any]]:
+        """
+        List all locks and number of jobs per status for each lock.
+
+        Parameters
+        ----------
+        queue : ``str``
+            Filter by job queue name
+        task : ``str``
+            Filter by job task name
+        status : ``str``
+            Filter by job status (``todo``/``doing``/``succeeded``/``failed``)
+        lock : ``str``
+            Filter by job lock
+
+        Returns
+        -------
+        ``List[Dict[str, Any]]``
+            A list of dictionaries representing locks stats (``name``, ``jobs_count``,
+            ``todo``, ``doing``, ``succeeded``, ``failed``).
+        """
+        result = []
+        for row in await self.connector.execute_query_all_async(
+            query=sql.queries["list_locks"],
+            queue_name=queue,
+            task_name=task,
+            status=status,
+            lock=lock,
+        ):
+            if row["name"] is None:
+                # https://github.com/peopledoc/procrastinate/issues/400
+                continue
+            result.append(
+                {
+                    "name": row["name"],
+                    "jobs_count": row["jobs_count"],
+                    "todo": row["stats"].get("todo", 0),
+                    "doing": row["stats"].get("doing", 0),
+                    "succeeded": row["stats"].get("succeeded", 0),
+                    "failed": row["stats"].get("failed", 0),
+                }
+            )
+        return result
+
 
 utils.add_method_sync_api(cls=JobManager, method_name="list_jobs_async")
 utils.add_method_sync_api(cls=JobManager, method_name="list_queues_async")
 utils.add_method_sync_api(cls=JobManager, method_name="list_tasks_async")
+utils.add_method_sync_api(cls=JobManager, method_name="list_locks_async")
 utils.add_method_sync_api(cls=JobManager, method_name="finish_job_by_id_async")
 utils.add_method_sync_api(cls=JobManager, method_name="retry_job_by_id_async")
--- a/procrastinate/manager.py
+++ b/procrastinate/manager.py
@@ -430,7 +430,7 @@ class JobManager:
         lock: Optional[str] = None,
     ) -> Iterable[Dict[str, Any]]:
         """
-        List all locks and number of jobs per status for each lock.
+        List all locks and number of jobs per lock for each lock value.
 
         Parameters
         ----------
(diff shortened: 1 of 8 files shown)


ewjoachim:
👋 @elemoine

procrastinate/manager.py
```diff
@@ -423,9 +423,60 @@ async def list_tasks_async(
...
 
+    async def list_locks_async(
+        self,
+        queue: Optional[str] = None,
+        task: Optional[str] = None,
+        status: Optional[str] = None,
+        lock: Optional[str] = None,
+    ) -> Iterable[Dict[str, Any]]:
+        """
+        List all locks and number of jobs per status for each lock.
```
elemoine, line 434: Do you mean "for each task" here?

procrastinate/sql/queries.sql
```diff
@@ -149,3 +149,35 @@ SELECT task_name AS name,
...
+          lock,
+          args,
+          status,
+          scheduled_at,
+          attempts
+     FROM procrastinate_jobs
+    WHERE (%(queue_name)s IS NULL OR queue_name = %(queue_name)s)
+      AND (%(task_name)s IS NULL OR task_name = %(task_name)s)
+      AND (%(status)s IS NULL OR status = %(status)s)
+      AND (%(lock)s IS NULL OR lock = %(lock)s)
```
elemoine, line 168: When https://github.com/peopledoc/procrastinate/pull/402 is merged I guess we don't want to have NULL locks in the list returned by list_locks, do we?

procrastinate/sql/queries.sql
```diff
@@ -149,3 +149,35 @@ SELECT task_name AS name,
   FROM jobs AS j
  GROUP BY name
  ORDER BY name;
+
+-- list_locks --
+-- Get list of locks and number of jobs per task
```
ewjoachim, line 154:
```suggestion
-- Get list of locks and number of jobs per lock
```

procrastinate/sql/queries.sql
ewjoachim: I believe it's already the case ? If `%(lock)s IS NULL`, we don't filter on lock.

ewjoachim: Ah you mean
```suggestion
      AND (%(lock)s IS NULL OR lock = %(lock)s) AND lock IS NOT NULL
```

procrastinate/manager.py
```diff
@@ -423,9 +423,60 @@ async def list_tasks_async(
...
+        for row in await self.connector.execute_query_all_async(
+            query=sql.queries["list_locks"],
+            queue_name=queue,
+            task_name=task,
+            status=status,
+            lock=lock,
+        ):
+            if row["name"] is None:
+                # https://github.com/peopledoc/procrastinate/issues/400
+                continue
```
ewjoachim, line 463: These 3 lines should be remove when 402 is merged

procrastinate/sql/queries.sql
elemoine: Yes, exactly.

procrastinate/sql/queries.sql
```diff
@@ -149,3 +149,35 @@ SELECT task_name AS name,
...
+          status,
+          scheduled_at,
+          attempts
+     FROM procrastinate_jobs
+    WHERE (%(queue_name)s IS NULL OR queue_name = %(queue_name)s)
+      AND (%(task_name)s IS NULL OR task_name = %(task_name)s)
+      AND (%(status)s IS NULL OR status = %(status)s)
+      AND (%(lock)s IS NULL OR lock = %(lock)s)
+)
+SELECT lock AS name,
```
elemoine, line 170: Super minor but I'd either remove the alias (`AS name`) or use `lock_name` as the alias (`AS lock_name`).

procrastinate/sql/queries.sql
```diff
@@ -149,3 +149,35 @@ SELECT task_name AS name,
...
+                  COUNT(*) AS jobs_count
+             FROM jobs
+            WHERE lock = j.lock
+            GROUP BY status
+           )
+           SELECT json_object_agg(status, jobs_count) FROM stats
+       ) AS stats
+  FROM jobs AS j
+ GROUP BY name
+ ORDER BY name;
```
elemoine, line 183: Don't you think that the following query is simpler?

```sql
WITH jobs AS (
  SELECT
    id,
    lock,
    status
  FROM procrastinate_jobs
  WHERE (%(queue_name)s IS NULL OR queue_name = %(queue_name)s)
  AND (%(task_name)s IS NULL OR task_name = %(task_name)s)
  AND (%(status)s IS NULL OR status = %(status)s)
  AND (%(lock)s IS NULL OR lock = %(lock)s)
  AND lock IS NOT NULL
), locks AS (
  SELECT
    lock,
    status,
    count(id) AS jobs_count
  FROM jobs
  GROUP BY lock, status
)
SELECT
  lock AS name,
  sum(jobs_count) AS jobs_count,
  json_object_agg(status, jobs_count)
FROM locks
GROUP BY name
ORDER BY name;
```

procrastinate/sql/queries.sql
```diff
@@ -149,3 +149,35 @@ SELECT task_name AS name,
...
+-- Get list of locks and number of jobs per lock
+WITH jobs AS (
+   SELECT id,
+          queue_name,
+          task_name,
+          lock,
+          args,
+          status,
+          scheduled_at,
+          attempts
```
elemoine, line 163: You just need to select the `id`, `lock` and `status` columns here.

elemoine:
This looks great @ewjoachim. I just added a few comments that we may want to take into account.

procrastinate/manager.py
ewjoachim: Nope, it's grouped by lock value, no?

PR #402, opened 2021-05-21
Null locks are ignored

### LINKED ISSUE #400: Periodic tasks have a NULL lock

ewjoachim:
Closes #400 

### Successful PR Checklist:

- [x] Tests < I'd like if we could add some tests, but not sure which
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)
- [ ] If #401 if merged before, we may need to change a thing or two

### DIFF
--- a/procrastinate/sql/migrations/00.19.00_02_null_locks_excluded.sql
+++ b/procrastinate/sql/migrations/00.19.00_02_null_locks_excluded.sql
@@ -1,0 +1,36 @@
+DROP FUNCTION IF EXISTS procrastinate_fetch_job(target_queue_names character varying[]);
+
+CREATE FUNCTION procrastinate_fetch_job(target_queue_names character varying[]) RETURNS procrastinate_jobs
+    LANGUAGE plpgsql
+AS $$
+DECLARE
+	found_jobs procrastinate_jobs;
+BEGIN
+    WITH candidate AS (
+        SELECT jobs.*
+            FROM procrastinate_jobs AS jobs
+            WHERE
+                -- reject the job if its lock has earlier jobs
+                NOT EXISTS (
+                    SELECT 1
+                        FROM procrastinate_jobs AS earlier_jobs
+                        WHERE
+                            jobs.lock IS NOT NULL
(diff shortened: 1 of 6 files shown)


ewjoachim:
@elemoine I'd love your input on this one someday :)

procrastinate/sql/migrations/00.19.00_02_null_locks_excluded.sql
```diff
@@ -0,0 +1,36 @@
+DROP FUNCTION IF EXISTS procrastinate_fetch_job(target_queue_names character varying[]);
```
ewjoachim, line 1: I think this will work with our deprecation policy but another brain validating that would be great :D

ewjoachim:
🤙 @elemoine

procrastinate/sql/migrations/00.19.00_02_null_locks_excluded.sql
elemoine: Yes, it will work. The new SQL function has the same arguments signature as the old SQL function, so it is compatible with the Python code of previous Procrastinate versions.

elemoine:
This looks good to me! Thank you again @ewjoachim.

elemoine (approved):

ewjoachim:
No idea how to test this cleanly though.

ewjoachim:
Brain switched on. Tests added.

Issue #403, opened 2021-05-21
Back to 100% coverage


ewjoachim:
We're missing a line in our coverage. Let's add it back. See [codecov](https://codecov.io/gh/peopledoc/procrastinate) for details.

ewjoachim:
Fixed by #456 !

PR #423, opened 2021-07-26
Add Blueprint and lazy registration pattern

### LINKED ISSUE #421: Declare tasks independently of an App
Currently the API allows up to define tasks with:

`/src/app.py`
```python
import procrastinate

task_app = procrastinate.App(
    connector=procrastinate.AiopgConnector()
    import_paths=[
        "tasks"
    ],
)
```

`/src/tasks.py`
```python
from app import task_app

@task_app.task()
def my_task():
    ...
```

But it might useful to make task definitions independent of the task_app
instance:

`/src/app.py`
```python
import procrastinate

task_app = procrastinate.App(
    connector=procrastinate.AiopgConnector()
    import_paths=[
        "tasks"
    ],
)
```

`/src/tasks.py`
```python
from procrastinate import task

# Then define tasks
@task()
def my_task():
    ...

```

This would allow task definitions to be moved and maintained in seperate
installable packages and configured like:

`/src/app.py`
```python
import procrastinate

task_app = procrastinate.App(
    connector=procrastinate.AiopgConnector()
    import_paths=[
        "other_package.procrastinate_tasks"
    ],
)
```

tomdottom:
Enables collections of tasks to be defined independently of the app
itself and registered at a later time.

Closes #421 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/docs/howto/blueprints.rst
+++ b/docs/howto/blueprints.rst
@@ -1,0 +1,24 @@
+Use Blueprints
+--------------
+
+You can use blueprints to break up your tasks.
+
+Firstly use blueprints to create a task collection::
+
+    from procrastinate import blueprints
+
+    bp = blueprints.BluePrint()
+
+    @bp.task()
+    def mytask(argument, other_argument):
+        ...
+
+Then register blueprint with the app after you have created it::
+
+    from procrastinate import AiopgConnector, App
+
+    app = App(connector=AiopgConnector())
+
+    app.register_blueprint(bp)
+
+Blueprint tasks take the same arguments as `App.task`.
--- a/docs/howto/blueprints.rst
+++ b/docs/howto/blueprints.rst
@@ -1,24 +1,39 @@
-Use Blueprints
---------------
+Create modular collections of tasks by using Blueprints
+-------------------------------------------------------
 
-You can use blueprints to break up your tasks.
+Procrastinate provides Blueprints as a way to factor a large number of tasks
+into smaller self contained collections.
 
-Firstly use blueprints to create a task collection::
+You may want to create a collection for simple organsational reasons within the
+same project.  Or you may want to maintain a collection of tasks in a seperate
+package which is maintained independently of you Procrastinate server and
+workers. eg::
+
+    ...
+
+    from my_external_package import tasks_blueprint
+    ...
+
+    app.register_blueprint(tasks_blueprint)
+
+
+Blueprints are easy to use, and task creation follows the pattern and API as
+`App.task`.
+
+Firstly, create a Blueprint instance and then create some tasks::
 
     from procrastinate import Blueprint
 
-    bp = Blueprint()
+    my_blueprint = Blueprint()
 
-    @bp.task()
+    @my_blueprint.task()
     def mytask(argument, other_argument):
         ...
 
-Then register blueprint with the app after you have created it::
+In your projcet register the blueprint with the `App` after you have created it::
 
     from procrastinate import AiopgConnector, App
 
     app = App(connector=AiopgConnector())
 
-    app.register_blueprint(bp)
-
-Blueprint tasks take the same arguments as `App.task`.
+    app.register_blueprint(my_blueprint)
--- a/docs/howto/blueprints.rst
+++ b/docs/howto/blueprints.rst
@@ -5,9 +5,9 @@ You can use blueprints to break up your tasks.
 
 Firstly use blueprints to create a task collection::
 
-    from procrastinate import blueprints
+    from procrastinate import Blueprint
 
-    bp = blueprints.BluePrint()
+    bp = Blueprint()
 
     @bp.task()
     def mytask(argument, other_argument):
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -21,7 +21,7 @@ class Blueprint(protocols.TaskCreator):
     Notes
     -----
     Deffering a blueprint task before the it is bound to an app will raise an
-    Assertion error::
+    UnboundTaskError::
 
         bp = Blueprint()
 
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -8,6 +8,35 @@ if TYPE_CHECKING:
 
 
 class Blueprint(protocols.TaskCreator):
+    """
+    A Blueprint provides a way to declare tasks that can be registered on an
+    :class:`~procrastinate.App` later::
+
+        bp = Blueprint()
+
+        ... declare tasks ...
+
+        app.register(bp)
+
+    Notes
+    -----
+    Calling a blueprint task before the it is bound to an app will raise an
+    Assertion error::
+
+        bp = Blueprint()
+
+        @bp.task
+        def my_task():
+            ...
+
+        my_task.defer()
+
+        >> AssertionError: Tried to configure task whilst self.app was None
+
+
+
+    """
+
     def __init__(self):
         self.tasks = {}
 
@@ -31,6 +60,59 @@ class Blueprint(protocols.TaskCreator):
         lock: Optional[str] = None,
         queueing_lock: Optional[str] = None,
     ) -> Any:
+        """
+        Like :meth:`App.task <procrastinate.App.task>` except tasks **are not** bound to
+        an app until the blueprint is registered on an app.
+
+        Declare a function as a task as you normally would::
+
+            @bp.task(...)
+            def my_task(args):
+                ...
+
+        Un-configured task decorator will use default values for all parameters::
+
+            @bp.task
+            def my_task(args):
+                ...
+
+
+        Parameters
+        ----------
+        _func :
+            The decorated function
+        queue :
+            The name of the queue in which jobs from this task will be launched, if
+            the queue is not overridden at launch.
+            Default is ``"default"``.
+            When a worker is launched, it can listen to specific queues, or to all
+            queues.
+        lock :
+            Default value for the ``lock`` (see `Task.defer`).
+        queueing_lock:
+            Default value for the ``queueing_lock`` (see `Task.defer`).
+        name :
+            Name of the task, by default the full dotted path to the decorated function.
+            if the function is nested or dynamically defined, it is important to give
+            it a unique name, and to make sure the module that defines this function
+            is listed in the ``import_paths`` of the `procrastinate.App`.
+        aliases:
+            Additional names for the task.
+            The main use case is to gracefully rename tasks by moving the old
+            name to aliases (these tasks can have been scheduled in a distant
+            future, so the aliases can remain for a long time).
+        retry :
+            Details how to auto-retry the task if it fails. Can be:
+
+            - A ``boolean``: will either not retry or retry indefinitely
+            - An ``int``: the number of retries before it gives up
+            - A `procrastinate.RetryStrategy` instance for complex cases
+
+            Default is no retry.
+        pass_context :
+            Passes the task execution context in the task as first
+        """
+
         def _wrap(func: Callable[..., "tasks.Task"]):
             from procrastinate import tasks
 
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -1,0 +1,54 @@
+import functools
+from typing import TYPE_CHECKING, Any, Callable, List, Optional
+
+from procrastinate import jobs, retry
+
+if TYPE_CHECKING:
+    from procrastinate import app, tasks
+
+
+class BluePrint:
+    def __init__(self):
+        self.tasks = {}
+
+    def _register_task(self, task: "tasks.Task") -> None:
+        self.tasks[task.name] = task
+
+    def register(self, app: "app.App") -> None:
+        for task in self.tasks.values():
+            app._register(task)
+
+    def task(
+        self,
+        _func: Optional[Callable] = None,
+        *,
+        name: Optional[str] = None,
+        aliases: Optional[List[str]] = None,
+        retry: retry.RetryValue = False,
+        pass_context: bool = False,
+        queue: str = jobs.DEFAULT_QUEUE,
+        lock: Optional[str] = None,
+        queueing_lock: Optional[str] = None,
+    ) -> Any:
+        def _wrap(func: Callable[..., "tasks.Task"]):
+            from procrastinate import tasks
+
+            task = tasks.Task(
+                func,
+                app=self,
+                queue=queue,
+                lock=lock,
+                queueing_lock=queueing_lock,
+                name=name,
+                aliases=aliases,
+                retry=retry,
+                pass_context=pass_context,
+            )
+            self._register_task(task)
+
+            return functools.update_wrapper(task, func, updated=())
+
+        if _func is None:  # Called as @app.task(...)
+            return _wrap
+
+        return _wrap(_func)  # Called as @app.task
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -16,6 +16,7 @@ class Blueprint(protocols.TaskCreator):
 
     def register(self, app: "app.App") -> None:
         for task in self.tasks.values():
+            task.app = app
             app._register(task)
 
     def task(
@@ -35,7 +36,7 @@ class Blueprint(protocols.TaskCreator):
 
             task = tasks.Task(
                 func,
-                app=self,
+                app=None,  # Causes a mypy error
                 queue=queue,
                 lock=lock,
                 queueing_lock=queueing_lock,
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -36,7 +36,7 @@ class Blueprint(protocols.TaskCreator):
 
             task = tasks.Task(
                 func,
-                app=None,  # Causes a mypy error
+                app=None,
                 queue=queue,
                 lock=lock,
                 queueing_lock=queueing_lock,
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
 class Blueprint(protocols.TaskCreator):
     """
     A Blueprint provides a way to declare tasks that can be registered on an
-    :class:`~procrastinate.App` later::
+    `App` later::
 
         bp = Blueprint()
 
@@ -20,7 +20,7 @@ class Blueprint(protocols.TaskCreator):
 
     Notes
     -----
-    Calling a blueprint task before the it is bound to an app will raise an
+    Deffering a blueprint task before the it is bound to an app will raise an
     Assertion error::
 
         bp = Blueprint()
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -7,7 +7,7 @@ if TYPE_CHECKING:
     from procrastinate import app, tasks
 
 
-class BluePrint:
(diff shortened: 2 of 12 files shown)


tomdottom:
@ewjoachim I wasn't sure how `procrastinate` deals with collisions of task names & task aliases.
This is probably more likely to happen with aliases than with fully qualified task names (fqtn).

I would have thought that adding the same task name or alias _SHOULD_ raise an exception.

We could force a blueprint to have a namespace. Say:

```python
bp = Blueprint(name="my-bp")
```

Or maybe leave this until the registration:

```python
app.register_blueprint(ns="my-ns", blueprint=bp)
```
What do you think?

tomdottom:
Also, do you want to call them `Blueprints` or is there a better name?
Maybe:
- `TaskCollections`
- `Collections`
- `Scaffold`
- `TaskGroup`
- etc ...

ewjoachim:
> Or maybe leave this until the registration:
> 
> ```python
> app.register_blueprint(ns="my-ns", blueprint=bp)
> ```

I like this idea because it makes it the responsibility of the caller (the only one who knows the different blueprints that will be combined) to ensure there is no overlap.

> Also, do you want to call them `Blueprints` or is there a better name?

`Blueprints` seems clear enough to me. :+1:

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
+Use Blueprints
+--------------
+
+You can use blueprints to break up your tasks.
+
+Firstly use blueprints to create a task collection::
+
+    from procrastinate import blueprints
```
ewjoachim, line 8: Anything that is to be accessed from outside of the project is aliased in the `procrastinate` namespace, so we'll need to add it to `procrastinate/__init__.py` and the doc can either:
```
import procrastinate
procrastinate.BluePrint()  # (preferred way)

# or

from procrastinate import BluePrint
BluePrint()
```

procrastinate/blueprints.py
```diff
@@ -0,0 +1,54 @@
...
+        self.tasks = {}
+
+    def _register_task(self, task: "tasks.Task") -> None:
+        self.tasks[task.name] = task
+
+    def register(self, app: "app.App") -> None:
+        for task in self.tasks.values():
+            app._register(task)
+
+    def task(
```
ewjoachim, line 21: I'd be tempted to have something break if the signature for this method deviates from the one in `app.task`. This could be a test that compare signatures from `inspect` (https://docs.python.org/3/library/inspect.html) but it could also be an `abc` (https://docs.python.org/3/library/abc.html).

Though I think my prefered way would be if this could be a protocol (using `runtime_checkable`) https://mypy.readthedocs.io/en/stable/protocols.html#using-isinstance-with-protocols.

I'm not even sure we need to expose this protocol outside of tests.

procrastinate/blueprints.py
```diff
@@ -0,0 +1,54 @@
+import functools
+from typing import TYPE_CHECKING, Any, Callable, List, Optional
+
+from procrastinate import jobs, retry
+
+if TYPE_CHECKING:
+    from procrastinate import app, tasks
+
+
+class BluePrint:
```
ewjoachim, line 10: Hm... or could `App` inherit `Blueprint` ? That would solve the problem.

ewjoachim: A nice advantage of that if it's possible is that `register_blueprint` could be called on an app but also on a blueprint, allowing recursive usage. Or allowing registering tasks from another app.

ewjoachim (approved):
Congratulations on a stellar work so far. Let's get he discussion going but I think we should be able to merge this soon !

tomdottom:
@ewjoachim given that there's no concept of a task namespace in the db schema it will obviously have to be stored by convention in the task name & aliases.

From what I understand a job can be created by it's task name, and a worker can look up a jobs code using the task name or any of its aliases.  So aliases primary use case is to allow task name changes and still allow workers to look up jobs based on old names.

With the :point_up: in mind it seems that only a task name should be prefixed with a namespace, and that it's the users responsibility to include any namespaces in the alias.  Does this seem right?

As for the namespace prefix what type of delimiter would you prefer? :
- `ns.my.task`
- `ns:my.task`
- `ns::my.task`
- `ns/my.task`
- etc ...

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
...
+
+    bp = blueprints.BluePrint()
```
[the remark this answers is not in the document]
tomdottom: Whilst a Blueprint itself is singular, it's very rare to have only one blueprint and they are typically referred to as a group.  So the module name would be plural and the class singular. 

This follows the convention of you `jobs.py` and `tasks.py` .  It's also the convention used in the flask project https://github.com/pallets/flask/blob/main/src/flask/blueprints.py

procrastinate/blueprints.py
tomdottom: I'm not sure I see the problem.

The process involved in `App.register_blueprint` and `Blueprint.register_blueprint` would be different owing to the fact that the former actually binds tasks to the App, but the latter is creating structures to be bound at a later time.

Allowing a blueprint to be registered in a nested fashion would probably be as simple as:

```python
class Blueprint:
    def __init__(self):
        self._blueprints = []
        
    def register_blueprint(self, blueprint: "Blueprint") -> None:
        self._blueprints.append(blueprint)
```

The `App.register_blueprint` would then be responsible for traversing the tree of nested blueprints.

Is this something for this PR, or a separate one?

procrastinate/blueprints.py
tomdottom: I did wonder about how to keep these two signatures inline.  

As far as I understand `abc.abstractmethod` only checks that the method exists, and does not make any assertions about the arguments.

Both inspect and protocols are type assertions.  I guess the question is whether you want any error to occur during `pytest` or `mypy`.

ewjoachim: I'd say both would work. We could have:
```python
class TaskRegisterer(typing.Protocol):
	...

class App(TaskRegisterer):
	...
```
and have it fail on mypy, or we could have
```python
@runtime_checkable
class TaskRegisterer(typing.Protocol):
	...

class App:
	...

# tests
@pytest.mark.parametrize("cls", [App])
def check_task_registering_api(cls):
	assert isinstance(cls, TaskRegisterer)
```

Which one do you like better ? (I think I trust pytest a bit more).

docs/howto/blueprints.rst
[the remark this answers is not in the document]
ewjoachim: Ah I was referring to the capital P in `BluePrint`.

ewjoachim:
As of today, within a worker, if a task is not registered, Procrastinate makes a last chance attempt and tries to load the Python object pointed at that dotted path. If it's a Procrastinate tasks, it logs a warning and launches it.

I believe that if we use `.` as a namespace separator, this will not be possible.

I'd say `:` or `::` are my favorite suggestions from your list. What's your take on that ?

procrastinate/blueprints.py
ewjoachim: > I'm not sure I see the problem.

The problem is ensuring that Blueprint.task() and App.task() signatures stay synchronized, but we've gone another route in the other thread.

Yep, if we don't do the recursive this now, I believe another PR will be fine. Also, we'll probably want to include a namespace here as well :D

tomdottom:
@ewjoachim 

> As of today, within a worker, if a task is not registered, Procrastinate makes a last chance attempt and tries to load the Python object pointed at that dotted path. If it's a Procrastinate tasks, it logs a warning and launches it.

So you're saying that as long as a task package is installed Procrastinate will dynamically load the module and look for the task by name?  How does the `import_paths` argument to the App affect this behaviour?

Furthermore, if there is a python module/package which has side-effects at import these could be triggered by creating a task of the module/package name.  Part of my :brain: is telling me that this is undesirable from a security position. 

:thinking: in a local setup would a task named `antigravity` open the browser?

tomdottom:
> We could allow blueprint providers to include {namespace} in their aliases and compute that at registration time.

Doesn't this defeat the main use case of aliases?  As far as I understand an alias is a prior name of a task, so that you can change the name of a task, and will have workers correct run deferred tasks on the queue with the previous name?

I would have thought any aliases would have to be fully qualified with no manipulation done on them.

For instance if a blueprint added a namespace to aliases it would be difficult to transition a task from:
- a non-namespaced blueprint to a namespaced blueprint
- one namespace to another namespace

This could be worked around if there were someway to opt out of namespacing aliases.  But this just strikes me as adding complexity for the sake of a feature that we don't know we need?

tomdottom:
> I'd say : or :: are my favorite suggestions from your list. What's your take on that ?

I guess the only thing I have against both of them is they look a little like the syntax for slicing in Python.

Also, considering nested blueprints, would you end up with nested namespaces, say`ns1::ns2::my.task`?

Some other thoughts:
- `<ns>my.task`
- `<ns1/ns2>my.task`
- `ns#my.task`
- `ns1#ns2#my.task`

docs/howto/blueprints.rst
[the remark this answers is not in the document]
tomdottom: Ah, that `P` was my fat finger hitting the shift key :laughing: 

All fixed now.

procrastinate/blueprints.py
tomdottom: From https://docs.python.org/3/library/typing.html#typing.runtime_checkable

> Note runtime_checkable() will check only the presence of the required methods, not their type signatures! For example, builtins.complex implements __float__(), therefore it passes an issubclass() check against SupportsFloat. However, the complex.__float__ method exists only to raise a TypeError with a more informative message.

As I read it, this will still not check the args, much as the `abc` does not.

I think using `Protocol` and relying on `mypy` may be the lowest cost check/test that can be implemented.

That said, I'm not that against writing a simple test which uses `inspect` and asserts the two functions have the same signature.  

I'll try and write the test and add the protocol to see if I run into any issues.

ewjoachim:
> thinking in a local setup would a task named antigravity open the browser?

output

I believe it might be interesting to change this :D

tests/unit/test_task_creation_signatures.py
```diff
@@ -0,0 +1,15 @@
...
+def test_task_signatures(app, mocker):
+    """Tasks can be created in two ways, both of which need to maintain an
+    identical API.  This test simple test that App.task and Blueprint.task have
+    the same function signature.
+
+    This is further enforced with protocols.TaskCreator and mypy checks.
+    """
+
+    bp = blueprints.Blueprint()
+    assert inspect.signature(bp.task) == inspect.signature(app.task)
```
ewjoachim, line 15: Wait... If we make a protocol, why don't we use it for the test ?

tomdottom: So I think you're suggesting we use `runtime_checkable` and a `isinstance`.
This would confirm that both `Blueprint` and `App` have a `task` method, but would not confirm anything about the `task` method signature.

As far as I understand, only mypy and `inspect.signature` can do this in depth check.

procrastinate/tasks.py
```diff
@@ -81,7 +81,7 @@ def __init__(
         self,
         func: Callable,
         *,
-        app: app.App,
+        app: Optional[app.App],
```
tomdottom, line 84: @ewjoachim this is needed to avoid `mypy` complaints when a Blueprint creates a task with `app = None`.

I'm not 100% sure if there are any 2nd/3rd order implications from this, hence bringing it to your attention.

All tests pass and mypy has now stopped complaining about it :man_shrugging:

ewjoachim:
Ok
- [ ] it looks like we're missing a bit of coverage (in addition to the missing line from #403)
- [ ] There seems to be a problem with the docs, but let's merge #426 and see what gives.
- [ ] I'll have to do a final review at some point
And then merge !

tomdottom:
@ewjoachim tests pushed which should get coverage back up

I'll merge/rebase off `master` once #426 has been merged in.

tomdottom:
Docs finally building correctly.

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
+Use Blueprints
```
ewjoachim, line 1: Would we use a title that would be more descriptive of the problem we're typing to solve ?
Something like `Define re-usable tasks in a third party application` ?

ewjoachim, line 1: Could we use a title that would be more descriptive of the problem we're typing to solve ?
Something like `Define re-usable tasks in a third party application` ?

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
...
+--------------
+
+You can use blueprints to break up your tasks.
+
+Firstly use blueprints to create a task collection::
```
ewjoachim, line 6:
```suggestion
First, use blueprints to create a task collection::
```

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
...
+
+    from procrastinate import Blueprint
+
+    bp = Blueprint()
```
ewjoachim, line 10:
```suggestion
    blueprint = Blueprint()
```

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
...
+    def mytask(argument, other_argument):
+        ...
+
+Then register blueprint with the app after you have created it::
+
+    from procrastinate import AiopgConnector, App
+
+    app = App(connector=AiopgConnector())
+
+    app.register_blueprint(bp)
```
ewjoachim, line 22:
```suggestion
    app.register_blueprint(blueprint=blueprint)
```

docs/howto/blueprints.rst
```diff
@@ -0,0 +1,24 @@
...
+
+    from procrastinate import Blueprint
+
+    bp = Blueprint()
+
+    @bp.task()
+    def mytask(argument, other_argument):
+        ...
+
+Then register blueprint with the app after you have created it::
```
ewjoachim, line 16:
```suggestion
In your project, register blueprint with the `App` after you have created it::
```

tests/unit/test_task_creation_signatures.py
ewjoachim:
```python
In [5]: @runtime_checkable
   ...: class Drivable(Protocol):
   ...:     def drive(self, from_place: str, to_place: str)->int:
   ...:         ...
   ...:

In [6]: class Car:
   ...:     def drive(self):
   ...:         return 0
   ...:

In [7]: isinstance(Car, Drivable)
Out[7]: True
```

This fills me with sadness but otherwise, you're absolutely right.

ewjoachim: Shouldn't we check that App signature  == protocol signature, and blueprint signature == protocol signature ?
Also I believe we could test that on the classes directly and not on the instances.

procrastinate/blueprints.py
```diff
@@ -0,0 +1,137 @@
...
+from procrastinate import jobs, protocols, retry
+
+if TYPE_CHECKING:
+    from procrastinate import app, tasks
+
+
+class Blueprint(protocols.TaskCreator):
+    """
+    A Blueprint provides a way to declare tasks that can be registered on an
+    :class:`~procrastinate.App` later::
```
ewjoachim, line 13:
```suggestion
    `App` later::
```

procrastinate/blueprints.py
```diff
@@ -0,0 +1,137 @@
...
+
+        bp = Blueprint()
```
ewjoachim, line 15:
```suggestion
        blueprint = Blueprint()
```

procrastinate/blueprints.py
```diff
@@ -0,0 +1,137 @@
...
+
+        ... declare tasks ...
+
+        app.register(bp)
+
+    Notes
+    -----
+    Calling a blueprint task before the it is bound to an app will raise an
```
ewjoachim, line 23:
```suggestion
    Deferring a blueprint task before it is bound to an app will raise an
```

procrastinate/blueprints.py
```diff
@@ -0,0 +1,137 @@
...
+    Assertion error::
```
ewjoachim, line 24: Most (if not all) of the exceptions we get from interacting with Procrastinate inherit our main exception class. I think it's worth creating a new custom exception type (and add it to `docs/reference.py`)

procrastinate/tasks.py
```diff
@@ -167,6 +167,9 @@ def configure(
         ValueError
             If you try to define both schedule_at and schedule_in
         """
+        if self.app is None:
+            raise AssertionError("Tried to configure task whilst self.app was None")
```
ewjoachim, line 171: (let's raise a custom exception here)

ewjoachim (approved):
This starts looking great !
(Ok, I admit I have one thing or 2 I'd like to change but I can do it in a subsequent PR, not to bother you with all my requests :D)

At any point if you'd like me to merge, I'll do it, and I can take care of what's left :)

tests/unit/test_task_creation_signatures.py
tomdottom: > ... test that on the classes directly and not on the instances.

Good call.   Have updated.

tomdottom: > Shouldn't we check that App signature == protocol signature, and blueprint signature == protocol signature

So interesting it appears that MyPy will ignore default values when comparing Protocol method signatrues, but our explicit test using `inspect` requires that defaults be identical.

I've update the Protocol.task signature to include defaults for the moment as this seems to be what we want.  But this might trip us up in the future.

docs/howto/blueprints.rst
tomdottom: The API doesn't require a kwarg, and seeing the word blueprint 3 times in succession just makes me dizzy :dizzy_face:

procrastinate/tasks.py
tomdottom: Created `UnboundTaskError`

docs/howto/blueprints.rst
tomdottom: I took a second stab at this HowTo.  Please re-review.

docs/howto/blueprints.rst
tomdottom: I took a second stab at this HowTo.  Please re-review.

tomdottom:
@ewjoachim 

> This starts looking great !
> (Ok, I admit I have one thing or 2 I'd like to change but I can do it in a subsequent PR, not to bother you with all my requests :D)
> 
> At any point if you'd like me to merge, I'll do it, and I can take care of what's left :)

Addressed many of these remaining comments.  Please re-review and let me know what's left :)

procrastinate/blueprints.py
```diff
@@ -20,7 +20,7 @@ class Blueprint(protocols.TaskCreator):
 
     Notes
     -----
-    Calling a blueprint task before the it is bound to an app will raise an
+    Deffering a blueprint task before the it is bound to an app will raise an
```
ewjoachim, line 23: (Deferring, not deferring, but I'll fix it, don't worry)

ewjoachim:
I'm going to merge now, I'll be glad to continue working this feature with you (e.g. for namespaces) but we can do it in a subsequent PR :) At least, we're starting from somewhere :)

PR #428, opened 2021-07-31
Second pass on blueprints

### LINKED ISSUE #421: Declare tasks independently of an App

ewjoachim:
Closes #421 again.

Ping @tomdottom:
- I might split this into multiple PRs, but the individual commits should make sense
- I'm not sure at all about the approach yet, I'm trying out things and deciding what I like or not. I haven't fixed tests yet, but I believe everything in there should be quite testable.
- And I welcome your opinion too!

So what is this PR doing:
- Removing app.queues, that was unused and undocumented (well except a bit in tests, we'll see)
- Making app inherit blueprint. At this point, all the code App did with tasks was copy-pasted in blueprint. I think it makes sense to say an App really is built from a blueprint :D This removes the need for the protocol and the problems of having matching signatures.
- Added the namespace thing. For now it's `:` the separator. Also, after giving it a though, it names
- As you mentioned, add a safeguard preventing the registration of multiple tasks with the same name. Also, the registration of multiple blueprint namespaces with the same app/blueprint.
- Renaming `register` as `add_tasks_from` because if we have `blueprint.register(blueprint)`, it becomes hard to understand who registers who.
- Made builtin tasks a Blueprint. Drink our own champagne. Our unique builtin task is `builtin:procrastinate.builtin_tasks.remove_old_jobs`
- Added `App.add_task_alias`, which helps when a task changes namespace (e.g. it's moved into a different blueprint). For example, this lets us keep compatibility with our builtin task.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)


tomdottom:
> ... Making app inherit blueprint ...

Removing duplicated code and the `Protocol` to enforce things seems list a step in the right direction.

However, the `add_tasks_from` will have to be slightly different on the `App` and the `Blueprint`; The `App` must bind tasks to the app instance, `Blueprint` will create unbound tasks. 

Currently `Blueprint.register` does the binding:
https://github.com/peopledoc/procrastinate/blob/446a79bb4f52c2ae3b3b4b564f631d8b9adb3f1f/procrastinate/blueprints.py#L46-L49
but there's no reason that this cannot be done in the `App` instead.

ewjoachim:
Right. We could bind the blueprint to the task identically, and on configure, check on the task that `isinstance(self.blueprint, App)` otherwise raise UnboundTaskError, do you think it would make sense ?

tomdottom:
# A thought about task aliases

The current clobbering of previous tasks by a new tasks alias seems wrong to me.  This becomes very apparent with your proposed `add_task_alias` helper.

For example the following shows two scenarios which I think will violate the principle of least surprise:
```python
app.task()
def foo():
    ...

app.task()
def bar():
    ...
       
app.task(aliases=["foo"])
def baz():
    ...

app.task()
def qux():
    ...

app.add_task_alias(qux, "bar")
```

There are two options which I think could be better

## Option 1

If an alias overwrites an existing task (and vice-versa) then throw a `TaskNameConflictError`.

## Option 2

Keep all alias to task mappings in a second object such as `self.aliased_tasks` and only look in this if a task name is not found in `self.tasks`.

That is to say `self.tasks` has a higher priority than `self.aliased_tasks`.

In addition the `App` could emit a warning the there is a name clash between the keys of `self.tasks` and `self.aliased_tasks`.

tomdottom:
# Last but not least

> Drink our own champagne

I've never heard this said before.  It's much classier than

> Eat your own dogfood

:laughing:

ewjoachim:
I think option 1 makes a lot of sense. With aliased_task, we would also need to consider that there could be a clash between the aliases themselves etc. Option 1 will consistently crash at loading time if there's a problem.

ewjoachim:
> ```python
> app.task()
> def foo():
>     ...
>        
> app.task(aliases=["foo"])
> def baz():
>     ...
> ```

I could be wrong, but I believe this would already have raised `TaskAlreadyRegistred` in the code you reviewed. That being said, we'll make sure to have tests for that at some point ^^'

tomdottom:
@ewjoachim I think that a task itself _MUST_ know about it's own name & namespace.

For example I have currently have a few recursive tasks which have to know how to call themselves.  They generally follow the pattern:

```python
@app.task(pass_context=True)
async def recursive_task(ctx: JobContext, next_token: str = None):
    # do something & maybe get a next_token
    if next_token:
        await ctx.app.configure_task(
            name=ctx.task.name
        ).defer(
            next_token=next_token
        )
```

With the addition of namespaces this will break.

If a task knows it's own namespace I will be able to do something like:

```python
@app.task(pass_context=True)
async def recursive_task(ctx: JobContext, next_token: str = None):
    # do something & maybe get a next_token
    if next_token:
        await ctx.app.configure_task(
            name=utils.add_namespace(
                name=ctx.task.name,
                namespace=ctx.task.namespace,
            )
        ).defer(
            next_token=next_token
        )
```

Or better yet:

```python
@app.task(pass_context=True)
async def recursive_task(ctx: JobContext, next_token: str = None):
    # do something & maybe get a next_token
    if next_token:
        await ctx.app.configure_task(
            name=ctx.task.fully_qualified_name
        ).defer(
            next_token=next_token
        )
```

tomdottom:
Also, another thought, is there any value in adding a `root` namespace?
Basically so that:

```python
# task.name -> 'root:my_task'
@app.task
def my_task(): 
    ....
```

ewjoachim:
> I think that a task itself MUST know about it's own name & namespace.

Isn't that already the case ?
https://github.com/peopledoc/procrastinate/pull/428/files#diff-6c59a81916c70409a40864b55e604a20c67558ac466db688a0320f8252d881f6R132

We a task is registered, its name and aliases are modified.
Also, I'd say the easiest way to defer a task is with `task[.configure()].defer()`, so in your examples:

```python
@app.task(pass_context=True)
async def recursive_task(ctx: JobContext, next_token: str = None):
    # do something & maybe get a next_token
    if next_token:
        await recursive_task.defer(
            next_token=next_token
        )
```

ewjoachim:
> Also, another thought, is there any value in adding a root namespace?

I don't like that this would be a special case. For example, if a task is `root:a.b.c` and you `add_tasks_from` with namespace `foo`, this should become `foo:a.b.c`, not `foo:root:a.b.c`.

As of today, a namespace is only added when you use `add_tasks_from`. A root namespace would break this invariant. I'm not sure what it would bring.

tomdottom:
> ... isn't that already the case ?

Ah I'd missed that a tasks name was updated by `add_task_from(...)`.

:thinking: I guess everything should work then.

tomdottom:
> I don't like that this would be a special case.

I guess I was envisioning a slightly different internal architecture were all tasks live on a Blueprint and the top level App has a default blueprint such that:

```python
@app.task
def my_task():
    ....
```

Is really just an alias for 
```python
@app.root_blueprint.task
def my_task():
    ...
```

If you look at Starlette's `Mount` system you might get more of an idea of the arch that I'm thinking of.

However, I'm not sure if it adds that much value, and could create more work than it's worth.

> For example, if a task is `root:a.b.c` and you `add_tasks_from` with namespace `foo`, this should become `foo:a.b.c`, not `foo:root:a.b.c`.
> 

I'm still not convinced you would ever do:

```python
app1 = App(...)

@app1.task
def task1():
    ....

app2 = App(...)
    
app2.add_tasks_from(app1)
```

> As of today, a namespace is only added when you use `add_tasks_from`. A root namespace would break this invariant. I'm not sure what it would bring.

A fair point.

I'm happy to shelve this idea and stick to the pattern as is :)

tomdottom:
@ewjoachim has this stalled ?

ewjoachim:
I'm lacking time to work on this (my "open source" time this month is a little more oriented toward preparing a talk that is due for a conference on August 23rd) and then I'll have PTO. Expect more from me starting August 30th :)

If you want to continue this PR, I won't say no, but otherwise, I do plan to continue it at some point.

tomdottom:
> I'm lacking time to work on this (my "open source" time this month is a little more oriented toward preparing a talk that is due for a conference on August 23rd) and then I'll have PTO. Expect more from me starting August 30th :)
> 
> If you want to continue this PR, I won't say no, but otherwise, I do plan to continue it at some point.

Completely understandable,  I'll try and make some time this weekend and see if I can help in any way.

PR #453, opened 2021-09-24
Add a connector for SQLAlchemy with Psycopg2

### LINKED ISSUE #445: Integration with SQLAlchemy
I'm  looking to integrate Procrastinate with an existing Flask project that uses SQLAlchemy 1.4. While I don't mind creating an additional PostgreSQL connection to defer jobs, it would be better for resource use if there was an option to reuse an existing SQLAlchemy session (or engine).

The advantage is then that the Proscratinate integration would use the same existing connection pool, rather than doubling up the number of connections per thread.

For workers, the existing connection options suffice (although SQLAlchemy 1.4 has [asyncio support options](https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html#using-multiple-asyncio-event-loops) too these days).

elemoine:
This PR adds a connector for SQLAlchemy with Psycopg2. The goal is for synchronous applications that use SQLAlchemy (e.g. a Flask application with the Flask-SQLAlchemy extension) to use just one SQLAlchemy engine, so just one connection pool, thereby minimizing the number of connections opened against the PostgreSQL server.

It works like this:

```py
from slqlalchemy import create_engine

from procrastinate import App
from procrastinate.contrib.sqlalchemy import SQLAlchemyPsycopg2Connector

engine = create_engine("postgresql+psycopg2://")

app = App(connector=SQLAlchemyPsycopg2Connector())
app.open(engine)
```

Closes #445

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -7,3 +7,5 @@ Sphinx==3.*
 sphinx-github-changelog
 sphinx-autodoc-typehints
 sphinxcontrib-programoutput
+sqlalchemy>=1.4,<2.0.0
+django>=2.2
--- a/procrastinate/app.py
+++ b/procrastinate/app.py
@@ -1,6 +1,16 @@
 import functools
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Set
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Union,
+)
 
 from procrastinate import connector as connector_module
 from procrastinate import exceptions, jobs, manager, protocols
@@ -326,8 +336,13 @@ class App(protocols.TaskCreator):
     def schema_manager(self) -> schema.SchemaManager:
         return schema.SchemaManager(connector=self.connector)
 
-    def open(self, pool: Optional[connector_module.Pool] = None) -> "App":
-        self.connector.open(pool)
+    def open(
+        self,
+        pool_or_engine: Optional[
+            Union[connector_module.Pool, connector_module.Engine]
+        ] = None,
+    ) -> "App":
+        self.connector.open(pool_or_engine)
         return self
 
     def close(self) -> None:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,12 @@ python-dateutil = "*"
 typing-extensions = "*"
 importlib-metadata = {version = "*", python = "<3.8"}
 importlib-resources = {version = "*", python = "<3.9"}
-Django = {version = ">=2.2", extras = ["django"]}
+django = {version = ">=2.2", optional = true}
+sqlalchemy = {version = "^1.4", optional = true}
+
+[tool.poetry.extras]
+django = ["django"]
+sqlalchemy = ["sqlalchemy"]
 
 [tool.poetry.dev-dependencies]
 dunamai = "*"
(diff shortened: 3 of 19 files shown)


procrastinate/sqlalchemy_connector.py
```diff
@@ -0,0 +1,176 @@
+import functools
```
ewjoachim, line 1: Should this be in contrib ?

pyproject.toml
```diff
@@ -35,7 +35,12 @@ python-dateutil = "*"
 typing-extensions = "*"
 importlib-metadata = {version = "*", python = "<3.8"}
 importlib-resources = {version = "*", python = "<3.9"}
-Django = {version = ">=2.2", extras = ["django"]}
+django = {version = ">=2.2", optional = true}
+sqlalchemy = {version = "^1.4", optional = true}
+
+[tool.poetry.extras]
+django = ["django"]
+sqlalchemy = ["sqlalchemy"]
```
ewjoachim, line 43: woo, new syntax ?

elemoine: I think the old syntax did not work.

procrastinate/sqlalchemy_connector.py
elemoine: Yes, could be. I have no problem with moving it to the contrib folder. But do we agree that it's just a folder change, and that contribs have nothing else specific?

pyproject.toml
elemoine: Proof is that we didn't have to use `poetry install --extras django` in the CI.

elemoine:
@ewjoachim, we get the following in the codecov report, and I don't know what to do to fix the problem:

```
Current head 17c25ba differs from pull request most recent head 0d0f407. Consider uploading reports for the commit 0d0f407 to get more accurate results
```

Do you know what we can do about it?

elemoine:
This is ready for review again.

ewjoachim:
> The branch rate is `99.72999999999999%`
Woops :D

pyproject.toml
ewjoachim: oh. Damn :/

procrastinate/sqlalchemy_connector.py
elemoine: I moved the connector to the `contrib.sqlalchemy` module.

docs/requirements.txt
```diff
@@ -7,3 +7,5 @@ Sphinx==3.*
 sphinx-github-changelog
 sphinx-autodoc-typehints
 sphinxcontrib-programoutput
+sqlalchemy>=1.4,<2.0.0
+django>=2.2
```
ewjoachim, line 11: Can we rather change `.` at the top of the file to `.[django,sqlalchemy]` ?

procrastinate/app.py
```diff
@@ -1,6 +1,16 @@
...
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Union,
+)
```
ewjoachim, line 13: Wow, I'm not used to having so many imports from a module they don't fit on a single line :D. Well pixel real estate is cheap and if Black wants to spend a bazillion lines, it may well.

ewjoachim (approved):

docs/requirements.txt
ewjoachim: Me, Galaxy brain, makes a comment requesting a change, then merges immediately.

elemoine: Good point. I'll create a PR for that if you haven't done it already.

PR #471, opened 2021-10-21
Arguments for periodic tasks 436

### LINKED ISSUE #436: arguments for periodic tasks
I want to add multiple periodic tasks that have different input arguments, which is currently not possible as far as I can tell.

For example:

```python

def foo(timestamp, a, b):
  return a + b

app.periodic(foo, cron="*/5 * * * *", kwargs={"a": 1, "b": 1))

app.periodic(foo, cron="*/1 * * * *", kwargs={"a": 2, "b": 2))
```
At the moment I use apscheduler to handle cronjobs that need arguments.

Are there plans to add support for this?

aleksandr-shtaub:
Closes #436 

### Notes:
- There is one acceptance test that keeps failing  even on master branch: `tests/acceptance/test_nominal.py::test_nominal`. Maybe, it's somehow connected with my dev environment (WSL Ubuntu).
- It seems like `@attr.dataclass(frozen=True)` doesn't support mutable attributes, so I put FrozenSet instead of dict in PeriodicTask.

### Successful PR Checklist:

- [X] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [X] (not applicable?)


ewjoachim:
Haha and I've broken your tests, perfect :D
I'll fix that at some point, but we need to answer the others remaining questions.

aleksandr-shtaub:
Good point!
I tried to run your example, and it seems like only the first is actually deferred, others are just ignored.

We can just concatenate this optional UID to the task_name. We can also try to use `hash(kwargs)` if `kwargs` is hashable and UID is not provided. In this case we don't need to change table indexes.

As for the second option, it can be acceptable for only a few arguments, otherwise it can turn into a mess like this: 

```python
a.b.my_task(
    arg1=1, 
    arg2={1: {"a": "val1", "b: val2"}, 2: {"a": "val3", "b": "val4"}}, 
    arg3=[<class '__main__.SomeClass'>, <class '__main__.SomeClass'>]
)
```

aleksandr-shtaub:
I've looked at the tests and found that it is failing at `tests/migration/test_migration.py::test_migration`.
The statements with `ALTER TABLE` in the last migration I already fixed. 
But there is another problem with `CREATE FUNCTION procrastinate_defer_periodic_job`: the test keeps failing if we're trying to create function without dropping it first.
I think we should keep `DROP FUNCTION IF EXISTS procrastinate_defer_periodic_job;` above `CREATE` by analogy like in 00.11.00_03, 00.14.00_01 and 00.15.02_01.

ewjoachim:
> I think we should keep `DROP FUNCTION IF EXISTS procrastinate_defer_periodic_job;` above `CREATE` by analogy like in 00.11.00_03, 00.14.00_01 and 00.15.02_01.

Definitely yes !

(As you may have seen, I've been spending some time lately on the PR backlog of procrastinate, I should quickly have more time to work this with you, hopefully)

ewjoachim:
Okay, added a few things:
- Fixed the migration so that they are blue-green.
- Fix (& type) InMemoryConnector & tests: name_suffix is a concept related to periodic tasks only, it should have been added to the tasks themselves. Also, the dictionnary that tracks what periodic jobs have been launched should use (task name, name_suffix) as its index.
- By avoiding lru_cache, PeriodicTask.kwargs can be a normal dict. This makes the code quite easier.
- rename name_suffix as periodic_id, because it's not a suffix anymore
- Added the documentation so that people know who to use the new feature
- Change: Additional periodic args are not tasks kwargs anymore but configure kwargs. This makes it easier to configure not just kwargs, but also locks, queuing locks etc.

ewjoachim:
And sorry it took so long for this PR to get moving !

ewjoachim (approved):

ewjoachim:
And sorry for tagging this good for newcomers, I didn't expect there would be so much work :D

PR #494, opened 2021-11-28
Stop the whole worker process when a coroutine raises

### LINKED ISSUE #493: Exceptions swallowed in the listener and periodic_deferrer
When the database connection drops, an exception gets raised in the listener and periodic_deferrer tasks, but they are logged and not re-raised here: https://github.com/procrastinate-org/procrastinate/blob/master/procrastinate/utils.py#L249

This is an issue because I'd like the workers to crash and restart if these two tasks need reset. Else the periodic tasks won't be running. 

Proposal: re-raise the exception and then wait on these two tasks in the [worker gather](https://github.com/procrastinate-org/procrastinate/blob/master/procrastinate/worker.py#L123-L128) too

ewjoachim:
Closes #493 

@kriscode1 you may want to checkout this PR and test that it solves your problem.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/utils.py
+++ b/procrastinate/utils.py
@@ -1,14 +1,25 @@
 import asyncio
-import contextlib
 import datetime
 import functools
 import importlib
 import inspect
 import logging
 import pathlib
+import sys
 import types
-from typing import Any, Awaitable, Callable, Iterable, Optional, Type, TypeVar
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Coroutine,
+    Iterable,
+    List,
+    Optional,
+    Type,
+    TypeVar,
+)
 
+import attr
 import dateutil.parser
 
 from procrastinate import exceptions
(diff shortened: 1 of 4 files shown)


procrastinate/utils.py
```diff
@@ -307,3 +287,173 @@ async def _inner_coro() -> U:
...
+    pass
+
+
+@attr.dataclass()
+class ExceptionRecord:
+    task: asyncio.Task
+    exc: Exception
+
+
+async def run_tasks(
```
ewjoachim, line 302: All in all this function is 150 lines long, but there's a lot of doc/comments. Also, I'm not sure how to split it more efficiently. The base logic would be shorter but it order to log in all the strategic places, and because asyncio.Task supports task names only on 3.8+, we end up with more complex code.

kriscode1:
@ewjoachim thank you for the speedy fix! I tested this branch in my environment and it solves the issue.

ewjoachim:
Thanks for your feedback :) I'll try to have this merged and released in the near future.

Issue #320, opened 2020-09-23
Add Django models


ewjoachim:
While I'm not sure it's a great idea to let people write things in the DB using Django, there's no harm in exposing our 2 tables as django models, so that people can easily add custom procrastinate monitoring to their Django projects.

So one model per table, and disable the save & delete methods (for now) as well as the manager update & delete. That should do the job, as a first approach.

ewjoachim:
Done by #906

PR #506, opened 2021-12-21
Draft

### LINKED ISSUE #502: Django - support for transaction.atomic
Hey guys, thanks for a great library.

### Context

Just to give you a context, I'd like to use it in Django project to reliably communicate with other microservices and implement [Transactional outbox](https://microservices.io/patterns/data/transactional-outbox.html) and [Message Relay](https://microservices.io/patterns/data/polling-publisher.html) patterns. 

### Problem
The problem is that currently it's not possible to save django models and defer a task in one transaction. Example code below:

```python
import procrastinate
from django.db import transaction
from .models import *
from procrastinate.contrib.django import connector_params
from asgiref.sync import sync_to_async

app = procrastinate.App(
    connector=procrastinate.Psycopg2Connector(**connector_params())
)

@app.task
def some_other_task():
     print("Hello Word")

@app.task
@sync_to_async  # Needed to be able to use Django ORM
@transaction.atomic
def send_order_placed_notification():
    some_other_task.defer()  # Task is deferred even when the next line throws IntegrityError (transaction rollback)
    SomeModel.objects.create()

@transaction.atomic
def some_view(request):
    send_order_placed_notification.defer()  # Task is deferred even when the next line throws IntegrityError (transaction rollback)
    Order.objects.create()
```

### Root cause
Psycopg2Connector/AiopgConnector create separate database connection instead of using `django.db.connection`

michaszcz:
TODO

Closes #502 

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)


procrastinate/contrib/django/django_connector.py
```diff
@@ -0,0 +1,207 @@
...
+        return self.execute_query_one(query, **arguments)
+
+    @wrap_exceptions
+    def execute_query_one(self, query: str, **arguments: Any) -> dict[str, Any]:
+        with self._connection() as connection:
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    PERCENT_PATTERN.sub("%%", query), self._wrap_json(arguments)
+                )
+                return self._dict_fetchone(cursor)
```
michaszcz, line 147: Django cursor returns _str_ instead of _dict_ for _args_ field (jsonb). What is the cleanest solution for this?

procrastinate/contrib/django/django_connector.py
```diff
@@ -0,0 +1,207 @@
...
+        connection: ConnectionProxy,
+        timeout: float = LISTEN_TIMEOUT,
+    ) -> None:
+        while True:
+            pg_conn = await self._get_pg_conn(connection)
+            if pg_conn.closed:
+                return
+
+            try:
+                await self._wait_for_file_descriptor(pg_conn.fileno(), timeout)
```
michaszcz, line 201: Is there a better solution then this?

procrastinate/contrib/django/django_connector.py
```diff
@@ -0,0 +1,207 @@
...
+            for key, value in arguments.items()
+        }
+
+    @contextlib.contextmanager
+    def _connection(self) -> Iterator[ConnectionProxy]:
+        yield django_connection
+
+    def _dict_fetchone(self, cursor) -> dict[str, Any]:
+        """Return row from a cursor as a dict"""
+        columns = [col[0] for col in cursor.description]
```
ewjoachim, line 96:
```suggestion
        columns = [col.name for col in cursor.description]
```
(from https://www.psycopg.org/docs/cursor.html#cursor.description)

procrastinate/contrib/django/django_connector.py
```diff
@@ -0,0 +1,207 @@
...
+        r = dict(zip(columns, cursor.fetchone()))
+        try:  # TODO fix workaround
+            r["args"] = json.loads(r["args"])
```
ewjoachim, line 99: instead of doing this based on the fact the column is named "args", we could base this on the column OID (`type_code`), but I'm a little surprised that Django hasn't registered the jsonb extension already.

procrastinate/contrib/django/django_connector.py
ewjoachim: I think that up until now, we've considered that the listen-notify took a dedicated connection. If we're trying to shoehorn everything into a single connection, I'm afraid there might be consequences. What about, as a first step, disabling listen-notify when using a Django connector ?

procrastinate/contrib/django/django_connector.py
michaszcz: It's done [here](https://github.com/django/django/blob/0153a63a674937e4a56d9d5e4ca2d629b011fbde/django/db/backends/postgresql/base.py#L206) in Django

ewjoachim:
I think we need to decide of an approach:
- either we try to make the Django Connector as close as possible to the Psycopg2 connector, and undo all the things Django does and I think we're going to fight against django all the way through
- Or we try to "embrace" the Django way: use the ORM and `Model.objects.raw()` (and do #320 at the same time). Cons: there might be more work, and I have no idea whether this will work. Pro: less likely to completely break with each new Django version.

What's your opinion here ?

procrastinate/contrib/django/django_connector.py
michaszcz: Actually I need listen-notify in workers in my project. Disabling should be possible with [this flag](https://procrastinate.readthedocs.io/en/stable/reference.html#procrastinate.App.run_worker).

Also I saw similar approach of using listen-notify in Django [here](https://github.com/gavinwahl/django-postgres-queue/blob/24a5d0297a1b7b8d57d2facb249c0600bed7a95b/dpq/queue.py#L65).

michaszcz:
Another option is to use django cursors to execute [custom sql directly](https://docs.djangoproject.com/en/3.2/topics/db/sql/#executing-custom-sql-directly) (like in this PR at the moment). Pro: no additional work to create models.
I think the best to use Django features as much as possible - so 2nd and 3rd approach

procrastinate/contrib/django/django_connector.py
michaszcz: I think creating separate aiopg connection just for listen-notify is also an option here.

procrastinate/contrib/django/django_connector.py
michaszcz: I see it's related to this [django 3.1.1 bugfixes](https://docs.djangoproject.com/en/3.1/releases/3.1.1/#bugfixes).

> Fixed a QuerySet.order_by() crash on PostgreSQL when ordering and grouping by JSONField with a custom decoder (#31956). As a consequence, fetching a JSONField with raw SQL now returns a string instead of pre-loaded data. You will need to explicitly call json.loads() in such cases.

ewjoachim:
I don't see the difference between your 3rd approach and my 1st ?

michaszcz:
Maybe I don't fully understand what you meant in 1st approach. Psycopg2 can't be used by workers. Aiopg can be used with workers, but it is not using django connection, so task deferring won't be possible inside transaction. So in 3rd approach I meant DjangoConnector would works like mix of Psycopg2 and Aiopg. Also Django connection/cursor objects would be used everywhere instead of postgres equivalents extracted from Django connection (excluding listen-notify case).

ewjoachim:
Ah ok, I understand your point. I had forgotten that the Psycopg2Connector was incompatible with workers.

But then it seems it would be easier to make the Psycopg2Connector compatible with workers, and then we'd be almost all good to have it work with Django ?

michaszcz:
I think it still wouldn't work, because Psycopg2Connector creates separate connections and django.db.connection must be used to make transactions work correctly.

procrastinate/contrib/django/django_connector.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @wrap_exceptions
+    def execute_query_one(self, query: str, **arguments: Any) -> Dict[str, Any]:
+        query = self._prepare_for_interpolation(
+            query=query, has_arguments=bool(arguments)
+        )
+        with django_connection.cursor() as cursor:
+            cursor.execute(query, self._wrap_json(arguments))
+            r = self._dict_fetchone(cursor)
+        django_connection.close()  # TODO try to remove this line
```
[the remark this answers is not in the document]
michaszcz: @ewjoachim do you have some suggestion how this could be fixed? (it should fix failing tests)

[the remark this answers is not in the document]
michaszcz: The connection is not closed, because sync_to_async function causes connection to be created in a different thread.

wgordon17:
Hey all 👋 I'm curious as to the current state of Procrastinate's support for Django integration. I've found this in the documentation, https://procrastinate.readthedocs.io/en/stable/howto/django.html, and I'm wondering how much functionality there is in this Draft PR that would be beneficial.

ewjoachim:
Hey :)

I feel a bit ashamed that I let this PR rot :/ I felt that I needed to discuss it with other maintainers and it never really happenned :(

My issue is that the "connector" situation is already a bit of a mess especially regarding sync vs async, and I'm afraid that making the wrong choice in integrating a django connector might make it even harder when we'll address the whole subject for real.

I'm not saying this PR makes the wrong choice but I'm actually a bit lost regarding what the correct path would be :/

wgordon17:
Hey @ewjoachim, I totally understand! After looking around, I really enjoy the API provided by Procrastinate, especially compared to a _lot_ of the other task/queue frameworks out there.

It would be great to have a tighter Django integration, but I'm going to proceed with the project as-is and see how it turns out ¯\_(ツ)_/¯ 

With that said, I certainly agree with your earlier sentiments (https://github.com/procrastinate-org/procrastinate/pull/506#issuecomment-998698519) on "embrace the Django way". I'll be honest, I don't have a lot of experience with the low-level Django ORM "magic" nor do I have _any_ experience with Procrastinate. But I'd still love to see this deeper integration happen, and I might be able to help out in smaller ways if you need any help on getting this through 😊 Docs, tests, etc.

mecampbellsoup:
> My issue is that the "connector" situation is already a bit of a mess especially regarding sync vs async, and I'm afraid that making the wrong choice in integrating a django connector might make it even harder when we'll address the whole subject for real.

Is this perhaps why we have a hard time using a sync app (with our django-allauth backend, synchronous Python context) along with our otherwise async ASGI FastApi application? When I try to use `@sync_app.task` decorator on our Django code/functions, I tend to see errors like this:

```python
@async_app.task(**request_status_io_task_args)
async def add_subscriber(email: str) -> None:
    """Add user as a subscriber to status.io maintenance notifications"""
```

```python
# Django model code
@receiver(user_signed_up)
def enroll_status_io(request, user: User, **kwargs) -> None:
    add_subscriber.defer(email=user.email)
```

```python
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/allauth/account/utils.py", line 183, in complete_signup
[cloud-app]     signals.user_signed_up.send(
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 180, in send
[cloud-app]     return [
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 181, in <listcomp>
[cloud-app]     (receiver, receiver(signal=self, sender=sender, **named))
[cloud-app]   File "/cloud-app/iam/models.py", line 256, in enroll_status_io
[cloud-app]     add_subscriber.defer(email=user.email)
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/procrastinate/tasks.py", line 129, in defer
[cloud-app]     return self.configure().defer(**task_kwargs)
[cloud-app]   File "/cloud-app/tests/conftest.py", line 728, in defer_from_jobdeferrer
[cloud-app]     job_id = original_defer_from_jobdeferrer(self, **task_kwargs)
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/procrastinate/jobs.py", line 163, in defer
[cloud-app]     job = self.job_manager.defer_job(job=job)
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/procrastinate/manager.py", line 52, in defer_job
[cloud-app]     result = self.connector.execute_query_one(
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/procrastinate/utils.py", line 148, in wrapper
[cloud-app]     return sync_await(awaitable=awaitable)
[cloud-app]   File "/usr/local/lib/python3.10/site-packages/procrastinate/utils.py", line 195, in sync_await
[cloud-app]     loop = asyncio.get_event_loop()
[cloud-app]   File "/usr/local/lib/python3.10/asyncio/events.py", line 656, in get_event_loop
[cloud-app]     raise RuntimeError('There is no current event loop in thread %r.'
[cloud-app] RuntimeError: There is no current event loop in thread 'WSGI_0'.
```

Our workaround historically has been to make an ad-hoc, one-time-use sync app as follows:

```python
@receiver(user_signed_up)
def enroll_status_io(request, user: User, **kwargs) -> None:
    defer_with_sync_app(add_subscriber, **request_status_io_task_args)(email=user.email)
```

```python
def defer_with_sync_app(task_func, **task_kwargs) -> Callable:
    def wrapper(**kwargs):
        app = make_sync_app()
        task = app.task(**task_kwargs)(task_func)
        with app.open():
            task_id = task.defer(**kwargs)
            return task_id

    return wrapper
```

ewjoachim:
Yes, that's probably the reason :/

adriangb:
> I'd say a potentially good case study candidate would be asgiref's sync_to_async and async_to_sync. It's possible that we might be able to use and/or vendor those and get something that works nicely enough for both.

Be cautious. You'll start to run into weird stuff with contextvars (they don't really work well to/from threads) and semaphores on threads (for example FastAPI can only run 40 sync dependencies/endpoints concurrently because of an implicit semaphore buried deep in the stack).

adriangb:
Just rubber ducking here.
What would you do if you were starting all over? Only support one? Support both but in completely separate codebases?

ewjoachim:
I guess we would need to make as much as possible of the procrastinate code independant from I/Os (fortunately, this is already what happens in many parts of the codebase, but not on the workers, though) and have high level functions that only take care of I/Os, and delegate all the "business logic" to our ["functional" core](https://www.destroyallsoftware.com/talks/boundaries).

It's probably not as easy as it sounds, because it's not just a question of using dependency injection to be loosely coupled from whoever does the I/Os, as we need to avoid having to choose whether we call `x()` or `await x()` in our function.

So... Maybe implement each "logical" entrypoint as a generator that would be sending "I/O requests" and receiving "I/O responses" and... I'm re-inventing async I/O all over again, aren't I? 😅

mecampbellsoup:
FTR I have the same threading problem when trying to use either our async app or a sync one.

adriangb:
For cases where you _know_ the async thing is not using tasks internally (maybe because you wrote it) you can do something like this: https://github.com/edgedb/edgedb-python/blob/master/edgedb/blocking_client.py#L343-L353

ewjoachim:
Worker heavily uses tasks, sadly... but we'll need a full rewrite anyway...

ewjoachim:
Superseeded by #906 .
Thank you to everyone who contributed though!

PR #669, opened 2022-08-16
Asyncpg Connector implementation

### LINKED ISSUE #417: Support for asyncpg
Would it be feasible to add support for [asyncpg](https://github.com/MagicStack/asyncpg)?  (Maybe even replace `aiopg`)

procrastinate is written specifically for running on postgres, and so is [asyncpg](https://github.com/MagicStack/asyncpg), which is why it is extremely performant. `asyncpg` is [way faster](https://magic.io/blog/asyncpg-1m-rows-from-postgres-to-python/) than `aiopg`, and it is even faster than golangs `pgx`.

Furthermore, it is supported by  [encode/databases](https://github.com/encode/databases) which plays very well with FastAPI,  SQLAlchemy Core and [ormar](https://github.com/collerek/ormar/).

nwillems:
Still very very early implementation. Mostly taking the suggested converter and loosely implementing the connector interface.

Needs:
- [ ] Error handling - steal from others?
- [ ] Tests - copy/paste from aiopg
- [ ] a second set of eyes, maybe even more

Closes #417 

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/asyncpg_connector.py
+++ b/procrastinate/asyncpg_connector.py
@@ -1,0 +1,170 @@
+from typing import Any, Callable, Coroutine, Dict, Iterable, List, Optional
+
+import asyncpg
+
+from procrastinate import connector, exceptions, sql
+
+
+def wrap_exceptions(coro: CoroutineFunction) -> CoroutineFunction:
+    """
+    Wrap asyncpg errors as connector exceptions.
+
+    This decorator is expected to be used on coroutine functions only.
+    """
+
+    @functools.wraps(coro)
+    async def wrapped(*args, **kwargs):
+        try:
+            return await coro(*args, **kwargs)
+        except asyncpg.errors.UniqueViolation as exc:
+            raise exceptions.UniqueViolation(constraint_name=exc.diag.constraint_name)
+        except asyncpg.Error as exc:
+            raise exceptions.ConnectorException from exc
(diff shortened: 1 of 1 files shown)


ewjoachim:
Hey there, sorry for the delay.

I haven't been very active on the project maintenance these days. It looks like you're going in the right direction. The structure of the project means that adding a new db driver should, as you saw, only result in a single new file (plus tests).

Regarding test, it would be worth writing specific tests for the connector. Also, a sencod set of brain to ensure there's no risk for SQL injection is never too much.

I wonder if it would make sense that the converter has a method that receives query and arguments and returns the new query and the new argument instead of having to do the replacement yourself outside of the converter with `%`.

nwillems:
I'm no longer working with python project on a daily basis, and hence don't have the time to finish this work. A lot has happened, sorry for the long pause. I hope someone will pick this up.

Thanks for a great project.

ewjoachim:
Hey, thanks for the time you spent anyway :)

I'm currently working towards a psycopg3 implementation, and I'm starting to believe it will probably be the only supported one in the future. If people want to support other connectors, They'll likely be able to do it on their own, potentially in a dedicated package.

PR #682, opened 2022-09-23
friendly error for get_full_path failures

### LINKED ISSUE #674: functools.partial not compatible with App.task decorator
## Issue

Functions that have been curried with python's built-in [functools.partial](https://docs.python.org/3.10/library/functools.html?highlight=functools#functools.partial) will crash inside the `App.task` decorator, because partial instances don't have a `__name__`.

Easy to work around (see below), so not urgent.

## Workaround

Setting `name` in the task config seems to solve this.

```diff
- my_task = pro.task()(functools.partial(decorated_fn, arg))
+ my_task = pro.task(name='module.task')(functools.partial(decorated_fn, arg))
```

## Possible solution

The error is happening in [get_full_path](https://github.com/procrastinate-org/procrastinate/blob/1a54ab3fedb8a421c93dc63b99aa567602801b86/procrastinate/utils.py#L237).

In theory you could detect a `partial` in there + use `obj.func.__name__` in place of `obj.__name__`. This risks `TaskAlreadyRegistered` if the same function is partialed twice in a module, but at least solves the 'first partial' case.

Sub-ideal options:
- serial suffix `f"{module}.{obj.func.__name__}-partial-{serial_index}"`. I think this is robust against import order changing, but could be broken in edge cases or if tasks are defined by functions which are called dynamically. Would also require keeping a counter somewhere
- use hash of partial arguments as suffix, but this isn't consistent across runs. (in my use case, where the partial argument is a class, hashes aren't consistent across runs)

I'm not sure how automatic task import works in the procrastinate worker, so I'm not sure if one of these breaks auto-import.

Happy to be assigned this PR if helpful + small in scope.

## Traceback

```python
File "backend/util.py", line 46, in <module>
  send_push = pro.task(queue='push')(functools.partial(push.send_push, PushToken))
File ".direnv/python-3.10.4/lib/python3.10/site-packages/procrastinate/blueprints.py", line 241, in _wrap
  task = tasks.Task(
File ".direnv/python-3.10.4/lib/python3.10/site-packages/procrastinate/tasks.py", line 93, in __init__
  self.name: str = name if name else self.full_path
File ".direnv/python-3.10.4/lib/python3.10/site-packages/procrastinate/tasks.py", line 113, in full_path
  return utils.get_full_path(self.func)
File ".direnv/python-3.10.4/lib/python3.10/site-packages/procrastinate/utils.py", line 239, in get_full_path
  return f"{_get_module_name(obj)}.{obj.__name__}"
AttributeError: 'functools.partial' object has no attribute '__name__'
```

abe-winter:
Closes #674

## Changes
- `utils.get_full_path` throws error on failure, with instructions for resolving
- new exception `FunctionPathError`
- added test

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

### DIFF
--- a/procrastinate/utils.py
+++ b/procrastinate/utils.py
@@ -235,8 +235,12 @@ def _get_module_name(obj: Any) -> str:
 
 
 def get_full_path(obj: Any) -> str:
-
-    return f"{_get_module_name(obj)}.{obj.__name__}"
+    try:
+        return f"{_get_module_name(obj)}.{obj.__name__}"
+    except AttributeError as exc:
+        raise exceptions.FunctionPathError(
+            f"Couldn't automatically name {type(obj)}. Try passing `name=` to the task decorator"
+        ) from exc
 
 
 def utcnow() -> datetime.datetime:
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -120,3 +120,7 @@ class InvalidTimestamp(ProcrastinateException):
     Periodic task launched with a timestamp kwarg not matching the
     defer_timestamp arg
     """
+
+
+class FunctionPathError(ProcrastinateException):
+    """Couldn't automatically generate a unique name for a function"""
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import functools
 import sys
 import time
 import types
@@ -229,6 +230,11 @@ def test_get_full_path():
     path = "procrastinate.utils.get_full_path"
     # Calling the function on itself
     assert utils.get_full_path(utils.get_full_path) == path
+
+
+def test_get_full_path_partial():
+    with pytest.raises(exceptions.FunctionPathError):
+        utils.get_full_path(functools.partial(test_get_full_path_partial))
 
 
 # `launched` and `finished` are sets used by the run_tasks tests, both coro and


procrastinate/utils.py
```diff
@@ -235,8 +235,12 @@ def _get_module_name(obj: Any) -> str:
 
 
 def get_full_path(obj: Any) -> str:
-
-    return f"{_get_module_name(obj)}.{obj.__name__}"
+    try:
+        return f"{_get_module_name(obj)}.{obj.__name__}"
+    except AttributeError as exc:
+        raise exceptions.FunctionPathError(
+            f"Couldn't automatically name {type(obj)}. Try passing `name=` to the task decorator"
```
ewjoachim, line 242: Maybe `Couldn't extract a relevant task name for {type(obj)}...`

Also, I wonder if printing `obj` itself would be better ?

procrastinate/utils.py
```diff
@@ -235,8 +235,12 @@ def _get_module_name(obj: Any) -> str:
 
 
 def get_full_path(obj: Any) -> str:
-
-    return f"{_get_module_name(obj)}.{obj.__name__}"
+    try:
+        return f"{_get_module_name(obj)}.{obj.__name__}"
+    except AttributeError as exc:
```
ewjoachim, line 240: Hm, I wonder if it's very obvious that _get_module_name() may raise an AttributeError.

- Either that should be a custom ProcrastinateException but that might (or might not) be overkill
- or we could add a line in that function's docstring saying it may raise an AttributeError and in what case.

ewjoachim (commented):
It's almost perfect :)

procrastinate/utils.py
abe-winter: the AttributeError isn't coming from _get_module, it's coming from `obj.__name__` in this function

(I may be misunderstanding your question here)

procrastinate/utils.py
abe-winter: changed

procrastinate/utils.py
ewjoachim: Ah damn you're right, I read too fast. Then, can we change the code so that we do `obj.__name__` alone it its try/except clause and only then we return the f-string ?

abe-winter: yup! fixed

ewjoachim (approved):
Perfect :)

abe-winter:
Thanks for the review -- also great work with the tooling here. The combination of the pre-commit hook, high test coverage and separate unit testing suite made working in this codebase great

ewjoachim:
Aw thanks :)

(It's released btw)

PR #695, opened 2022-10-20
switch from poetry to hatch for packaging

singingwolfboy:
I'm not sure that you have any interest in accepting this pull request, but I thought I would send it just in case. I prefer [Hatch](https://hatch.pypa.io/latest/) over [Poetry](https://python-poetry.org), in part because Hatch follows [modern Python packaging standards](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/) while Poetry does its own thing.

I'm interested in exploring if I can make this project work with `asyncpg` (taking the work in #669 and continuing it), but [it sounds like this project is not very active at the moment](https://github.com/procrastinate-org/procrastinate/pull/669#issuecomment-1222350330), so I'll probably end up working on my fork of the project. As a result, the first thing I want to do is switch to Hatch, hence this pull request. If you accept this PR, I'm happy to send more in the future! If not, I'll probably just focus on my fork without sending further PRs.


ewjoachim:
I'm not interested for now. I quite like poetry. Also, they're also slowly evolving towards [PEP 621](https://github.com/python-poetry/roadmap/issues/3). I hope you didn't miss the fact that for all "its own thing" poetry does, its creator is listed as a co-author of the said PEP :)

It also makes a lot of sense to me to try and steer all my projects to use the same set of technologies at any given time, to limit cognitive overload, so if I start switching to Hatch, I'd try to do it for my other projects too over time, and it not in my priorities for now.

Plus, I don't understand why your PR removes a lot of things not linked to Poetry (the dev env, all the things you removed from the Dockerfile, you're also removing the part that synchronize pre-commit versions with the versions of the project, etc.)

singingwolfboy:
> I'm not interested for now. I quite like poetry.

Totally fair. We're allowed to have different preferences!

> Also, they're also slowly evolving towards [PEP 621](https://github.com/python-poetry/roadmap/issues/3). I hope you didn't miss the fact that for all "its own thing" poetry does, its creator is listed as a co-author of the said PEP :)

Nifty!

> It also makes a lot of sense to me to try and steer all my projects to use the same set of technologies at any given time, to limit cognitive overload, so if I start switching to Hatch, I'd try to do it for my other projects too over time, and it not in my priorities for now.

Also very fair. You should use the tools that you want to use.
 
> Plus, I don't understand why your PR removes a lot of things not linked to Poetry (the dev env, all the things you removed from the Dockerfile, you're also removing the part that synchronize pre-commit versions with the versions of the project, etc.)

I tried to simplify the dev environment situation, and had to rewrite anything that depended on Poetry. As for `pre-commit`, that action is named "sync_with_poetry", so I didn't think it was relevant after Poetry was removed. 😄

As I said in the description, I didn't think you would be interested in accepting this pull request, and that's OK. I'll work on my own fork, and if it goes anywhere useful, maybe we can re-sync in the future!

PR #754, opened 2023-03-05
Update middleware docs

adibsaad:
The code in the middleware docs wasn't working for me when I tried to call .defer_async() on the wrapped function, I was getting this error

```
    await noop.defer_async()
AttributeError: 'function' object has no attribute 'defer_async'
```

This only happens when I use the decorator like `@task` , but when I use it like `@task()`, it works.

I was also getting this error

```
    f"A task named {name} was already registered"
procrastinate.exceptions.TaskAlreadyRegistered: A task named src.tasks.app.new_func was already registered
```

Likely because I have multiple tasks using the `@task` decorator, so it was registering the inner `new_func` function instead of the original one.

The new code I submitted fixes the issue.

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/docs/howto/middleware.rst
+++ b/docs/howto/middleware.rst
@@ -7,6 +7,8 @@ your own decorator instead of ``@app.task`` and have this decorator
 implement the actions you need and delegate the rest to ``@app.task``.
 It might look like this::
 
+    import functools
+
     def task(*args, **kwargs):
         def wrap(func):
             def new_func(*job_args, **job_kwargs):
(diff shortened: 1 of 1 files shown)


docs/howto/middleware.rst
```diff
@@ -16,7 +18,12 @@ It might look like this::
                 log_something_else()
                 return result
 
-            return app.task(*args, **kwargs)(new_func)
+            wrapped_func = functools.update_wrapper(new_func, func, updated=())
+            return app.task(*args, **kwargs)(wrapped_func)
+
+    if not original_func:
         return wrap
```
ewjoachim, line 25: This looks like the indentation is wrong ?

ewjoachim (approved):

PR #787, opened 2023-05-11
Update blueprints.py

### LINKED ISSUE #786: Enable the tasks in blueprints.py to handle Pyright's strict mode
Currently, if you do something like this: Pyright on strict mode will complain that "Type of "task" is partially unknown"

By changing the type of Callable to Callable[.... Any] it acknowledges that the decorated function can take any number of arguments and return any value. Which seems like the sort of vague typing you would want for a decorator! 

@app.task(name="my_cool_task")
def check_periodic(timestamp: int) -> None:
    logger.info("Procrastinate is running periodically!")

I opened a PR as well: https://github.com/procrastinate-org/procrastinate/pull/787

paulzakin:
Closes #786

### Successful PR Checklist:

- [ ] Tests
  - [X] (not applicable?)
- [ ] Documentation
  - [X] (not applicable?)

### DIFF
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -174,7 +174,7 @@ class Blueprint:
 
     def task(
         self,
-        _func: Optional[Callable] = None,
+        _func: Optional[Callable[..., Any]] = None,
         *,
         name: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


ewjoachim:
Awesome, thanks !

PR #845, opened 2023-10-05
Log exceptions that trigger job retries at INFO

### LINKED ISSUE #708: Log expected exceptions triggering retries at non-error level
I would like to have `INFO` level logging switched on for Procrastinate's `worker.py` in order to see the start/stop points for the worker itself, along with any errors and warnings about misconfiguration, etc..

However, `worker.py` currently logs _all_ unhanded exceptions from a task at `ERROR`, which results in error-level logging for expected exceptions. This is not desirable where the procrastinate task is being used to make use of the retry mechanism e.g. to repeatedly attempt a call to a remote API that can fail intermittently. In this context., I only want the final exception to be logged as an error (as now that there will be no more retry attempts, it's truly a problem). Prior to that, if the retry strategy is triggered, I believe INFO would be more appropriate.

 Would you be open to this change being made? (Am happy to put a PR together).

ashleyheath:
Closes #708

### Successful PR Checklist:

- [X] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [X] (not applicable?)

### DIFF
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -247,6 +247,7 @@ class Worker:
             if retry_exception:
                 log_title = "Error, to retry"
                 log_action = "job_error_retry"
+                log_level = logging.INFO
                 raise retry_exception from e
             raise exceptions.JobError() from e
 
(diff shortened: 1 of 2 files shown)


ewjoachim (approved):
Seems great !

ewjoachim:
https://github.com/procrastinate-org/procrastinate/releases/tag/0.29.0

PR #873, opened 2023-12-13
Set encoding=utf-8 when reading files

ashleyheath:
(No ticket raised)

To avoid encoding warnings running under python 3.10+.

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/contrib/django/migrations_magic.py
+++ b/procrastinate/contrib/django/migrations_magic.py
@@ -72,7 +72,7 @@ def list_migration_files() -> Iterable[Tuple[str, str]]:
     Returns a list of filenames and file contents for all migration files
     """
     return [
-        (p.name, p.read_text())
+        (p.name, p.read_text(encoding="utf-8"))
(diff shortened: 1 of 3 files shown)


ewjoachim (approved):
I'm a bit surprised that we would get warnings, as I believe utf-8 was the default value, so feel free to add some details on what exactly caused Warnings to appear. That said, "explicit is better than implicit", I don't have an issue stating that.

ewjoachim:
https://github.com/procrastinate-org/procrastinate/releases/tag/0.29.1

ashleyheath:
> I'm a bit surprised that we would get warnings, as I believe utf-8 was the default value, so feel free to add some details on what exactly caused Warnings to appear. That said, "explicit is better than implicit", I don't have an issue stating that.

To clarify, this ocurrs when running with [`PYTHONWARNDEFAULTENCODING`](https://peps.python.org/pep-0597/) (thanks for the merge!)

PR #874, opened 2023-12-14
Issue 871

### LINKED ISSUE #871: `Worker.run_job()` not resiliant to asyncio cancellation
Currently, if a job fails with an error that extends `BaseException` but not `Exception`, the worker will find itself in an invalid state and crash (specifically, the `log_action` variable will not be set when trying to log the outcome of attempting the job).

A seemingly common situation this will arise in is when an asyncio `CancelledError` is raised. Happy to patch, just wanted to clarify what the expected behaviour _should_ be - special case for catching `CancelledError` or catch and supress all `BaseException`s?

ashleyheath:
Closes #871 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

### DIFF
--- a/procrastinate/retry.py
+++ b/procrastinate/retry.py
@@ -17,7 +17,7 @@ class BaseRetryStrategy:
     """
 
     def get_retry_exception(
-        self, exception: Exception, attempts: int
+        self, exception: BaseException, attempts: int
     ) -> Optional[exceptions.JobRetry]:
         schedule_in = self.get_schedule_in(exception=exception, attempts=attempts)
         if schedule_in is None:
@@ -26,7 +26,7 @@ class BaseRetryStrategy:
         schedule_at = utils.utcnow() + datetime.timedelta(seconds=schedule_in)
         return exceptions.JobRetry(schedule_at.replace(microsecond=0))
 
-    def get_schedule_in(self, *, exception: Exception, attempts: int) -> Optional[int]:
+    def get_schedule_in(self, *, exception: BaseException, attempts: int) -> Optional[int]:
         """
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


ewjoachim (commented):
A very promising PR :) We're not far from being in a mergeable state. I've included comments, and on a copy of this branch, I've added a commit for how I would have seen it (it's a suggestion more than anything mandatory, but I have the feeling that it's slightly simpler). Feel free to cherry-pick and/or take part of it.

My idea is that we had 2 "failing" outcomes to `run_job`: fail with retry and fail without retry (2 exceptions). We now have 4 outcomes: critical/non-critical fail with/without retry. This smells a lot like the "inheritence vs composition debate", and indicates (to me) that the difference between retry and fail should be a difference of an attribute, not of a subtype. Sadly, we can't easily change `RetryStrategy.get_retry_exception`'s return type because that's part of the public API (though we could change `Task.get_retry_exception`'s signature, might be better that what I did here).


https://github.com/procrastinate-org/procrastinate/commit/9f5e5be8af0ba1105ecc32e2865e77cf96de6729

ewjoachim:
I've added a commit for taking care of tests in my branch (https://github.com/procrastinate-org/procrastinate/commits/issue-871-ewjoachim/) (both fixing what my previous commit has broken, and adding the test I mentioned).

As it stands, I'd be ready to merge our 2 commits together. But this is your PR, so I'll wait for you opinion before doing so :) And rest assured it's perfectly ok to not agree with my changes!

ashleyheath:
Morning @ewjoachim! Your changes make perfect sense; very happy for you to merge with your additions

PR #1052, opened 2024-05-02
docs: add more context around queueing_locks

### LINKED ISSUE #1050: queueing_lock allows multiple jobs to run concurrently in some scenarios
Let me start by thanking the authors and contributors for this excellent library.

The description of `queueing_lock` is :
> No two jobs with the same queueing lock can be waiting in the queue.

I noticed that when multiple jobs sharing the same `queueing_lock` are deferred and the worker is fast enough to pick one (or more) jobs up, it can end up in multiple jobs with the same `queueing_lock` running concurrently.

Another way to reproduce it is by deferring a job once, and then deferring a second job while the first job is running.

It seems caused by the unique constraint that only considers  jobs in the `todo` status.

Technically, one could argue that when a job is in `doing` status, then it is no longer waiting in the queue.

That said, it appears that the intent of `queueing_lock` is to ensure a single job runs at any time for a given `queueing_lock`. In that case, the unique constraint should be modified to be:

```
CREATE UNIQUE INDEX procrastinate_jobs_queueing_lock_idx ON procrastinate_jobs (queueing_lock) WHERE status = 'todo' or status = 'doing';
```

Maybe I am missing something though and there are valid reasons to ignore `doing` jobs. 
What do you think?

EDIT: I realise the behaviour I am after can be mostly achieved by combining `lock` and `queueing_lock`. It is still not ideal because it would allow the task to be run twice in a row.

onlyann:
Closes #1050 

### Successful PR Checklist:

- [ ] Tests
  - [X] (not applicable?)
- [X] Documentation
  - [ ] (not applicable?)

#### PR label(s): 
  - [X] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20documentation%20%F0%9F%93%9A

### DIFF
--- a/docs/howto/advanced/queueing_locks.md
+++ b/docs/howto/advanced/queueing_locks.md
@@ -33,3 +33,9 @@ can define the value when you register the task:
 def my_task(**kwargs):
     ...
 ```
+
+`queueing_lock` allows a single job in `todo` status. Meanwhile, it allows multiple jobs to be in `doing` status.
+
+To enforce that only one job runs at a time while limiting the queue size, you can combine `queueing_lock` with [lock](./locks.md).
+
+You might also want to [schedule](./schedule.md) jobs for such task in the near future, which can be effective to throttle the task.


ewjoachim (approved):

ewjoachim (approved):
Awesome !

Issue #1040, opened 2024-04-17
Unclear on how to proceed with future migrations


stratosgear:
I tried to understand how the project handles schema migrations but after reading all the documentation pages regarding migrations and browsing through related existing open/closed issues, I have not found a concrete explanation of how it works.... :(

In my use case I have introduced procrastinate in an existing code base (non Django based).

I have executed: `procrastinate -a my.src.app schema --apply` that correctly applied the procrastinate structures required.  Procrastinate seems to be working fine.

My concern now is to how can I remain current with all potential migrations that might be coming along in the future.

I was hoping that I would be able to keep executing `procrastinate -a my.src.app schema --apply` everytime I update my python dependencies and during the project startup, and hopefully automatically catch any potential future migrations required, but I am not sure if this will actually work.

Am I right in thinking that I have to somehow have to adopt any new procrastinate migration scripts as my own and find a way to apply them myself with my existing migration methodology (basically using Alembic)?

Because this is something really fragile and will require a lot of coordinated work to implement, test and maintain.  This will increase the friction of adopting procrastinate too much! :(

Am I missing an obvious solution?

ewjoachim:
I think you're right, except if you use Django.

When we worked on the migration system, we wanted to keep it a bit minimal to avoid tying to a specific system, since there were multiple existing system and choosing one would likely have made it very difficult for people using another one. Since migration systems usually come with their own way of tracking migrations it would be complicated to add our custom way of tracking what has been run or not.

The one may thing we commit to doing is that each [release](https://github.com/procrastinate-org/procrastinate/releases) lists its migrations. Each migration script is written so as to be runnable as-is, if you need to modify it it's probably a bug, and most migration systems accept migrations where you give the SQL code to run directly, so they should be compatible.
Also, we've made a dedicated Django integration that lists procrastinate migrations as Django migrations.

But then you're right, nothing yet has been made to ease that part of the lib.

We could do the same with Alembic as we do with Django, I guess, it would probably cover most of what people would be using. I've never played with alembic so I'd love if someone would like to have a look.

stratosgear:
Although I have not done the full analysis, why Procrastinate cannot handle it's own migrations?

For example Alembic, keeps a table where it notes what was the last migration script that was applied.

[...]

Issue #251, opened 2020-06-17
Fail CI when migrations are targetting the wrong version


ewjoachim:
It would be nice if the CI could fail when a PR contains migration files whose name reference a procrastinate version (a.k.a git tag) that is not the latest. This would avoid things like #250.

This would probably either be a test implemented in pytest, or a short script in a new `bin` folder, alongside a tox target or integrated in check-lint. Whatever feels the most natural to the implementer.

PR #1070, opened 2024-06-04
Add job priority feature

### LINKED ISSUE #386: Priority functionality to procrastinate queues
Procrastinate provides queues for tasks, which I extensively use in my coding. Kudos for that!

Those queues are regular queues.
With priority queues as a feature, task selection from queues would become much more flexible.

I sifted through `procrastinate` code to find out whether procrastinate provides such a functionality, yet, if my memory and skills serve me right, I did not find this.

## In-depth

Assume there is a queue `q`, and the following tasks in it:

id | task | args |
--- | --- | ---
1 | task | some_args_1 |
2 | task | some_args_2 |
3 | task | some_args_3 |

The task with `id=1` goes first, and the `id=2`, and finally `id=3`. 

The cost of tasks are high, and in my case, they need to be ordered to be picked in certain order. And the arrival of tasks are not known beforehand.
While the priority of each task is known at their enqueueing time, such information cannot be reflected to `procrastinate`'s DB records and its task selection routine as of now.

 If that was the case that `procrastinate` works with and pick from a queue based on priority, then it would allow for different workflows:

id | task | args | priority
--- | --- | --- | ---
1 | task | some_args_1 | 5
2 | task | some_args_2 | 10
3 | task | some_args_3 | 15

Then, the first would have been `id=3`, and then `id=2`, and finally `id=1`.

In interactive environments, there might be more enqueue operations that is happening that would make prioritizing tasks  feasible, so that a task that has been enqueue very late yet has high priority can be given higher preference by the task selection algorithm.

`schedule_in` seems to be the closest one to prioritization, that it allows delaying of a task for a later time (for it being picked up). However, abstraction of giving weights to task at their enqueueing is way more understandable than making up `schedule_in` arguments.

A small change in SQL configuration might allow it to work:
```py
task = procrastinate.task(lambda _: _, queue=queue_name, name=task_name, queueing_lock=str(entity['id']))
await task.defer_async(entity=entity)
```

```py
task = procrastinate.task(lambda _: _, queue=queue_name, name=task_name, queueing_lock=some_string)
await task.defer_async(entity=args, priority=procrastinate.priority.HIGH) # enum or similar
```

## Questions
1) is such a functionality already available in `procrastinate`?
2) if not, is such a functionality planned for the future?

medihack:
Closes #386 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### Some design decisions:

- Priority is an optional integer
- The default priority is 0
- Periodic tasks always create a job with priority 0

### Questions:

- I am not sure if I modified `static_migrations.py` correctly.


docs/howto/advanced/priorities.md
```diff
@@ -0,0 +1,29 @@
+# Provide a job priority
+
+We can assign an optional priority to a job. Jobs with higher priority will be
+preferred by a worker. Priority is represented as an integer, where a higher
+number indicates a higher priority. If no priority is specified, it defaults
+to 0.
```
ewjoachim, line 6: (let's be explicit that priorities can be positive or negative)

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
```diff
@@ -0,0 +1,124 @@
+ALTER TABLE procrastinate_jobs ADD COLUMN priority integer DEFAULT 0 NOT NULL;
+
+CREATE UNIQUE INDEX idx_procrastinate_jobs_priority_id ON procrastinate_jobs (priority, id) WHERE status = 'todo';
+
+DROP FUNCTION IF EXISTS procrastinate_defer_job(character varying, character varying, text, text, jsonb, timestamp with time zone);
+CREATE OR REPLACE FUNCTION procrastinate_defer_job(
```
ewjoachim, line 6: Do we need the `DROP` if we use `CREATE OR REPLACE` ? It's very possible that we do, there's something in the back of my mind telling me there's a gotcha in the `CREATE OR REPLACE` thing, and I can't remember what it is.

ewjoachim (commented):
Impressive work on a short timespan! I'll do my best to make a focused review quickly, a few comment while you wait, but it's very nice that you thought about CLI, migrations, Django, the shell etc. Congratz.

Of course, I'm always interested on feedback on the codebase (if you saw surprising or annoying things, etc)

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: I just did it analogous to the existing migrations, but I was also wondering. But it should at least not break something.

medihack:
> Of course, I'm always interested on feedback on the codebase (if you saw surprising or annoying things, etc)

What I saw was quite easy to understand (as far as I can tell after such a short time). Looks like a clean and systematic structure. Its really a gem of a library :-) It should be better known (I added it to [Django Packages](https://djangopackages.org/grids/g/workers-queues-tasks/)).

Is the modification of `static_migrations.py` correct (especially the `dependencies` and the last row)? I only have a vague idea what it does.

ewjoachim:
> Is the modification of static_migrations.py correct (especially the dependencies and the last row)? I only have a vague idea what it does.

I think it's not but it's not your fault, it's mine.
Let me explain:

At first we had no Django.
We had migrations, that were pure SQL
Then we got a django integration that dynamically generated migrations based on the SQL
Then I recently added a better Django integration with models and even if it's not Django's job to create the tables linked to the models, Django still need migrations describing the models, which is the "static migration" (the only migration that is not dynamic)
Ideally, the static migration should have came before the dynamic ones, but I don't want to change old migrations because people are going to have problems if I do that (I should do some kind of squashing I guess).

But then your contribution is the first one that does a normal SQL migration after the introduction of the Django model, and I had absolutely not foreseen that it would be awkward: because we'll need to have the old migrations and then the static and then your migration and all the future ones. Also, I'll need to try and think if the change you made should be backported in the static migration or if it should be another migration. Mixing static and dynamic migrations, sadly, is a recipe for awkward times. I'm considering switching to plain migrations with a bit of tooling to generate it rather than generating them on the fly even though that's one of my favorite piece of Python code 😀 

I'd tend to say: the easiest way out for you is probably to let Django generate the migration it needs with makemigrations (and not change the static migration). You may need a `--merge`. I'll cleanup my own mess :) 

Of **course** if you want to take care of that, you can, but you don't have to.

medihack:
> At first we had no Django. We had migrations, that were pure SQL Then we got a django integration that dynamically generated migrations based on the SQL Then I recently added a better Django integration with models and even if it's not Django's job to create the tables linked to the models, Django still need migrations describing the models, which is the "static migration" (the only migration that is not dynamic) Ideally, the static migration should have came before the dynamic ones, but I don't want to change old migrations because people are going to have problems if I do that (I should do some kind of squashing I guess).
> 
> But then your contribution is the first one that does a normal SQL migration after the introduction of the Django model, and I had absolutely not foreseen that it would be awkward: because we'll need to have the old dynamic migrations and then the static one and then your dynamic migration and all the future dynamic ones. Also, I'll need to try and think if the change you made should be backported in the static migration or if it should be another migration. Mixing static and dynamic migrations, sadly, is a recipe for awkward times. I'm considering switching to plain migrations with a bit of tooling to generate it rather than generating them on the fly even though that's one of my favorite piece of Python code 😀

So, if I understand it correctly, it is there to make Django aware that those models were created. Why not just use [`managed=False`](https://docs.djangoproject.com/en/5.0/ref/models/options/#django.db.models.Options.managed) on the models (see also [legacy databases](https://docs.djangoproject.com/en/5.0/howto/legacy-databases/))? That way you could just keep the dynamic migrations and eliminate the (redundant) static migration. But I may overlook something here. 🤷‍♂️

medihack:
> So, if I understand it correctly, it is there to make Django aware that those models were created. Why not just use [`managed=False`](https://docs.djangoproject.com/en/5.0/ref/models/options/#django.db.models.Options.managed) on the Django models (see also [legacy databases](https://docs.djangoproject.com/en/5.0/howto/legacy-databases/))? That way you could just keep the dynamic migrations and eliminate the (redundant) static migration. But I may overlook something here. 🤷‍♂️

Never mind. I just saw that `managed=False` is already set on the models. But that makes me wonder even more why _Django still need migrations describing the models, which is the "static migration"_. Is this not what `managed=False` is for? To tell Django that is should not care about modification of those models and that we handle them ourselves (which the dynamic migrations already do)? Unfortunately, I am still not sure what to do after reverting the stuff in `static_migrations.py` 😞.

ewjoachim:
I'm really sorry for the mess, it's all my fault :)

I'll push some fixes to your branch as soon as I can :)

ewjoachim:
> But that makes me wonder even more why Django still need migrations describing the models, which is the "static migration". Is this not what managed=False is for? To tell Django that it should not care about modifications of those models and that we handle them ourselves (which the dynamic migrations already do)?

Sadly... `managed=False` tells django that migrations are purely informational, but it still needs to follow the migrations because, as far as I can tell, it needs to understand where each model is coming from and how it evolved, even if it's not django's role to update the schema on the DB side. Kind of like the migrations that track things that don't exist in the DB, like `help_text`.

medihack:
> Sadly... `managed=False` tells django that migrations are purely informational, but it still needs to follow the migrations because, as far as I can tell, it needs to understand where each model is coming from and how it evolved, even if it's not django's role to update the schema on the DB side. Kind of like the migrations that track things that don't exist in the DB, like `help_text`.

Now I get it (at least I think so). Django even generates some pseudo migrations for unmanaged models for some internal state management. Oh, that makes things tricky. How about managing the migrations directly in procrastinate (that would also resolve #1040). It would have its own table where the applied migrations are stored, new migrations are applied in a `pre_migrate` signal ( or outside Django in a new migration command), and Django migrations are just generated by makemigrations. I'm just thinking loud 😉 and I'm also not sure how backward compatibility can be achieved.

> I'm really sorry for the mess, it's all my fault :)

Let's say its Django's fault 😃. They make it quite hard to develop a library that should also be independent from it (don't get me wrong ... I love Django ... it is what I mainly use).

> I'll push some fixes to your branch as soon as I can :)

That's very nice of you! Take your time. No hurry on my side.

ewjoachim:
Ok, I've fixed the issue on main.

After rebasing, you should be able to create a Django migration in procrastinate/contrib/django/migrations looking like this:
```python
from django.db import migrations
from .. import migrations_utils

class Migration(migrations.Migration):
    operations = [migrations_utils.RunProcrastinateSQL(name="02.00.03_01_add_job_priority.sql")]
    name = "0026_add_job_priority"
    dependencies = [("procrastinate", "0025_add_models")]

```

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
ewjoachim: Ok, re-read the [Postgres doc](https://www.postgresql.org/docs/current/sql-createfunction.html). `CREATE OR REPLACE` won't let you change the number or types of arguments or the return type.

It's also noted that if we drop and recreate a function, we need to recreate the corresponding triggers, I wasn't aware of that so it's a good thing I got to re-read that :) The following functions would be impacted: `procrastinate_notify_queue`, `procrastinate_trigger_status_events_procedure_update`, `procrastinate_trigger_status_events_procedure_insert`, `procrastinate_trigger_scheduled_events_procedure`, `procrastinate_unlink_periodic_defers`. I think we never needed to change those, so never got to trigger a bug with that, but I'll know for the future.

In our case it... doesn't hurt to use CREATE OR REPLACE, even if there shouldn't be a case where it's replaced.

Also, there's the question of backwards compatibility. The [rule we have in place](https://github.com/procrastinate-org/procrastinate/blob/main/CONTRIBUTING.md#backward-compatibility) says you should:
- Not delete the old function (so remove the DROP statement)
- Keep a copy of both the old and new functions in schema.sql (the old ones are at the bottom)
- If you want, prepare a future migration in `procrastinate/sql/future_migrations/03.00.00_01_drop_old_functions_before_job_priority.sql` that contains the DROP statements. It will be integrated next time we do a major version (if we ever do one).

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
```diff
@@ -0,0 +1,122 @@
...
+    INSERT INTO procrastinate_jobs (queue_name, task_name, priority, lock, queueing_lock, args, scheduled_at)
+    VALUES (queue_name, task_name, priority, lock, queueing_lock, args, scheduled_at)
+    RETURNING id INTO job_id;
+
+    RETURN job_id;
+END;
+$$;
+
+DROP FUNCTION IF EXISTS procrastinate_defer_periodic_job(character varying, character varying, character varying, character varying, character varying, bigint, jsonb);
+CREATE OR REPLACE FUNCTION procrastinate_defer_periodic_job(
```
ewjoachim, line 30: I think it would make sense if the priority was customizable on periodic jobs, but we don't have to do that on this PR. If you'd rather not, let's create an issue so we don't forget.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
```diff
@@ -0,0 +1,122 @@
...
+                        FROM procrastinate_jobs AS earlier_jobs
+                        WHERE
+                            jobs.lock IS NOT NULL
+                            AND earlier_jobs.lock = jobs.lock
+                            AND earlier_jobs.status IN ('todo', 'doing')
+                            AND earlier_jobs.id < jobs.id)
+                AND jobs.status = 'todo'
+                AND (target_queue_names IS NULL OR jobs.queue_name = ANY( target_queue_names ))
+                AND (jobs.scheduled_at IS NULL OR jobs.scheduled_at <= now())
+            ORDER BY jobs.priority DESC, jobs.id ASC LIMIT 1
```
ewjoachim, line 113: 👍

procrastinate/sql/schema.sql
```diff
@@ -51,7 +52,7 @@ CREATE TABLE procrastinate_events (
     at timestamp with time zone DEFAULT NOW() NULL
 );
 
--- Contraints & Indices
+-- Constraints & Indices
```
ewjoachim, line 55: haha oops :D

procrastinate/tasks.py
```diff
@@ -179,9 +185,11 @@ def configure(self, **options: Unpack[ConfigureTaskOptions]) -> jobs.JobDeferrer
             Converted to schedule_at internally. See `python timedelta documentation
             <https://docs.python.org/3/library/datetime.html#timedelta-objects>`__
             (incompatible with schedule_at)
-
         queue :
             By setting a queue on the job launch, you override the task default queue
+        priority :
+            Set the priority of the job as an integer. Jobs with higher priority
+            are run first. The default priority is 0.
```
ewjoachim, line 192:
```
Priority can be positive or negative.
```

procrastinate/cli.py
```diff
@@ -408,6 +408,15 @@ def configure_defer_parser(subparsers: argparse._SubParsersAction):
         help="Number of seconds after which to launch the job",
         envvar="DEFER_IN",
     )
+    add_argument(
+        defer_parser,
+        "--priority",
+        default=argparse.SUPPRESS,
+        dest="priority",
+        type=int,
+        help="Job priority. Defaults to 0",
```
ewjoachim, line 417:
```
May be positive or negative (higher values indicate jobs that should execute first)
```

procrastinate/testing.py
```diff
@@ -192,17 +192,24 @@ def finished_jobs(self) -> list[JobRow]:
...
                 and (queues is None or job["queue_name"] in queues)
                 and (not job["scheduled_at"] or job["scheduled_at"] <= utils.utcnow())
                 and job["lock"] not in self.current_locks
-            ):
-                job["status"] = "doing"
-                self.events[job["id"]].append({"type": "started", "at": utils.utcnow()})
+            )
+        ]
+
+        filtered_jobs.sort(key=lambda job: job["priority"], reverse=True)
```
ewjoachim, line 206: Hm, I haven't seen `.sort()` in years, I guess it's more usual to used `sorted` that doesn't work by side effect.

Also (unrelated):
I know that Python's sort is stable but for the sake of clarity, we should sort by id (ASC) too. Theas means: `key=lambda job: (-job["priority"], job["id"])` (and no `reverse)`

procrastinate/testing.py
```diff
@@ -192,17 +192,24 @@ def finished_jobs(self) -> list[JobRow]:
...
 
-                return job
+        if filtered_jobs:
+            job = filtered_jobs[0]
+            job["status"] = "doing"
+            self.events[job["id"]].append({"type": "started", "at": utils.utcnow()})
+            return job
 
         return {"id": None}
```
ewjoachim, line 214: I think it's slightly more readable the other way around:
```python
if filtered_jobs:
    return {"id": None} 
```

This lets more code avoid indentation. As a general rule, and with possible exceptions, I tend to try and identify the special case that I can handle first (especially if they allow returning early) and try to have the general case be the one that is as little indented as possible, and goes all the way to the bottom of the function. This avoids the "tower of if-statements" when possible.

docs/howto/advanced/priorities.md
```diff
@@ -0,0 +1,29 @@
...
+
```
ewjoachim, line 7: I think it might be useful to state the following even it may seem a bit obvious if we give it some thought (feel free to rephrase as needed):

> Priority is used as way to order _available_ jobs. If a procrastinate worker requests a task, and the 2 scheduled tasks are a high-priority task that is blocked by a lock, and a low-priority that is available, the worker will take the low-priority task. Procrastinate will never wait for a high-priority task to become available if there are lower-priority tasks already available.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
```diff
@@ -1,5 +1,7 @@
 ALTER TABLE procrastinate_jobs ADD COLUMN priority integer DEFAULT 0 NOT NULL;
 
+CREATE UNIQUE INDEX idx_procrastinate_jobs_priority_id ON procrastinate_jobs (priority, id) WHERE status = 'todo';
```
ewjoachim, line 3: Excellent idea! Did you benchmark in some way?

ewjoachim (commented):
Ok, that was a good first round of review. Overall, I'm impressed by the quality of your (first time) contribution.

I'll reread my comments, but as far as I can tell, only the migration + backwards compatibility stuff is really something we need to fix, the rest is very minor.

Thanks again for your time, let's get this merged soon !

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: Nope, no benchmarking ... just divine inspiration 😂. If there are many jobs, many of them have to be sorted by priority. So, an index will help. I guess, in many cases, it won't make any difference; on the other hand, it's only a small index.

ewjoachim: Ah, then can I ask you to `EXPLAIN ANALYZE SELECT procrastinate_defer_job()` and check that the index is actually used?

There are already other indexes, maybe what we need is to modify an existing index rather than create new ones ?

docs/howto/advanced/priorities.md
medihack: I am a bit confused by the terms `job` and `task`. You mean ...

>  Priority is used as way to order available jobs. If a procrastinate worker requests a **_job_**, and the 2 scheduled **_jobs_** are a high-priority **_job_** that is blocked by a lock, and a low-priority that is available, the worker will take the low-priority **_job_**. Procrastinate will never wait for a high-priority **_job_** to become available if there are lower-priority **_jobs_** already available.

Or do you use the word job and task synonymously? I see jobs more as instances of a task and tasks more as a template or the code that gets executed behind it. But maybe I am pedantic 😉

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: Sure, makes sense!👍

docs/howto/advanced/priorities.md
ewjoachim: You're definitely not pedantic, it's confusing and it's my fault. The 2 terms are  explained [here](https://procrastinate.readthedocs.io/en/stable/glossary.html#). When we started the project, we needed to decide what would be the word for the function that can be deferred, and what would be the word for a function deferred with a specific set of arguments, options, etc. We kinda arbitrarily settled for "task" and "job". And the fact that I, after having spent years on the project, frequently get those mixed up, is a continued proof that these were not the right terms. I'm afraid this ship has sailed much too far away to get this fixed.

Yes, in my paragraph above, I meant `job` everywhere, tasks (for now) are unconcerned with priorities. (That said, tasks could have a default priority, as they have a default queue, lock, etc)

medihack:
> Ok, I've fixed the issue on main.
> 
> After rebasing, you should be able to create a Django migration in procrastinate/contrib/django/migrations looking like this:
> 
> ```python
> from django.db import migrations
> from .. import migrations_utils
> 
> class Migration(migrations.Migration):
>     operations = [migrations_utils.RunProcrastinateSQL(name="02.00.03_01_add_job_priority.sql")]
>     name = "0026_add_job_priority"
>     dependencies = [("procrastinate", "0025_add_models")]
> ```

I guess we still need to put `priority` into a Django model migration. So, I would create two migration files: one for the SQL and one for the model itself (`migrations.AddField`). Right?

ewjoachim:
Ah the Django AddField may be in the same migration as the other one, it's not a problem. With the new system we can edit migrations manually, so let's create only 1 django migration per PR if we don't actually need to do otherwise.

docs/howto/advanced/priorities.md
medihack: Thanks for the clarification. Naming stuff is always hard. The funny thing is that in the pipeline where I want to integrate Procrastinate, it's exactly the other way around. A job consists of several tasks 😂

procrastinate/testing.py
medihack: > Hm, I haven't seen `.sort()` in years, I guess it's more usual to used `sorted` that doesn't work by side effect.

Yeah, I am more of an older guy (quite an old profile picture 😉).

> Also (unrelated): I know that Python's sort is stable but for the sake of clarity, we should sort by id (ASC) too. Theas means: `key=lambda job: (-job["priority"], job["id"])` (and no `reverse)`

That's why I used `sort` instead of `sorted` initially, as I also sorted by `id` (and I could have it in one lambda). Then I thought it was not necessary, as it is initially sorted by `id`. But, of course, let's make it explicit.

procrastinate/testing.py
medihack: Sounds good. I am always in favor when it comes to improving code structure.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: I also considered that, but I was not sure if it was really necessary or what the API would look like. Let's create a new issue for that; this one is large enough now 😉 But I can work on that, too.

medihack: Tracked in #1072

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: Ok, this is what I get:

```
procrastinate=# EXPLAIN ANALYZE SELECT id, status, task_name, priority, lock, queueing_lock, args, scheduled_at, queue_name, attempts FROM procrastinate_fetch_job(ARRAY['index']);
                                                       QUERY PLAN                                                       
------------------------------------------------------------------------------------------------------------------------
 Function Scan on procrastinate_fetch_job  (cost=0.25..0.26 rows=1 width=672) (actual time=0.395..0.395 rows=1 loops=1)
 Planning Time: 0.037 ms
 Execution Time: 0.408 ms
(3 rows)
```

It seems that this index isn't used at all 😳. Does this mean it is simply not necessary? Or is it wrongly defined? I think I need your help here 🙏.

medihack: I did some further experiments. Here are some results with about 150,000 jobs waiting to be processed and a random priority for each job.

```
procrastinate=# EXPLAIN ANALYZE SELECT id, status, task_name, priority, lock, queueing_lock, args, scheduled_at, queue_name, attempts FROM procrastinate_fetch_job(ARRAY['index']);
                                                         QUERY PLAN                                                         
----------------------------------------------------------------------------------------------------------------------------
 Function Scan on procrastinate_fetch_job  (cost=0.25..0.26 rows=1 width=672) (actual time=128.042..128.043 rows=1 loops=1)
 Planning Time: 0.042 ms
 Execution Time: 128.057 ms
(3 rows)
```

It is still quite fast, even without an index that is being used. So should we just go without any additional index?

ewjoachim: Seeing your results, I'd say you'd have more details if you ran the explain on the query that is ran by the procedure rather than running the procedure itself so:

```
EXPLAIN ANALYZE SELECT jobs.*
FROM procrastinate_jobs AS jobs
WHERE
    -- reject the job if its lock has earlier jobs
    NOT EXISTS (
        SELECT 1
            FROM procrastinate_jobs AS earlier_jobs
            WHERE
                jobs.lock IS NOT NULL
                AND earlier_jobs.lock = jobs.lock
                AND earlier_jobs.status IN ('todo', 'doing')
                AND earlier_jobs.id < jobs.id)
    AND jobs.status = 'todo'
    AND (target_queue_names IS NULL OR jobs.queue_name = ANY( target_queue_names ))
    AND (jobs.scheduled_at IS NULL OR jobs.scheduled_at <= now())
ORDER BY jobs.priority DESC, jobs.id ASC LIMIT 1
```

Also, feel free to try with 10x more jobs (or more) if it helps make the effect of the index visible

medihack: Ok, here are some more results:

For 1.000.000 jobs
```
procrastinate=# EXPLAIN ANALYZE SELECT id, status, task_name, priority, lock, queueing_lock, args, scheduled_at, queue_name, attempts FROM procrastinate_fetch_job(ARRAY['index']);
                                                        QUERY PLAN                                                        
--------------------------------------------------------------------------------------------------------------------------
 Function Scan on procrastinate_fetch_job  (cost=0.25..0.26 rows=1 width=672) (actual time=95.765..95.766 rows=1 loops=1)
 Planning Time: 0.040 ms
 Execution Time: 95.780 ms
(3 rows)
```

```
procrastinate=# EXPLAIN ANALYZE SELECT jobs.*                                                                                                                                      
            FROM procrastinate_jobs AS jobs
            WHERE
                -- reject the job if its lock has earlier jobs
                NOT EXISTS (
                    SELECT 1
                        FROM procrastinate_jobs AS earlier_jobs
                        WHERE
                            jobs.lock IS NOT NULL
                            AND earlier_jobs.lock = jobs.lock
                            AND earlier_jobs.status IN ('todo', 'doing')
                            AND earlier_jobs.id < jobs.id)
                AND jobs.status = 'todo'
                AND (jobs.scheduled_at IS NULL OR jobs.scheduled_at <= now())
            ORDER BY jobs.priority DESC, jobs.id ASC LIMIT 1;
			
QUERY PLAN                                                                                                       
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=11885.73..11885.85 rows=1 width=191) (actual time=54.537..59.374 rows=1 loops=1)
   ->  Gather Merge  (cost=11885.73..22189.50 rows=89598 width=191) (actual time=54.536..59.371 rows=1 loops=1)
         Workers Planned: 1
         Workers Launched: 1
         ->  Sort  (cost=10885.72..11109.72 rows=89598 width=191) (actual time=52.067..52.070 rows=1 loops=2)
               Sort Key: jobs.priority DESC, jobs.id
               Sort Method: top-N heapsort  Memory: 25kB
               Worker 0:  Sort Method: top-N heapsort  Memory: 25kB
               ->  Parallel Hash Anti Join  (cost=4548.28..10437.73 rows=89598 width=191) (actual time=11.318..34.528 rows=76164 loops=2)
                     Hash Cond: (jobs.lock = earlier_jobs.lock)
                     Join Filter: ((jobs.lock IS NOT NULL) AND (earlier_jobs.id < jobs.id))
                     ->  Parallel Seq Scan on procrastinate_jobs jobs  (cost=0.00..4534.28 rows=89598 width=191) (actual time=0.013..14.251 rows=76164 loops=2)
                           Filter: ((status = 'todo'::procrastinate_job_status) AND ((scheduled_at IS NULL) OR (scheduled_at <= now())))
                           Rows Removed by Filter: 10
                     ->  Parallel Hash  (cost=3754.97..3754.97 rows=63465 width=40) (actual time=10.003..10.004 rows=76174 loops=2)
                           Buckets: 262144  Batches: 1  Memory Usage: 2048kB
                           ->  Parallel Index Only Scan using procrastinate_jobs_id_lock_idx on procrastinate_jobs earlier_jobs  (cost=0.42..3754.97 rows=63465 width=40) (actual time=0.027..7.096 rows=76174 loops=2)
                                 Heap Fetches: 53
 Planning Time: 0.444 ms
 Execution Time: 59.413 ms
(20 rows)
```

And also 2.000.000 jobs
```
procrastinate=# EXPLAIN ANALYZE SELECT id, status, task_name, priority, lock, queueing_lock, args, scheduled_at, queue_name, attempts FROM procrastinate_fetch_job(ARRAY['index']);
                                                        QUERY PLAN                                                        
--------------------------------------------------------------------------------------------------------------------------
 Function Scan on procrastinate_fetch_job  (cost=0.25..0.26 rows=1 width=672) (actual time=97.055..97.055 rows=1 loops=1)
 Planning Time: 0.050 ms
 Execution Time: 97.070 ms
(3 rows)
```

```
procrastinate=# EXPLAIN ANALYZE SELECT jobs.*                                                                                                                                      
            FROM procrastinate_jobs AS jobs
            WHERE
                -- reject the job if its lock has earlier jobs
                NOT EXISTS (
                    SELECT 1
                        FROM procrastinate_jobs AS earlier_jobs
                        WHERE
                            jobs.lock IS NOT NULL
                            AND earlier_jobs.lock = jobs.lock
                            AND earlier_jobs.status IN ('todo', 'doing')
                            AND earlier_jobs.id < jobs.id)
                AND jobs.status = 'todo'
                AND (jobs.scheduled_at IS NULL OR jobs.scheduled_at <= now())
            ORDER BY jobs.priority DESC, jobs.id ASC LIMIT 1;
			

QUERY PLAN                                                                                                       
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=11885.73..11885.85 rows=1 width=191) (actual time=67.032..71.506 rows=1 loops=1)
   ->  Gather Merge  (cost=11885.73..22189.50 rows=89598 width=191) (actual time=67.031..71.504 rows=1 loops=1)
         Workers Planned: 1
         Workers Launched: 1
         ->  Sort  (cost=10885.72..11109.72 rows=89598 width=191) (actual time=64.441..64.444 rows=1 loops=2)
               Sort Key: jobs.priority DESC, jobs.id
               Sort Method: top-N heapsort  Memory: 25kB
               Worker 0:  Sort Method: top-N heapsort  Memory: 25kB
               ->  Parallel Hash Anti Join  (cost=4548.28..10437.73 rows=89598 width=191) (actual time=14.048..43.027 rows=76164 loops=2)
                     Hash Cond: (jobs.lock = earlier_jobs.lock)
                     Join Filter: ((jobs.lock IS NOT NULL) AND (earlier_jobs.id < jobs.id))
                     ->  Parallel Seq Scan on procrastinate_jobs jobs  (cost=0.00..4534.28 rows=89598 width=191) (actual time=0.015..18.812 rows=76164 loops=2)
                           Filter: ((status = 'todo'::procrastinate_job_status) AND ((scheduled_at IS NULL) OR (scheduled_at <= now())))
                           Rows Removed by Filter: 10
                     ->  Parallel Hash  (cost=3754.97..3754.97 rows=63465 width=40) (actual time=12.739..12.740 rows=76174 loops=2)
                           Buckets: 262144  Batches: 1  Memory Usage: 2048kB
                           ->  Parallel Index Only Scan using procrastinate_jobs_id_lock_idx on procrastinate_jobs earlier_jobs  (cost=0.42..3754.97 rows=63465 width=40) (actual time=0.024..9.350 rows=76174 loops=2)
                                 Heap Fetches: 53
 Planning Time: 0.382 ms
 Execution Time: 71.563 ms
(20 rows)
```

So, my index isn't used at all, but the performance does not seem to degrade anymore with more jobs. I will delete my added index.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
```diff
@@ -0,0 +1,124 @@
+ALTER TABLE procrastinate_jobs ADD COLUMN priority integer DEFAULT 0 NOT NULL;
+
+CREATE UNIQUE INDEX idx_procrastinate_jobs_priority_id ON procrastinate_jobs (priority, id) WHERE status = 'todo';
```
ewjoachim, line 3: You removed the index from the schema.sql file but not from the migration.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
ewjoachim: I think the backwards compat element is the only thing left before we can merge ? I'll re-read and double check.

ewjoachim:
(I can take over for the backwards compat migration if you prefer)

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: Uups 🙄. Done.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
ewjoachim: Good for me.

procrastinate/sql/migrations/02.00.03_01_add_job_priority.sql
medihack: Thanks for the reminder. I'll look at the backward-compatible migrations later (or maybe tomorrow). I'm not exactly sure about it, but I will figure it out.

ewjoachim: Honnestly, I think just not deleting the old functions and keeping both old and new functions in schema.sql should be enough for the functions that change signature. 

For the functions that don't change signature, you probably don't have anything to do.


Keeping in mind that the new priority column has a default and if every task has priority zero, then it works the same as before, I don't foresee specific issues.

ewjoachim: So removing this line:
https://github.com/procrastinate-org/procrastinate/pull/1070/files#diff-78aec0cd2e7f3a027ae5711665e34f4068d8bea1ecfaeca5cd35afca2c26ba76R3

And duplicating the `procrastinate_defer_job` function in schema.sql so that both the old function without the pirority param and the new function with the priority param exist (put the legacy function at the bottom)

medihack: Ah, now I get it. It works because PostgreSQL supports function overloading. I hope it's ok now.

By the way ... shouldn't the old `procrastinate_finish_job` function shouldn't be removed by now? Above the old function, it says "to be removed after 2.0.0," and above the new "it is to be removed after 1.0.0 is released". Is it ok if I clean the comments up a bit (maybe just add at the "Old versions of functions" title that those will be removed with a version 3 release)?

ewjoachim: I'll need to double check, this is probably just the comment that's outdated. I'll check outside of the PR.

ewjoachim (approved):
Congratulations for an excellent first contribution ! It was awesome working with you on this, and I'd love to see you contributing again :)

medihack:
> Congratulations for an excellent first contribution ! It was awesome working you you on this, and I'd love to see you contributing again :)

It was my pleasure, and I also learned something along the way. As promised, I will work on #1072 in the upcoming days.

medihack:
@alexwilson1 Have you applied the [migration files](https://procrastinate.readthedocs.io/en/stable/howto/production/migrations.html)? In your case, it sounds like you have to apply:
02.05.00_01_add_periodic_job_priority.sql
02.06.00_01_add_cancel_states.sql

ewjoachim:
> @ewjoachim It seems that `02.00.03_01_add_job_priority.sql` isn't named correctly (it should be `02.04.*`, or? Maybe associated with #251). I think this is because there were releases after I started the PR.

Yep, I made a note of it when I realized that [on the release](https://github.com/procrastinate-org/procrastinate/releases/tag/2.5.0). I didn't want to rename the migration post-release because I was afraid it would cause issues.

There's a detail that should appear in the doc and doesn't: you can call `procrastinate schema --migrations-path` to get the path in which migrations are stored in a venv where procrastinate is installed.

ewjoachim:
If that's ok with you, let's discuss [there](https://github.com/procrastinate-org/procrastinate/pull/1086) instead

PR #1081, opened 2024-06-23
Allow to cancel a job or request an abort

### LINKED ISSUE #511: A canceled task that completes fails with "Job was not found or not in "doing" or "todo" status"
A task that is canceled has it status set to `FAILED`, however when the task completes only `DOING` and `TODO` is accepted for task completion. This causes `procrastinate` to fail and terminate it's process.

https://github.com/procrastinate-org/procrastinate/blob/65829b3fc17374985ee17c2911083f60132f1bfd/procrastinate/shell.py#L142 https://github.com/procrastinate-org/procrastinate/blob/65829b3fc17374985ee17c2911083f60132f1bfd/procrastinate/sql/schema.sql#L207

```
Dec 30 16:24:00 host1 python3[1280546]: Error:
Dec 30 16:24:00 host1 python3[1280546]:     Database error.
Dec 30 16:24:00 host1 python3[1280546]:     
Dec 30 16:24:00 host1 python3[1280546]: Job was not found or not in "doing" or "todo" status (job id: 47)
Dec 30 16:24:00 host1 python3[1280546]: CONTEXT:  PL/pgSQL function procrastinate_finish_job(integer,procrastinate_job_status,timestamp with time zone,boolean) line 24 at RAISE
Dec 30 16:24:00 host1 systemd[1]: procrastinate.service: Main process exited, code=exited, status=1/FAILURE
Dec 30 16:24:00 host1 systemd[1]: procrastinate.service: Failed with result 'exit-code'.
```

medihack:
Closes #511 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

The only thing it breaks is when using the `shell` that `cancel` will leave the `job` in `cancelled` status (before the change it was left with `failed`). But I see it more as a fix than a breaking change.

The design decisions should be clear from the documentation page.

There are maybe parts to improve (maybe the acceptance tests?). Providing and testing sync and async APIs is quite hard, and sometimes also quite ugly 🙄. And in the end, it was more additional code than I thought.

### DIFF
--- a/docs/howto/advanced/cancellation.md
+++ b/docs/howto/advanced/cancellation.md
@@ -1,0 +1,71 @@
+# Cancel a job
+
+We can cancel a job that has not yet been processed by a worker. We can also
+mark a job that is currently being processed for abortion, but this request
+has to be handled by the task itself.
+
+## Cancel a job (that is not being processed yet)
+
+```python
+# by using the sync method
+app.job_manager.cancel_job_by_id(33)
+# or by using the async method
+await app.job_manager.cancel_job_by_id_async(33)
+```
+
+## Delete the canceled job
+
+A cancelled job can also be deleted from the database.
+
+```python
+# by using the sync method
+app.job_manager.cancel_job_by_id(33, delete_job=True)
+# or by using the async method
+await app.job_manager.cancel_job_by_id_async(33, delete_job=True)
+```
+
+## Mark a currently being processed job for abortion
+
+If a worker has not picked up the job yet, the below command behaves like the
+command without the `abort` option. But if a job is already in the middle of
+being processed, the `abort` option marks this job for abortion (see below
+how to handle this request).
+
+```python
+# by using the sync method
+app.job_manager.cancel_job_by_id(33, abort=True)
+# or by using the async method
+await app.job_manager.cancel_job_by_id_async(33, abort=True)
+```
+
+## Handle a abortion request inside the task
+
+In our task, we can check (for example, periodically) if the task should be
+aborted. If we want to respect that request (we don't have to), we raise a
+`JobAborted` error.
+
+```python
+@app.task(pass_context=True)
+def my_task(context):
+  for i in range(100):
+    if context.should_abort():
+      raise exceptions.JobAborted
+    do_something_expensive()
+```
+
+There is also an async API
+
+```python
+@app.task(pass_context=True)
+async def my_task(context):
+  for i in range(100):
+    if await context.should_abort_async():
+      raise exceptions.JobAborted
+    do_something_expensive()
+```
+
+:::{warning}
+`context.should_abort()` and `context.should_abort_async()` does poll the
+database and might flood the database. Ensure you do it only sometimes and
+not from too many parallel tasks.
+:::
--- a/docs/howto/advanced/cancellation.md
+++ b/docs/howto/advanced/cancellation.md
@@ -42,7 +42,8 @@ await app.job_manager.cancel_job_by_id_async(33, abort=True)
 
 In our task, we can check (for example, periodically) if the task should be
 aborted. If we want to respect that request (we don't have to), we raise a
-`JobAborted` error.
+`JobAborted` error. Any message passed to `JobAborted` (e.g.
+`raise JobAborted("custom message")`) will end up in the logs.
 
 ```python
 @app.task(pass_context=True)
--- a/docs/howto/advanced/cancellation.md
+++ b/docs/howto/advanced/cancellation.md
@@ -13,7 +13,7 @@ app.job_manager.cancel_job_by_id(33)
 await app.job_manager.cancel_job_by_id_async(33)
 ```
 
-## Delete the canceled job
+## Delete the cancelled job
 
 A cancelled job can also be deleted from the database.
 
--- a/procrastinate/contrib/django/models.py
+++ b/procrastinate/contrib/django/models.py
@@ -64,6 +64,9 @@ class ProcrastinateJob(ProcrastinateReadOnlyModelMixin, models.Model):
         "doing",
         "succeeded",
         "failed",
+        "cancelled",
+        "aborting",
+        "aborted",
     )
     id = models.BigAutoField(primary_key=True)
     queue_name = models.CharField(max_length=128)
@@ -91,6 +94,8 @@ class ProcrastinateEvent(ProcrastinateReadOnlyModelMixin, models.Model):
         "failed",
         "succeeded",
         "cancelled",
+        "request_to_abort",
+        "aborted",
         "scheduled",
     )
     id = models.BigAutoField(primary_key=True)
--- a/procrastinate/contrib/django/models.py
+++ b/procrastinate/contrib/django/models.py
@@ -94,7 +94,7 @@ class ProcrastinateEvent(ProcrastinateReadOnlyModelMixin, models.Model):
         "failed",
         "succeeded",
         "cancelled",
-        "request_to_abort",
+        "abort_requested",
         "aborted",
         "scheduled",
     )
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -63,6 +63,12 @@ class JobError(ProcrastinateException):
         self.critical = critical
 
 
+class JobAborted(ProcrastinateException):
+    """
+    Job was aborted (usually as reaction to an abortion request).
+    """
+
+
 class AppNotOpen(ProcrastinateException):
     """
     App was not open. Procrastinate App needs to be opened using:
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -65,7 +65,7 @@ class JobError(ProcrastinateException):
 
 class JobAborted(ProcrastinateException):
     """
-    Job was aborted (usually as reaction to an abortion request).
+    Job was aborted.
     """
 
 
--- a/procrastinate/manager.py
+++ b/procrastinate/manager.py
@@ -215,13 +215,13 @@ class JobManager:
         delete_job: bool,
     ) -> None:
         """
-        Set a job to its final state (``succeeded`` or ``failed``).
+        Set a job to its final state (``succeeded``, ``failed`` or ``aborted``).
 
         Parameters
         ----------
         job : `jobs.Job`
         status : `jobs.Status`
-            ``succeeded`` or ``failed``
+            ``succeeded``, ``failed`` or ``aborted``
         """
         assert job.id  # TODO remove this
         await self.finish_job_by_id_async(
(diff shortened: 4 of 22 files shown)


procrastinate/sql/schema.sql
```diff
@@ -198,12 +203,12 @@ AS $$
 DECLARE
```
ewjoachim, line 203: procrastinate_fetch_job above hasn't been updated. This means that the second you request a job abortion, the next job with the same lock will start.

procrastinate/exceptions.py
```diff
@@ -63,6 +63,12 @@ def __init__(
         self.critical = critical
 
 
+class JobAborted(ProcrastinateException):
+    """
+    Job was aborted (usually as reaction to an abortion request).
```
ewjoachim, line 68: This message will end up in the logs unless users put a special message. It's less documentation than it's communication. I think `Job was aborted` should be good.

docs/howto/advanced/cancellation.md
```diff
@@ -0,0 +1,71 @@
...
+In our task, we can check (for example, periodically) if the task should be
+aborted. If we want to respect that request (we don't have to), we raise a
+`JobAborted` error.
+
+```python
+@app.task(pass_context=True)
+def my_task(context):
+  for i in range(100):
+    if context.should_abort():
+      raise exceptions.JobAborted
```
ewjoachim, line 52: It might be worth nothing that whatever message is passed to the exception will end up in the logs.

tests/acceptance/test_async.py
```diff
@@ -39,3 +44,83 @@ async def product_task(a, b):
...
+    await async_app.run_worker_async(queues=["default"], wait=False)
+
+    assert sum_results == [7]
+
+
+async def test_abort(async_app):
+    @async_app.task(queue="default", name="task1", pass_context=True)
+    async def task1(context):
+        while True:
+            await asyncio.sleep(1)
```
ewjoachim, line 96: Computers are fast. Let's wait much less than this unless we need to wait for this much.

https://www.youtube.com/watch?v=6xijhqU8r2A

Suggestion: 0.1s

tests/unit/test_job_context.py
```diff
@@ -104,3 +104,43 @@ def test_job_description_job_time(job_factory):
...
     ).job_description(current_timestamp=30.0)
     assert descr == "worker 2: some_task[12](a='b') (started 10.000 s ago)"
+
+
+def test_should_abort(app, job_factory):
+    job = job_factory(id=12)
+    app.job_manager.get_job_status = lambda job_id: jobs.Status.ABORTING
+    context = job_context.JobContext(app=app, job=job)
+
+    assert context.should_abort() is True
```
ewjoachim, line 114: You have implemented the InMemoryConnector. You don't need to overwrite get_job_status, you can just simulate a normal job. (same below)

Also, this and the code below only differ by `jobs.Status.ABORTING/jobs.Status.DOING` and `True/False`. Looks like it could be a single parametrized test (though this advice is probably incompatible with the one above :D You choose, but I prefer the one above.)

ewjoachim (commented):
Looking awesome ! A few comments.

Thanks a lot for your work !

procrastinate/contrib/django/models.py
```diff
@@ -91,6 +94,8 @@ class ProcrastinateEvent(ProcrastinateReadOnlyModelMixin, models.Model):
         "failed",
         "succeeded",
         "cancelled",
+        "request_to_abort",
```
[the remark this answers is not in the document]
medihack: I will use `abort_requested` then.

procrastinate/manager.py
```diff
@@ -241,6 +241,104 @@ async def finish_job_by_id_async(
...
+            itself has to respect the abortion request. If False, only jobs in ``todo`` state
+            will be set to ``cancelled`` and won't be processed by a worker anymore.
+        delete_job : ``bool``
+            If True, the job will be deleted from the database after being cancelled. Does
+            not affect the jobs that should be aborted.
+
+        Returns
+        -------
+        ``bool``
+            True if the job to be cancelled was found, False otherwise.
```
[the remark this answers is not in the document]
medihack: Alright, I tried to be clearer about this. I also added an assertion about what the returned `id` really is.

docs/howto/advanced/cancellation.md
```diff
@@ -0,0 +1,71 @@
...
+  for i in range(100):
+    if await context.should_abort_async():
+      raise exceptions.JobAborted
+    do_something_expensive()
+```
+
+:::{warning}
+`context.should_abort()` and `context.should_abort_async()` does poll the
+database and might flood the database. Ensure you do it only sometimes and
+not from too many parallel tasks.
```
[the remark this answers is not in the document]
medihack: Nice idea. Let me tackle this in another PR. This one is already large enough.

docs/howto/advanced/cancellation.md
medihack: Ok, I added a line to the documentation page.

procrastinate/sql/migrations/02.06.00_01_add_cancel_states.sql
```diff
@@ -0,0 +1,115 @@
...
+        WHERE id = job_id AND status IN ('todo', 'doing', 'aborting')
+        RETURNING id INTO _job_id;
+    END IF;
+    IF _job_id IS NULL THEN
+        RAISE 'Job was not found or not in "doing", "todo" or "aborting" status (job id: %)', job_id;
+    END IF;
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION procrastinate_trigger_status_events_procedure_update()
```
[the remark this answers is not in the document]
medihack: And so the circle closes 😉. Does this mean I have to "create or replace" the `procrastinate_trigger_status_events_update` trigger (because this is the one that executes this procedure)?

procrastinate/sql/schema.sql
medihack: Unfortunately, I don't fully understand your comment 🤷‍♂️. Do I have to change something in `procrastinate_fetch_job`, too?

tests/unit/test_job_context.py
medihack: Ok, I simulated the normal job cycle (with fetching and cancelling it).

procrastinate/sql/migrations/02.06.00_01_add_cancel_states.sql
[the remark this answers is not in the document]
ewjoachim: I guess so. If you can do a quick manual check by applying the previous schema, then the migration, then have the trigger execute and check that it's the right code, I'll be convinved, no need for an automated test or anything :)

procrastinate/sql/schema.sql
ewjoachim: I think so, here's what an integration test for the correct behaviour would look like (I'm on my phone, so this is pseudocode)

```python
@app.task(look="foo")
def task():
    pass

job1 = task.defer()
job2 = task.defer()

fetch_job1 = app.job_manager.fetch_job()
fetch_job1.cancel(abort=True)

assert app.job_manager.fetch_job() is None
```

When a job is aborting, it's not finished yet, locks should still apply.

procrastinate/sql/migrations/02.06.00_01_add_cancel_states.sql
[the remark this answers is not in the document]
medihack: The [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-createfunction.html) also says:

> Use CREATE OR REPLACE FUNCTION to change a function definition without breaking objects that refer to the function

And this is what we do here. But I can double-check and try it out.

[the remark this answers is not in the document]
medihack: Ok, I tried it out. It still seems to work without recreating the trigger. Here is what I did:

1. Started with an empty db
2. Checked out the main branch
3. Applied the schema
4. Ran the async demo with one job
5. Checked that the events were triggered (in the db table)
6. Checked out the cancel-job-feature branch
7. Applied the add_cancel_states migration
8. Ran the async demo again with one job
9. Checked that events were triggered again

Everything seemed to work fine. The resulting table:

```
procrastinate=# SELECT * FROM procrastinate_events;
 id | job_id |   type    |              at               
----+--------+-----------+-------------------------------
  1 |      1 | deferred  | 2024-06-25 09:31:44.600697+00
  2 |      1 | started   | 2024-06-25 09:31:44.613338+00
  3 |      1 | succeeded | 2024-06-25 09:31:44.623551+00
  4 |      2 | deferred  | 2024-06-25 09:39:07.207435+00
  5 |      2 | started   | 2024-06-25 09:39:07.219847+00
  6 |      2 | succeeded | 2024-06-25 09:39:07.231104+00
(6 rows)
```

procrastinate/sql/schema.sql
medihack: It's amazing what you can write on your mobile 😃. Of course, I totally forgot the locks. Thanks a lot, that would have been a nasty bug. I hope I fixed it now.

procrastinate/sql/migrations/02.06.00_01_add_cancel_states.sql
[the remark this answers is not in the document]
ewjoachim: Perfect.

procrastinate/manager.py
```diff
@@ -241,6 +241,116 @@ async def finish_job_by_id_async(
...
+            state will be set to ``cancelled`` and won't be processed by a worker anymore.
+        delete_job : ``bool``
+            If True, the job will be deleted from the database after being cancelled. Does
+            not affect the jobs that should be aborted.
+
+        Returns
+        -------
+        ``bool``
+            True if the job to be cancelled (or to mark for abortion) was found. False if
+            no job with that ID or in a state to be cancelled was found.
```
ewjoachim, line 266: Sorry, the sentence is correct but I find it a bit hard to parse.
Can I suggest:
```suggestion
            If True, the job was canceled (or aborted). If False, nothing was done: either there is no job with this id or it's not in a state where it may be canceled (`todo` or `doing`)
```

WDYT?

ewjoachim (approved):
One last comment, but it's minor. Ok for the rest. I'll leave some time if you want to resolve the last item, and we can merge!

procrastinate/manager.py
medihack: Absolutely, yours sounds better 🙂.

medihack: Ok, I updated those docstrings and replaced `canceled` with `cancelled` in some other locations to be consistent. It looks like we have settled on British English.

ewjoachim (approved):

ewjoachim:
Congratulations :)

medihack:
> Congratulations :)

Thank you for the super nice review process and the welcoming manner. I'll integrate Procrastinate in our applications in the upcoming days, but I will soon be back to take care of #1083. I also open another issue to track and discuss a possible listen-notify abort feature on top of this one.

PR #1038, opened 2024-04-13
Enable using ProcrastinateShell.onecmd from cli shell

indrat:
Allows invoking the shell commands without entering the cmdloop

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20breaking%20%F0%9F%92%A5
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20feature%20%E2%AD%90%EF%B8%8F
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20bugfix%20%F0%9F%95%B5%EF%B8%8F
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20miscellaneous%20%F0%9F%91%BE
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20dependencies%20%F0%9F%A4%96
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20documentation%20%F0%9F%93%9A

PR #1095, opened 2024-07-03
fix: add default args to cli shell command to avoid missing args error

linspw:
Closes #<ticket number>

Hi! How is it going?

First, thank you so much for this project; it is amazing!

The problem that this PR proposes to resolve occurs when we try to use `./manage.py procrastinate shell` in django. The command raises the following error:

```
2024-07-03 13:42:31,117 DEBUG   procrastinate.cli Exception details:
  File "/usr/local/lib/python3.11/site-packages/procrastinate/cli.py", line 522, in execute_command
    await parsed.pop("func")(app=app, **parsed)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: shell_() missing 1 required positional argument: 'args'
shell_() missing 1 required positional argument: 'args'
```

With that simple change, everthing works normally.

### Successful PR Checklist:

- [ ] Tests
  - [X] (not applicable?)
- [ ] Documentation
  - [X] (not applicable?)

#### PR label(s): 
  - [X] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20bugfix%20%F0%9F%95%B5%EF%B8%8F

### DIFF
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -644,7 +644,7 @@ async def healthchecks(app: procrastinate.App):
     print("Found procrastinate_jobs table: OK")
 
 
-async def shell_(app: procrastinate.App, args: list[str]):
+async def shell_(app: procrastinate.App, args: list[str] = []):
     """
     Administration shell for procrastinate.
     """
--- a/procrastinate/cli.py
+++ b/procrastinate/cli.py
@@ -644,7 +644,7 @@ async def healthchecks(app: procrastinate.App):
     print("Found procrastinate_jobs table: OK")
 
 
-async def shell_(app: procrastinate.App, args: list[str] = []):
+async def shell_(app: procrastinate.App, args: list[str] | None = None):
     """
     Administration shell for procrastinate.
     """


ducdetronquito:
Hey @linspw !

I'm not a maintainer of procrastinate, but I would like to just give some feedbacks on your PR.

You should avoid to use mutable default argument in Python because they are persisted accross calls to the function and therefore can lead to very hard to debug bugs if this argument is mutated inside the function

An example:
```python
def do_something(args: list[str] = []) -> None:
    args.append("Hello")
    print(args)
    
    
do_something()
# Output: ["Hello"]

do_something()
# Output: ["Hello", "Hello]

# Etc....
```

I would recommend to use `None` as a replacement to notify that no args are provided by the caller:

```python
async def shell_(app: procrastinate.App, args: list[str] | None = None):
    args = args or []
    # Continue with the rest of the function
```

Have a nice day :)

medihack:
@linspw Thanks a lot for your contribution. Even when we don't mutate args currently, I think @ducdetronquito is right (one never knows what the future brings 😉). How about just changing the function definition to `args: list[str] | None = None`? (I don't think we need the `args = args or []` in this case).

ewjoachim:
Hm, upon closer inspection, you're treating a symptom, not the cause.

The issue is that `args` seems to not be passed all the way to the command, but only when used through Django's `manage.py procrastinate`, but even with a default value, `./manage.py procrasinate shell list_tasks` won't work. We need to understand and solve **that** issue, I think.

Note: It's most certainly linked to #1038 which added `args` to `procrastinate shell`.

ewjoachim:
Here's the reason: [Django is removing](https://github.com/django/django/blob/main/django/core/management/base.py#L410) parameters named "args" for compatibility with optparse. If our option was named anything else than `args`, it would work.

My suggestion for a fix is twofold:
- Rename `args` to `shell_command`
- add a test that will fail with an explanation if we even create an `args` param in the future.

ewjoachim:
(suggested an aleternative implementation, as outlined above, in #1121 )

linspw:
Hello guys, thank you very much for your comments and review.
Sorry for my disappearance, I've had a busy few weeks lately
But I agree with everything you said, from the implementation suggestion (@ducdetronquito  and @medihack) to @ewjoachim  point.
I'll adjust it but I believe that @ewjoachim 's implementation is more correct

PR #1102, opened 2024-07-06
app.replace_connector to ease tests

### LINKED ISSUE #1099: Django testing documentation outdated
It seems that the [documentation about how to test Procrastinate](https://procrastinate.readthedocs.io/en/stable/howto/django/tests.html) in Django is wrong (or outdated). For example, there is no `defer` method of `app` (`app.defer("my_task", args=(1, 2))`). I guess it should be something like `app.configure_task("my_task", args=(1, 2)).defer()`.
There is also some problem when using the in_memory_app fixture as described in the documentation. When deferring a task like `in_memory_app.configure_task("example_project.example_app.tasks.example_task").defer()` the app still tries to use the DjangoConnector. So defer doesn't respect the InMemoryConnector, even when it is correctly set. The problem is that the Task uses the [`job manager` of the blueprint](https://github.com/procrastinate-org/procrastinate/blob/59a3f7c5b260c83d1967dc09028954b9ce5b7c77/procrastinate/tasks.py#L221), which is set [while it is registered](https://github.com/procrastinate-org/procrastinate/blob/59a3f7c5b260c83d1967dc09028954b9ce5b7c77/procrastinate/blueprints.py#L320) (so before the `in_memory_app` is created). We have to figure out a way for the task to respect the `current_app` and its `job_manager` or to somehow overwrite the `job_manager` (or the blueprint) of the task itself.

ewjoachim:
Closes #1099 

@medihack This is  PR where I add `app.replace_connector` that should be the answer to the issue you raised. If you can check that it answers your issue.

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/docs/howto/django/tests.md
+++ b/docs/howto/django/tests.md
@@ -14,18 +14,17 @@ from procrastinate.contrib.django import procrastinate_app
 
 @pytest.fixture
 def in_memory_app(monkeypatch):
-    app = procrastinate_app.current_app.with_connector(
-        testing.InMemoryConnector()
-    )
-    monkeypatch.setattr(procrastinate_app, "current_app", app)
-    return app
+    in_memory = testing.InMemoryConnector()
+    with procrastinate_app.current_app.replace_connector(in_memory) as app:
+        monkeypatch.setattr(procrastinate_app, "current_app", app)
+        yield app
 ```
 
 :::{note}
 `procrastinate.contrib.django.procrastinate_app.current_app` is not exactly
 _documented_ but whatever instance this variable points to is what
 will be returned as `procrastinate.contrib.django.app`. It's probably a bad
-idea to use this outside of tests.
+idea to manipulate this outside of tests.
 :::
 
 ## Integration tests
@@ -44,7 +43,7 @@ from procrastinate.contrib.django.models import ProcrastinateJob
 
 def test_my_task(app):
     # Run the task
-    app.defer("my_task", args=(1, 2))
+    task.defer(a=1, b=2)
 
     # Check the job has been created
     assert ProcrastinateJob.objects.filter(task_name="my_task").count() == 1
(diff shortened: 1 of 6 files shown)


medihack:
Thanks a lot, this works fine! There are still some other occurrences of `app.defer(...)` in the [documentation](https://procrastinate.readthedocs.io/en/stable/howto/django/tests.html) (that's why I thought it may be outdated). I guess those must be replaced by `task.defer(...)` as well. Maybe we can also add some "pseudo" import like `from mypackage.procrastinate import task` to make it more clear where the `task` object comes from (and maybe even name it `my_task`).

docs/howto/django/tests.md
```diff
@@ -44,7 +43,7 @@ from procrastinate.contrib.django.models import ProcrastinateJob
 
 def test_my_task(app):
```
medihack, line 44: What is `app` here? Some pytest fixture I am not aware of? Or should it just be `from procrastinate.contrib.django import app`?

docs/howto/django/tests.md
```diff
@@ -44,7 +43,7 @@ from procrastinate.contrib.django.models import ProcrastinateJob
...
     # Run the task
-    app.defer("my_task", args=(1, 2))
+    task.defer(a=1, b=2)
```
medihack, line 46: Maybe we should name it `my_task` and add a pseudo import at the top (e.g. `from mypackage.procrastinate import my_task` to make it clearer that this is the task defined by the user.

ewjoachim:
I've added a commit that should hopefully answer your comments :)

medihack:
Cool, thanks so much. Looks more consistent now. Let's merge and publish 🙂.

PR #1113, opened 2024-07-15
Allow to change the priority of a to be retried job

### LINKED ISSUE #1096: Allow to change the priority when retrying
A nice additional feature would be to change the priority when retrying a job. For example, I would like to slightly increase the priority when a job fails and will be retried. Maybe we could use an additional parameter like `RetryStrategy(change_priority=2)` to increase it by 2.

medihack:
Closes #1096 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

As discussed in #1096 this PR allows to change the priority of a to be retried job by implementing a new API (`get_retry_decision` and `RetryDecision`). This PR is hopefully fully backward compatible, but I had to change the method signature of `BaseRetryStrategy.get_retry_exception` which I assume is internal API.

### DIFF
--- a/docs/howto/advanced/retry.md
+++ b/docs/howto/advanced/retry.md
@@ -64,9 +64,34 @@ between retries:
 ## Implementing your own strategy
 
 - If you want to go for a fully fledged custom retry strategy, you can implement your
-  own retry strategy (though we recommend always keeping a max_retry):
+  own retry strategy. This also allows to (optionally) change the priority of the job.
 
+  ```python
+  import random
+  from procrastinate import Job, RetryDecision
+
+  class RandomRetryStrategy(procrastinate.BaseRetryStrategy):
+      max_attempts = 3
+      min = 1
+      max = 10
+
+      def get_retry_decision(self, *, exception:Exception, job:Job) -> RetryDecision:
+          if job.attempts >= max_attempts:
+              return RetryDecision(should_retry=False)
+
+          wait = random.uniform(self.min, self.max)
+
+          return RetryDecision(
+              should_retry=True,
+              schedule_in=wait,
+              new_priority=job.priority + 1,
+          )
   ```
+
+- There is is also a legacy `get_schedule_in` method that is depreciated an will be
+  removed in a future version in favor of the above `get_retry_decision` method.
+
+  ```python
   import random
 
   class RandomRetryStrategy(procrastinate.BaseRetryStrategy):
@@ -74,12 +99,9 @@ between retries:
       min = 1
       max = 10
 
-      def get_schedule_in(self, *, exception:Exception, attempts: int, **kwargs) -> int:
+      def get_schedule_in(self, *, exception:Exception, attempts: int) -> int:
           if attempts >= max_attempts:
               return None
 
           return random.uniform(self.min, self.max)
   ```
-
-It's interesting to add a catch-all parameter `**kwargs` to make your strategy more
-resilient to possible changes of Procrastinate in the future.
--- a/docs/howto/advanced/retry.md
+++ b/docs/howto/advanced/retry.md
@@ -88,7 +88,7 @@ between retries:
           )
   ```
 
-- There is is also a legacy `get_schedule_in` method that is depreciated an will be
+- There is also a legacy `get_schedule_in` method that is deprecated an will be
   removed in a future version in favor of the above `get_retry_decision` method.
 
   ```python
--- a/docs/howto/advanced/retry.md
+++ b/docs/howto/advanced/retry.md
@@ -63,12 +63,15 @@ between retries:
 
 ## Implementing your own strategy
 
-- If you want to go for a fully fledged custom retry strategy, you can implement your
-  own retry strategy by returning a `RetryDecision` object from the
-  `get_retry_decision` method. This also allows to (optionally) change the priority,
-  the queue or the lock of the job. The time to wait between retries can be specified
-  with `retry_in` or alternatively with `retry_at`. If `None` is returned
-  from `get_retry_decision` then the job will not be retried.
+If you want to go for a fully fledged custom retry strategy, you can implement your
+own retry strategy by returning a `RetryDecision` object from the
+`get_retry_decision` method. This also allows to (optionally) change the priority,
+the queue or the lock of the job. If `None` is returned from `get_retry_decision`
+then the job will not be retried.
+
+The time to wait between retries can be specified with `retry_in` or alternatively
+with `retry_at`. This is similar to how `schedule_in` and `schedule_at` are used
+when {doc}`scheduling a job in the future <schedule>`.
 
   ```python
   import random
@@ -86,27 +89,12 @@ between retries:
           wait = random.uniform(self.min, self.max)
 
           return RetryDecision(
-              retry_in={"seconds": 10}, # or retry_at (a datetime object)
+              retry_in={"seconds": wait}, # or retry_at (a datetime object)
               priority=job.priority + 1, # optional
               queue="another_queue", # optional
               lock="another_lock", # optional
           )
   ```
 
-- There is also a legacy `get_schedule_in` method that is deprecated an will be
-  removed in a future version in favor of the above `get_retry_decision` method.
-
-  ```python
-  import random
-
-  class RandomRetryStrategy(procrastinate.BaseRetryStrategy):
-      max_attempts = 3
-      min = 1
-      max = 10
-
-      def get_schedule_in(self, *, exception:Exception, attempts: int) -> int:
-          if attempts >= max_attempts:
-              return None
-
-          return random.uniform(self.min, self.max)
-  ```
+There is also a legacy `get_schedule_in` method that is deprecated an will be
+removed in a future version in favor of the above `get_retry_decision` method.
--- a/docs/howto/advanced/retry.md
+++ b/docs/howto/advanced/retry.md
@@ -11,7 +11,7 @@ app / machine reboots.
 
 - Retry 5 times (so 6 attempts total):
 
-  ```
+  ```python
   @app.task(retry=5)
   def flaky_task():
       if random.random() > 0.9:
@@ -21,7 +21,7 @@ app / machine reboots.
 
 - Retry indefinitely:
 
-  ```
+  ```python
   @app.task(retry=True)
   def flaky_task():
       if random.random() > 0.9:
@@ -42,7 +42,7 @@ Advanced strategies let you:
 
 Define your precise strategy using a {py:class}`RetryStrategy` instance:
 
-```
+```python
 from procrastinate import RetryStrategy
 
 @app.task(retry=procrastinate.RetryStrategy(
@@ -64,7 +64,11 @@ between retries:
 ## Implementing your own strategy
 
 - If you want to go for a fully fledged custom retry strategy, you can implement your
-  own retry strategy. This also allows to (optionally) change the priority of the job.
+  own retry strategy by returning a `RetryDecision` object from the
+  `get_retry_decision` method. This also allows to (optionally) change the priority,
+  the queue or the lock of the job. The time to wait between retries can be specified
+  with `retry_in` or alternatively with `retry_at`. If `None` is returned
+  from `get_retry_decision` then the job will not be retried.
 
   ```python
   import random
@@ -82,9 +86,10 @@ between retries:
           wait = random.uniform(self.min, self.max)
 
           return RetryDecision(
-              should_retry=True,
-              schedule_in=wait,
-              new_priority=job.priority + 1,
+              retry_in={"seconds": 10}, # or retry_at (a datetime object)
+              priority=job.priority + 1, # optional
+              queue="another_queue", # optional
+              lock="another_lock", # optional
           )
   ```
 
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import datetime
+from procrastinate.retry import RetryDecision
 
 
 class ProcrastinateException(Exception):
@@ -45,11 +45,8 @@ class JobRetry(ProcrastinateException):
     Job should be retried.
     """
 
-    def __init__(
-        self, scheduled_at: datetime.datetime, new_priority: int | None = None
-    ):
-        self.scheduled_at = scheduled_at
-        self.new_priority = new_priority
+    def __init__(self, retry_decision: RetryDecision):
+        self.retry_decision = retry_decision
         super().__init__()
 
 
--- a/procrastinate/exceptions.py
+++ b/procrastinate/exceptions.py
@@ -45,8 +45,11 @@ class JobRetry(ProcrastinateException):
     Job should be retried.
     """
 
-    def __init__(self, scheduled_at: datetime.datetime):
+    def __init__(
+        self, scheduled_at: datetime.datetime, new_priority: int | None = None
+    ):
         self.scheduled_at = scheduled_at
+        self.new_priority = new_priority
         super().__init__()
 
 
--- a/procrastinate/retry.py
+++ b/procrastinate/retry.py
@@ -48,6 +48,28 @@ class RetryDecision:
         queue: str | None = None,
         lock: str | None = None,
     ) -> None:
+        """
+        Specifies when and how a job should be retried.
+
+        Parameters
+        ----------
+        retry_at : ``Optional[datetime.datetime]``
+            If set at present time or in the past, the job may be retried immediately.
+            Otherwise, the job will be retried no sooner than this date & time.
+            Should be timezone-aware (even if UTC). Defaults to present time.
+        retry_in : ``Optional[types.TimeDeltaParams]``
+            If set, the job will be retried after this duration. If not set, the job will
+            be retried immediately.
+        priority : ``Optional[int]``
+            If set, the job will be retried with this priority. If not set, the priority
+            remains unchanged.
+        queue : ``Optional[int]``
+            If set, the job will be retried on this queue. If not set, the queue remains
+            unchanged.
+        lock : ``Optional[int]``
+            If set, the job will be retried with this lock. If not set, the lock remains
+            unchanged.
+        """
         if retry_at and retry_in is not None:
             raise ValueError("Cannot set both retry_at and retry_in")
 
@@ -127,14 +149,6 @@ class BaseRetryStrategy:
             The exception raised by the job
         job:
             The current job
-
-        Returns
-        -------
-        ``RetryDecision``
-            A RetryDecision object with the following attributes:
-            - should_retry: a boolean indicating whether the job should be retried
-            - schedule_in: an optional float indicating the number of seconds to wait
-            - priority: an optional integer indicating the new priority of the job
         """
         raise NotImplementedError("Missing implementation of 'get_retry_decision'.")
 
--- a/procrastinate/retry.py
+++ b/procrastinate/retry.py
@@ -6,11 +6,20 @@ try again? And when?
 from __future__ import annotations
 
 import datetime
+import warnings
 from typing import Iterable, Union
 
 import attr
 
 from procrastinate import exceptions, utils
+from procrastinate.jobs import Job
+
+
+@attr.dataclass
+class RetryDecision:
+    should_retry: bool
+    schedule_in: float | None = None
+    new_priority: int | None = None
 
 
 class BaseRetryStrategy:
@@ -20,14 +29,35 @@ class BaseRetryStrategy:
     """
 
     def get_retry_exception(
-        self, exception: BaseException, attempts: int
+        self, exception: BaseException, job: Job
     ) -> exceptions.JobRetry | None:
-        schedule_in = self.get_schedule_in(exception=exception, attempts=attempts)
-        if schedule_in is None:
-            return None
+        try:
+            retry_decision = self.get_retry_decision(exception=exception, job=job)
+            if not retry_decision.should_retry:
+                return None
 
-        schedule_at = utils.utcnow() + datetime.timedelta(seconds=schedule_in)
-        return exceptions.JobRetry(schedule_at.replace(microsecond=0))
+            schedule_at = utils.utcnow() + datetime.timedelta(
+                seconds=retry_decision.schedule_in
+                if retry_decision.schedule_in is not None
+                else 0
+            )
+            return exceptions.JobRetry(
+                schedule_at.replace(microsecond=0), retry_decision.new_priority
+            )
+        except NotImplementedError:
+            warnings.warn(
+                "`get_schedule_in` is deprecated, use `get_retry_decision` instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            schedule_in = self.get_schedule_in(
+                exception=exception, attempts=job.attempts
+            )
+            if schedule_in is None:
+                return None
+
+            schedule_at = utils.utcnow() + datetime.timedelta(seconds=schedule_in)
+            return exceptions.JobRetry(schedule_at.replace(microsecond=0))
 
     def get_schedule_in(self, *, exception: BaseException, attempts: int) -> int | None:
         """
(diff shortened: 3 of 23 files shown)


medihack:
I will add a test to bring the coverage of `retry.py` back to 100% by also testing the fallback.

docs/howto/advanced/retry.md
```diff
@@ -64,22 +64,44 @@ between retries:
...
+      def get_retry_decision(self, *, exception:Exception, job:Job) -> RetryDecision:
+          if job.attempts >= max_attempts:
+              return RetryDecision(should_retry=False)
+
+          wait = random.uniform(self.min, self.max)
+
+          return RetryDecision(
+              should_retry=True,
+              schedule_in=wait,
+              new_priority=job.priority + 1,
```
ewjoachim, line 87: Priority is a very specific thing we change here, does it make sense to change other parameters ? If it's really just priority, then fine, otherwise, should this accept `**kwargs` passed to configure the new job ? I think it might make sense to change the queue (e.g you may have task running on small machines and want retries to run on a smaller number of bigger machines)
I don't imagine a usecase for changing queueing lock (I'm not sure I even see when one would use retry and queueing lock)
Changing lock, I don't know, I guess but not sure how to articulate.
Changing schedule in/at: we already do that but what we're doing could be expressed as a special case of changing schedule_at. Also, it might make sense to change decide to retry tomorrow, but in a calendar way and not in 24h (or "retry at night when there's less activity")
Changing the task or the args: really depends on the business case. We could imagine that on certain exception, we'd retry and add a param saying it's safe to skip part of the task.

So all in all: it looks it might make sense to override a large proportion of the job attributes, it seems. What do you think ?

procrastinate/retry.py
```diff
@@ -97,6 +154,16 @@ def get_schedule_in(self, *, exception: BaseException, attempts: int) -> int | N
...
 
+    def get_retry_decision(
+        self, *, exception: BaseException, job: Job
+    ) -> RetryDecision:
+        schedule_in = self.get_schedule_in(exception=exception, attempts=job.attempts)
+
+        if schedule_in is None:
+            return RetryDecision(should_retry=False)
+
+        return RetryDecision(should_retry=True, schedule_in=schedule_in)
```
ewjoachim, line 165:
```suggestion
        return RetryDecision(should_retry=schedule_in is not None, schedule_in=schedule_in)
```
(If too ugly, store `should_retry` in an intermediate variable)

procrastinate/exceptions.py
```diff
@@ -45,8 +45,11 @@ class JobRetry(ProcrastinateException):
     Job should be retried.
     """
 
-    def __init__(self, scheduled_at: datetime.datetime):
+    def __init__(
```
ewjoachim, line 48: I think it might make more sense to attach the retry decision to the exception, even if the should_retry will always be True.

ewjoachim (commented):
First round of review, with big questions :) I'll make a more detailed round later (when we'll have agreed on the specifics)

This is, again, an excellent job you've done. Thank you very much!

docs/howto/advanced/retry.md
ewjoachim: (Also: sorry to come up with the idea so late in the dev phase. Had I suggested it earlier, that might have been more respectful of your time, so I want to apologize for this)

medihack: > Priority is a very specific thing we change here, does it make sense to change other parameters ? If it's really just priority, then fine, otherwise, should this accept `**kwargs` passed to configure the new job ? I think it might make sense to change the queue (e.g you may have task running on small machines and want retries to run on a smaller number of bigger machines) I don't imagine a usecase for changing queueing lock (I'm not sure I even see when one would use retry and queueing lock) Changing lock, I don't know, I guess but not sure how to articulate. 

Yes, sounds like a reasonable suggestion. But I'm not quite convinced to use `**kwargs` yet (I wouldn't even know how to map it to the SQL query). I think I would just add the additional attributes that make sense to the `RetryDecision` (somehow a clearer API, in my opinion). Or what API do you exactly have in mind?

> Changing schedule in/at: we already do that but what we're doing could be expressed as a special case of changing schedule_at. Also, it might make sense to change decide to retry tomorrow, but in a calendar way and not in 24h (or "retry at night when there's less activity") 

Isn't this all possible with `schedule_in`? The library user could take the current date/time and calculate the delta he wants (whether it is in 10 minutes, tomorrow, or ten years). I would keep it simple, but open to other suggestions.

> Changing the task or the args: really depends on the business case. We could imagine that on certain exception, we'd retry and add a param saying it's safe to skip part of the task.

Changing the task sounds like cool and advanced stuff, but on the other hand, it can also be a way to shoot yourself in the foot 😉. When messing up the args, for example, the task would crash instantly. I am also not sure this is really something that library users would do (and I never saw this in a job pipeline before, but that doesn't mean anything). 

And another small detail. I am also not super convinced of the `should_retry` attribute. Why not just return None when no retry should take place?

> (Also: sorry to come up with the idea so late in the dev phase. Had I suggested it earlier, that might have been more respectful of your time, so I want to apologize for this)

No worries. Sometimes, you have to have things in front of you to come up with new ideas.

ewjoachim: Agree with all your points :)

medihack: How about ...
```python
@attr.dataclass
class RetryDecision:
    schedule_in: float | None = None
    priority: int | None = None
    queue: str | None = None
    lock: str | None = None
```
I would only use the `new_` prefix  (for example `new_priority`) in the SQL query or do you like that naming in the `RetryDecision` with the `new_` prefix?
Regarding changing the `args`, I would prefer to add something like a `meta` field where additional information for that job could be stored. But I wouldn't do that on this issue; I would do it more when there is really a need for that.

procrastinate/retry.py
medihack: As discussed above, I will remove `should_retry` and allow `get_retry_decision` to return `None` (which means no retry then). Ok?

procrastinate/exceptions.py
medihack: Wouldn't that mean we could break something? Isn't it public API (it has some docstring but is, on the other hand, not in the API reference)? We also would have to calculate `schedule_at` somewhere else (`RetryDecision` uses `schedule_in`).

procrastinate/retry.py
ewjoachim: Yep !

procrastinate/exceptions.py
ewjoachim: I don't think it's exposed anywhere. Do you have an idea how a user might manipulate a `JobRetry`? I could be missing something (honnestly, I really could) but I don't _think_ it's exposed at the moment.

> We also would have to calculate schedule_at somewhere else (RetryDecision uses schedule_in).

Hm, I haven't dived into the question enough to see exactly what you have in mind, but... `schedule_in` and `schedule_at` are often slightly annoying in the API, but it's meant to be as helpful as possible for the user. `schedule_at` is the source of truth, `schedule_in` is the syntactic sugar. I guess this points towards the fact `RetryDecision` should probably have a `schedule_at` attribute, and the constructor should accept both. Probably worth factoring [this code](https://github.com/procrastinate-org/procrastinate/blob/f349c5be9044efb7276a6fa375f2e05ad2868150/procrastinate/tasks.py#L46-L55) out to a function and use it in both places, WDYT?

medihack: > I don't think it's exposed anywhere. Do you have an idea how a user might manipulate a `JobRetry`? I could be missing something (honnestly, I really could) but I don't _think_ it's exposed at the moment.

What does finally end up in the Reference API? I thought everything with a docstring, but that doesn't seem to be the case.

> 
> > We also would have to calculate schedule_at somewhere else (RetryDecision uses schedule_in).
> 
> Hm, I haven't dived into the question enough to see exactly what you have in mind, but... `schedule_in` and `schedule_at` are often slightly annoying in the API, but it's meant to be as helpful as possible for the user. `schedule_at` is the source of truth, `schedule_in` is the syntactic sugar. I guess this points towards the fact `RetryDecision` should probably have a `schedule_at` attribute, and the constructor should accept both. 

From a user perspective, I find `schedule_in` more convenient because most of the time, I just want to do something like "just retry in 10 minutes", regardless of the current date or time (and let the library figure it out). In the rare cases where I would like to specify a specific date/time, I can still do that with a bit of calculation.
I am not a big fan of having mutually exclusive options like `schedule_in` and `schedule_at`, especially as one can be quite easily calculated from the other. If it's ok for you I would stay with `schedule_in`, but would rename it to `retry_in` in the `RetryDecision`. I then would move the logic to conversion into `retry_at` from [`retry.py`](https://github.com/procrastinate-org/procrastinate/blob/234f5f0ef3571cd9ff9053ca174c98e18d0f6631/procrastinate/retry.py#L29) to somewhere else (as we only store the `RetryDecision` in the `JobRetry`).
Is this ok for you, or do you really want to have both `retry_in` and `retry_at` in `RetryDescision`. I can live with that, too 😂.

> Probably worth factoring [this code](https://github.com/procrastinate-org/procrastinate/blob/f349c5be9044efb7276a6fa375f2e05ad2868150/procrastinate/tasks.py#L46-L55) out to a function and use it in both places, WDYT?

I see task scheduling and retries as two separate features even though they both ultimately use the same underlying logic (the `schedule_at` column in Postgres). At least I wouldn't refactor that in this issue. I also think that the API there is okay and not worth breaking.

medihack: Hmm, I just found out that `schedule_in` of the task is more advanced than I thought (especially regarding `TimeDeltaParams`). Now, I am a bit insecure about what the best decision would be if we really want to make both APIs similar. Then maybe we should indeed allow `retry_in` and `retry_at` in the `RetryDecision` (even when both are mutually exclusive, which I kind of dislike a bit). You decide 😉.

ewjoachim: > What does finally end up in the Reference API? I thought everything with a docstring, but that doesn't seem to be the case.

I try to put a docstring in anything public, but the opposite isn't true. Anything people should need to import (including in typing) should be `procrastinate.__all__` (in `__init__.py`) and should be documented in `reference.rst` (thus the docstrings). This is all manual, I try to maintain a mental model of what's officially exposed or not but I don't think such a thing appears explicitly in the code apart from what I mentioned. It could be interesting to make hard rules.

> [...retry_in, retry_at]

Of course for the naming, go ahead. Regarding the mutually exclusive: do you know that thanks to `typing.overload`, you can make it explicit that the 2 params can't be used together (for the type checker)? (also... Maybe this ship has sailed, but I should probably have done a single param that receives either a `datetime` or a `timedelta`).

> If it's ok for you I would stay with schedule_in, but would rename it to retry_in in the RetryDecision. I then would move the logic to conversion into retry_at from [retry.py](https://github.com/procrastinate-org/procrastinate/blob/234f5f0ef3571cd9ff9053ca174c98e18d0f6631/procrastinate/retry.py#L29) to somewhere else (as we only store the RetryDecision in the JobRetry).

Could also be in the `RetryDecision` constructor? But elsewhere works too (in that case, `RetryDecision` will keep 2 attributes). I think it's not very important, but I'd love to only have this logic coded once and reused. Datetimes are hard enough, if we can avoid consistency issues :D

> You decide 😉.

You drive a hard bargain. Ok, if I'm in charge, let's say we keep both, for the sake of API consistency throughout Procrastinate. If we can do it when we configure a task, we can do it at retry time too. Might be overkill but at least no one will wonder why it's available on one side and not on the other. But let me be clear: it's a subjective decision. I don't foresee this decision making us sad in the future (compared to a _bunch_ of past decisons making me feel sad in the present :D )

medihack: > > What does finally end up in the Reference API? I thought everything with a docstring, but that doesn't seem to be the case.
> 
> I try to put a docstring in anything public, but the opposite isn't true. Anything people should need to import (including in typing) should be `procrastinate.__all__` (in `__init__.py`) and should be documented in `reference.rst` (thus the docstrings). This is all manual, I try to maintain a mental model of what's officially exposed or not but I don't think such a thing appears explicitly in the code apart from what I mentioned. It could be interesting to make hard rules.

Thanks, good to know.

> 
> > [...retry_in, retry_at]
> 
> Of course for the naming, go ahead. Regarding the mutually exclusive: do you know that thanks to `typing.overload`, you can make it explicit that the 2 params can't be used together (for the type checker)? (also... Maybe this ship has sailed, but I should probably have done a single param that receives either a `datetime` or a `timedelta`).
> 
> > If it's ok for you I would stay with schedule_in, but would rename it to retry_in in the RetryDecision. I then would move the logic to conversion into retry_at from [retry.py](https://github.com/procrastinate-org/procrastinate/blob/234f5f0ef3571cd9ff9053ca174c98e18d0f6631/procrastinate/retry.py#L29) to somewhere else (as we only store the RetryDecision in the JobRetry).
> 
> Could also be in the `RetryDecision` constructor? But elsewhere works too (in that case, `RetryDecision` will keep 2 attributes). I think it's not very important, but I'd love to only have this logic coded once and reused. Datetimes are hard enough, if we can avoid consistency issues :D

Yes, doing this in the constructor and keeping only `retry_at` sounds nice. Thereby, I could kill two birds with one stone by also using `typing.overload` on the constructor.

> > You decide 😉.
> 
> You drive a hard bargain. Ok, if I'm in charge, let's say we keep both, for the sake of API consistency throughout Procrastinate. If we can do it when we configure a task, we can do it at retry time too. Might be overkill but at least no one will wonder why it's available on one side and not on the other. But let me be clear: it's a subjective decision. I don't foresee this decision making us sad in the future (compared to a _bunch_ of past decisons making me feel sad in the present :D )

I already edited my post, and it's nice to see that we came to the same conclusion 😉

ewjoachim: > I already edited my post, and it's nice to see that we came to the same conclusion 😉

Very nice indeed. Have I mentioned I'm thrilled to have you on board ? Because I am 😀

medihack:
Okay, I hope I have thought of everything now. What do you think of the documentation? Is there something I could improve?

docs/howto/advanced/retry.md
```diff
@@ -64,22 +64,49 @@ between retries:
 ## Implementing your own strategy
 
 - If you want to go for a fully fledged custom retry strategy, you can implement your
-  own retry strategy (though we recommend always keeping a max_retry):
+  own retry strategy by returning a `RetryDecision` object from the
+  `get_retry_decision` method. This also allows to (optionally) change the priority,
+  the queue or the lock of the job. The time to wait between retries can be specified
+  with `retry_in` or alternatively with `retry_at`. If `None` is returned
```
ewjoachim, line 70: Let's mention that they have se same values as `task.configure(schedule_in=..., schedule_at=...)`.

docs/howto/advanced/retry.md
```diff
@@ -64,22 +64,49 @@ between retries:
...
+      max = 10
+
+      def get_retry_decision(self, *, exception:Exception, job:Job) -> RetryDecision:
+          if job.attempts >= max_attempts:
+              return RetryDecision(should_retry=False)
+
+          wait = random.uniform(self.min, self.max)
+
+          return RetryDecision(
+              retry_in={"seconds": 10}, # or retry_at (a datetime object)
```
ewjoachim, line 89: It's a nice `wait` variable you have here. A shame it's unused :D

docs/howto/advanced/retry.md
```diff
@@ -64,22 +64,49 @@ between retries:
...
+              priority=job.priority + 1, # optional
+              queue="another_queue", # optional
+              lock="another_lock", # optional
+          )
   ```
+
+- There is also a legacy `get_schedule_in` method that is deprecated an will be
```
ewjoachim, line 96:
```suggestion
- There is also a deprecated `get_schedule_in` method, which will be
```

docs/howto/advanced/retry.md
ewjoachim: (also, let's remove the unordered list. Those can be 2 paragraphs.

docs/howto/advanced/retry.md
ewjoachim: Actually... Let's mention `get_schedule_in` but not the example. I don't think people need example of how to implement deprecated methods.

procrastinate/retry.py
```diff
@@ -6,28 +6,88 @@
...
+            if retry_decision is None:
+                return None
+
+            return exceptions.JobRetry(retry_decision=retry_decision)
+        except NotImplementedError:
+            warnings.warn(
+                "`get_schedule_in` is deprecated, use `get_retry_decision` instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
ewjoachim, line 82: We should probably only do that if `self.get_schedule_in` is actually implemented. If someone just forgot to implement any method on the class, this error message will be confusing. Also, in that case, we want to raise that `self.get_retry_decision` was not found, not that `self.get_schedule_in` was not found.

procrastinate/retry.py
```diff
@@ -85,6 +171,7 @@ class RetryStrategy(BaseRetryStrategy):
     retry_exceptions: Iterable[type[Exception]] | None = None
 
     def get_schedule_in(self, *, exception: BaseException, attempts: int) -> int | None:
+        # TODO: remove this method in a future version and move logic to `get_retry_decision`
```
ewjoachim, line 174: Is there any benefit in not doing it now ? Users aren't going to call get_schedule_in themselves, are they ?

procrastinate/tasks.py
```diff
@@ -51,7 +41,7 @@ def configure_task(
         raise ValueError("Cannot set both schedule_at and schedule_in")
 
     if schedule_in is not None:
-        schedule_at = utils.utcnow() + datetime.timedelta(**schedule_in)
+        schedule_at = utils.datetime_from_timedelta_params(schedule_in)
```
ewjoachim, line 44: 👍

procrastinate/utils.py
```diff
@@ -432,3 +434,17 @@ def async_context_decorator(func: Callable) -> Callable:
...
+
+
+class TimeDeltaParams(TypedDict):
+    weeks: NotRequired[int]
+    days: NotRequired[int]
+    hours: NotRequired[int]
+    minutes: NotRequired[int]
+    seconds: NotRequired[int]
+    milliseconds: NotRequired[int]
+    microseconds: NotRequired[int]
```
ewjoachim, line 446: (either this or `types.py` as we're using it in multiple places)

tests/integration/test_manager.py
```diff
@@ -302,6 +302,27 @@ async def test_retry_job(pg_job_manager, fetched_job_factory):
...
+        retry_at=datetime.datetime.now(datetime.timezone.utc),
+        priority=5,
+        queue="queue_b",
+        lock="some_lock",
+    )
+
+    job2 = await pg_job_manager.fetch_job(queues=None)
+
+    assert job2.id == job1.id
+    assert job2.attempts == job1.attempts + 1
```
ewjoachim, line 322: I think we could assume this is 1?

ewjoachim (commented):
A few small comments. Very nice job overall, thank you again!

docs/howto/advanced/retry.md
medihack: Uups 🧐.

docs/howto/advanced/retry.md
medihack: Makes sense.

procrastinate/retry.py
```diff
@@ -6,28 +6,88 @@
...
+        retry_in: utils.TimeDeltaParams | None = None,
+        priority: int | None = None,
+        queue: str | None = None,
+        lock: str | None = None,
+    ) -> None: ...
+    @overload
+    def __init__(
+        self,
+        retry_at: datetime.datetime | None = None,
+        retry_in: None = None,
```
ewjoachim, line 37: You can remove parameters entirely from overloads if it helps. Here I believe you may remove retry_at from the top one and retry_in from the bottom one

procrastinate/utils.py
medihack: Yes, types.py sounds like a better location.

procrastinate/retry.py
medihack: I thought so too, but then I ended up with the below type error (at least in VS Code with pyright extension):
```
Overloaded implementation is not consistent with signature of overload 1
  Type "(self: Self@RetryDecision, retry_at: datetime | None = None, retry_in: TimeDeltaParams | None = None, priority: int | None = None, queue: str | None = None, lock: str | None = None) -> None" is incompatible with type "(self: Self@RetryDecision, retry_in: TimeDeltaParams | None = None, priority: int | None = None, queue: str | None = None, lock: str | None = None) -> None"
    Parameter name mismatch: "retry_in" versus "retry_at"
    Parameter 2: type "TimeDeltaParams | None" is incompatible with type "datetime | None"
    Parameter name mismatch: "priority" versus "retry_in"
    Parameter 3: type "int | None" is incompatible with type "TimeDeltaParams | None"
    Parameter name mismatch: "queue" versus "priority"
    Parameter 4: type "str | None" is incompatible with type "int | None"
    Parameter name mismatch: "lock" versus "queue"Pyright[reportInconsistentOverload](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInconsistentOverload)
retry.py(25, 9): Overload signature is defined here
Overloaded implementation is not consistent with signature of overload 2
  Type "(self: Self@RetryDecision, retry_at: datetime | None = None, retry_in: TimeDeltaParams | None = None, priority: int | None = None, queue: str | None = None, lock: str | None = None) -> None" is incompatible with type "(self: Self@RetryDecision, retry_at: datetime | None = None, priority: int | None = None, queue: str | None = None, lock: str | None = None) -> None"
    Parameter name mismatch: "priority" versus "retry_in"
    Parameter 3: type "int | None" is incompatible with type "TimeDeltaParams | None"
    Parameter name mismatch: "queue" versus "priority"
    Parameter 4: type "str | None" is incompatible with type "int | None"
    Parameter name mismatch: "lock" versus "queue"Pyright[reportInconsistentOverload](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInconsistentOverload)
retry.py(33, 9): Overload signature is defined here
```

procrastinate/retry.py
medihack: Probably not. I removed it.

procrastinate/retry.py
```diff
@@ -6,28 +6,93 @@
...
+                return None
+
+            return exceptions.JobRetry(retry_decision=retry_decision)
+        except NotImplementedError as err:
+            try:
+                schedule_in = self.get_schedule_in(
+                    exception=exception, attempts=job.attempts
+                )
+            except NotImplementedError:
+                raise err
```
medihack, line 83: I wonder why coverage tells me that 82-83 is not tested. I would have expected that this is covered by `test_missing_implementation_in_custom_retry_strategy`, or is it because I re-raised the error there?

medihack: Never mind ... I tested the wrong method. Will fix it.

ewjoachim: `raise err from None` I believe, otherwise the traceback will be ugly, I think.

procrastinate/retry.py
ewjoachim: Can you try adding a `*` to make those kw only ?

medihack: Sure, I haven't thought of that. This works fine 😌.

procrastinate/retry.py
```diff
@@ -43,8 +109,34 @@ def get_schedule_in(self, *, exception: BaseException, attempts: int) -> int | N
...
+        job:
+            The current job
+
+        Returns
+        -------
+        ``RetryDecision``
+            A RetryDecision object with the following attributes:
+            - should_retry: a boolean indicating whether the job should be retried
+            - schedule_in: an optional float indicating the number of seconds to wait
+            - priority: an optional integer indicating the new priority of the job
```
ewjoachim, line 137: This is outdated. We should not list the attributes here but rather point to the corresponding doc.

ewjoachim (approved):
One last small change. Feel free to merge once fixed :)

procrastinate/retry.py
medihack: I'll just add the `RetryDecision` to the `reference.rst`. From looking at the reference documentation, it seems that when not providing a `Returns` in the docstring, it automatically links to the correct reference section by evaluating the return type hint.

medihack: Alright, I updated the API reference. Could you have another look at it? By the way, maybe we should add `TimeDeltaParams` directly to the reference. Currently, it is only mentioned as an aside in the [`Task.configure`](https://procrastinate.readthedocs.io/en/stable/reference.html#procrastinate.tasks.Task.configure) method. But I would do this in another PR, if you agree.

ewjoachim (approved):

ewjoachim:
Screenshot 2024-07-20 at 23 40 46
Seems awesome!

PR #1132, opened 2024-07-29
Fix retry when status of job is aborting

### LINKED ISSUE #1131: Retrying a task, that has been aborted
Hi guys!

I've been doing some manual QA for a project I'm working on and found some issues with the newly added `abort` logic for task cancellation. By the way, a neat feature I was waiting for, glad it has been added!

Here's the general context - I have a single type of task for the procrastinate, and I use a retry strategy over it to handle some expected exceptions. I also use a `context.should_abort_async` within the task to handle its cancellation. The issue appears with the following steps:

- a worker has started to process a task, marking its state as `doing`;
- in the middle of it, the task has been aborted from a different process, which changes the task state to `aborting`;
- after that, some expected error happens in the task itself (before the `should_abort_async` check), causing the worker to retry this task due to a retry strategy being used;
- during the retry `procrastinate_retry_job` SQL function will be used, which will raise  a `Job was not found or not in "doing" status` error;

This will shut down the worker process and will leave a task in the `procrastinate_jobs` table with an `aborting` status. If I just restart it, it won't also process newly added tasks for what it seems due to hanging an `aborting` task.

I understand that this might be a niche scenario, but I wonder if this behavior with the retry logic has been considered for a task abortion process. A quick and simple fix for this would be to mind the `aborting` status during job retry logic, but there is probably more to it.

Thanks for your attention!

medihack:
Closes #1131 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/procrastinate/contrib/django/migrations/0030_fix_retry_when_aborting.py
+++ b/procrastinate/contrib/django/migrations/0030_fix_retry_when_aborting.py
@@ -1,0 +1,15 @@
+from __future__ import annotations
+
+from django.db import migrations
+
+from .. import migrations_utils
+
+
+class Migration(migrations.Migration):
+    operations = [
+        migrations_utils.RunProcrastinateSQL(
+            name="02.09.01_01_fix_retry_when_aborting.sql"
+        ),
+    ]
+    name = "0030_fix_retry_when_aborting"
+    dependencies = [("procrastinate", "0029_add_additional_params_to_retry_job")]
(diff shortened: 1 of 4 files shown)


ewjoachim (commented):
Code is great but we need to decide whether we want to retry or not. I'm not convinced yet that it makes sense to retry.

Here is my rationale. The following is a timeline.

-> job is deferred

If the job is cancelled here, it is immediately discarded

-> job starts

If the job is cancelled here, it is just marked for abortion. It may not be aborted at all. _of course, given the job has started, it makes sense that it is treated differently_

-> job ends with an error, is enqueued for retrying (potentially with a delay)

If the job is cancelled here, it is immediately discarder

-> job starts again

What seems strange is that if the job is nearing its end (or error), if you wait for it to end with an error and **then** cancel it, it will be cancelled sooner that if you cancel it immediately. Similarily, if you cancel it again while it's enqueued for retrying, it will be discarded immediately. A process that requires the user to startegically wait before giving an order for it to be executed sooner or requires the user to request it multiple times seems like procrastinate wouldn't be trying to follow it, I think?

That said, I'm not 100% sure. At all.

PR #1133, opened 2024-07-29
Set job to failed when task raises and job is aborting (regardless retry)

### LINKED ISSUE #1131: Retrying a task, that has been aborted

medihack:
Closes #1131

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

This time, a job with the status `aborting` of a task that raises is set to `failed` (regardless of whether it should be retried or not). It feels a bit more consistent to me, but somehow, there is no optimal way. If we set a to-be-retried job in status `aborting` to `aborted` when the task raises, but one to `failed` that should not be retried, it seems somehow inconsistent to me.

@ewjoachim A note in the documentation is still missing. Just want to hear what you think.

### DIFF
--- a/tests/unit/test_worker.py
+++ b/tests/unit/test_worker.py
@@ -366,7 +366,7 @@ async def test_run_job_aborted(app, caplog):
     )
 
 
-async def test_run_job_error(app, caplog):
+async def test_run_job_error(app, caplog, mocker):
     caplog.set_level("INFO")
 
     def job_func(a, b):  # pylint: disable=unused-argument
@@ -385,6 +385,7 @@ async def test_run_job_error(app, caplog):
         task_name="job",
         queue="yay",
     )
+    app.job_manager.get_job_status_async = mocker.AsyncMock(return_value="doing")
     test_worker = worker.Worker(app, queues=["yay"])
     with pytest.raises(exceptions.JobError):
         await test_worker.run_job(job=job, worker_id=3)
@@ -401,7 +402,7 @@ async def test_run_job_error(app, caplog):
     )
 
 
-async def test_run_job_critical_error(app, caplog):
+async def test_run_job_critical_error(app, caplog, mocker):
     caplog.set_level("INFO")
 
     def job_func(a, b):  # pylint: disable=unused-argument
@@ -420,6 +421,7 @@ async def test_run_job_critical_error(app, caplog):
         task_name="job",
         queue="yay",
     )
+    app.job_manager.get_job_status_async = mocker.AsyncMock(return_value="doing")
     test_worker = worker.Worker(app, queues=["yay"])
     with pytest.raises(exceptions.JobError) as exc_info:
         await test_worker.run_job(job=job, worker_id=3)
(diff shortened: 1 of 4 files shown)


tests/unit/test_worker.py
```diff
@@ -385,6 +385,7 @@ def job_func(a, b):  # pylint: disable=unused-argument
         task_name="job",
         queue="yay",
     )
+    app.job_manager.get_job_status_async = mocker.AsyncMock(return_value="doing")
```
ewjoachim, line 388: Just curious: what happens if we don't mock? The InMemory implementation doesn't answer correctly ?

ewjoachim (approved):
That's great! I believe there may be a bit of conflict when we merge the worker refactor, though. 

Thank you so much for the time you spent and doing it all over again after my comment!

tests/unit/test_worker.py
medihack: It'll throw. The job is not put into the worker queue. The `run_job` method is tested in isolation.

medihack:
> That's great! I believe there may be a bit of conflict when we merge the worker refactor, though.

@onlyann When I commit this (later today), can you include it in the worker refactoring?

> Thank you so much for the time you spent and doing it all over again after my comment!

No prob, the comment was very relevant. And I am happy that we don't need a migration 🙂.

tests/unit/test_worker.py
ewjoachim: Ah, makes sense!

onlyann:
That's not a problem, I will resolve the conflict.

I think it's fine to merge this fix but wanted to highlight that there can still be a race condition where the job is changed to "aborting" between the check and the moment the job is set to retry.

Something we can look at improving if we end up removing that extra status in v3

medihack:
Alright, I will write a note in the documentation, merge it, and make a patch release.

ewjoachim:
:+1:

Issue #1084, opened 2024-06-25
Actively notfiy a job of an abort request


medihack:
With PR #1081, we added a feature to mark a currently running job for abortion, but we have to poll the database in the task ourselves to determine whether this mark was set (represented by the `aborting` status). As discussed in #511, it would be nice if the task was actively notified of an abortion (without the need for intense database polling) using a PostgreSQL listen-notify. We could then raise an `asyncio.CancelledError` in an async task. It's unclear what to do with a sync task (maybe a callback method like `context.listen_abort(cb: Callable`?!). It should be optional and triggered by something like `job_manager.cancel_job_by_id(33, abort=True, notify=True)`.
It would be best implemented after the worker refactoring #933, as it must be integrated into the worker class.

ewjoachim:
> It would be best implemented after the worker refactoring https://github.com/procrastinate-org/procrastinate/issues/933

Or be part of the design of #933

ewjoachim:
If we do #1083, is this worth the complexity ? I won't close this but I think it's interesting to spell out the gains we expect from this before tackling it.

medihack:
In some cases, it could simplify the abortion process and save the user some code (e.g., no need for another asyncio task that monitors the abort status). On the other hand, it's not a feature that needs to have high priority or be integrated right now as part of the worker refactoring (the refactoring is complicated enough anyway). It's more a nice to have feature and something I have not seen in any other distributed task queue library. So I would say, let us leave this issue open and see what the future brings. Regarding #1083, it's the next thing I'll deal with.

ewjoachim:
Closed as part of v3

PR #1139, opened 2024-07-31
Use additional abort field on jobs table for abortion requests

### LINKED ISSUE #1136: Reconsider the "aborting" status
It would be simpler if we had a dedicated field for an abortion request instead of an aborting status.

- The `aborting` state masks the `doing` state. State transitions from `aborting` are the same as from `doing`.
- Also, since the user does not necessarily have to follow this request with a dedicated field, it would still be preserved in a subsequent possible retry.
- Race conditions may occur that could solved with a dedicated field.

medihack:
Closes #1136

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 


onlyann:
Looking good. What about naming the column `abort_requested`? 

Alternatively, it could be a nullable timestamp `abort_requested_at` that can also be inferred as a flag but also conveys the time it is requested in the job table.

medihack:
> Looking good. What about naming the column `abort_requested`?

Yes, that sounds good. I renamed it to `abort_requested`.
> 
> Alternatively, it could be a nullable timestamp `abort_requested_at` that can be inferred as a flag but also conveys the time it is requested in the job table.

For the sake of simplicity, I would keep it a boolean flag. I make sure that the event of an abortion request is recorded (and the event itself has a timestamp). Are you okay with that?

medihack:
Some design decisions:

- I haven't intentionally made `abort_requested` available on the [`Job`](https://github.com/openradx/procrastinate/blob/463511ce7a39f668b6aea03c2b62497e2b24711f/procrastinate/jobs.py#L46) itself because if the user checks `job.abort_requested` of a passed context inside a task (in contrast to calling `should_abort`), this information is not up-to-date.
- I reset the `abort_requested` to `False` when the job is finished.
The migration becomes quite long as it is not possible to remove a value (in this case, `aborting`) from an enum type. A new enum type must be created, the old one must be swapped out, and all dependencies (indexes, functions, trigger) must be recreated. I tried to add good comments to make things clear.

I will switch the branch to merge against when we have a new branch for our beta.

onlyann:
> The migration becomes quite long as it is not possible to remove a value (in this case, aborting) from an enum type. A new enum type must be created, the old one must be swapped out, and all dependencies (indexes, functions, trigger) must be recreated. I tried to add good comments to make things clear.

That is the main drawback of postgres `enum`. They are not designed to have their values deleted

> I haven't intentionally made abort_requested available on the [Job](https://github.com/openradx/procrastinate/blob/463511ce7a39f668b6aea03c2b62497e2b24711f/procrastinate/jobs.py#L46) itself because if the user checks job.abort_requested of a passed context inside a task (in contrast to calling should_abort), this information is not up-to-date.

And as long as the `Job` class is marked as frozen, we can't simply change the flag.

Here is a thought on how the `should_abort` could work:

The task doesn't call the database through the `JobManager`.
Instead, instead: 
- a side task from the worker polls the database at a configured interval for any job that should be aborted (and it would only be interested in the jobs that it currently runs)
- additionally, a listener is notified every time a job is requested to be aborted which provides low latency.
The worker maintains a set of the job ids that should be aborted.
The `JobContext` exposes a function that returns whether the job id is in that set.

This means the tasks can check through the `JobContext` without introducing database overhead.

For async task, the worker could call `cancel` on the asyncio Task. The user code can then handle `CancellationError` to control abortion.

medihack:
> Here is a thought on how the `should_abort` could work:
> 
> The task doesn't call the database through the `JobManager`. Instead:
> 
> * a side task from the worker polls the database at a configured interval for any job that should be aborted (and it would only be interested in the jobs that it currently runs)
> * additionally, a listener is notified every time a job is requested to be aborted which provides low latency.
> 
> The worker maintains a set of the job ids that should be aborted. The `JobContext` exposes a function that returns True if the job id is in that set.
> 
> This means the tasks can check through the `JobContext` without introducing database overhead.
> 
> For async task, the worker could call `cancel` on the asyncio Task. The user code can then handle `CancellationError` to control abortion.

Yes, that all sounds reasonable. But let's start by getting this PR (as it is) in the v3 branch, then the worker refactoring, and then start from there. At least we have a solid base, then. @ewjoachim Do you agree?

ewjoachim:
I agree. @onlyann your proposal seems to match what #1084 describes.

ewjoachim:
(is the PR ready for review after the worker refactor ?)

medihack:
> (is the PR ready for review after the worker refactor ?)

Yes, I think so.

ewjoachim (approved):
Congratulations, stellar job as always

PR #1140, opened 2024-08-01
Make django Admin interface support slightly better UX wise

ticosax:
- Display some fields on table view ["task_name", "queue_name", "attempts", "status"]
- Allow to filter on status
- Display Events inline

Sorry no screenshot.

Closes ø

### Successful PR Checklist:

- [ ] Tests
  - [x] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

#### PR label(s): 
  - [x] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20miscellaneous%20%F0%9F%91%BE


ewjoachim:
Hi there ! Thank you for your contribution ! As mentioned in [the 2.0 release note](https://github.com/procrastinate-org/procrastinate/releases/tag/2.0.0):
> The Admin panels are very very basic for now

😅 

Thank you for tackling this, but if you have some spare time, I think we could go a bit further!
Screenshot 2024-08-01 at 19 02 28
Screenshot 2024-08-01 at 19 02 39

- I think it would make sense if the first `list_display` element stays the pk. That or the `call_string` (potentially in a `<code>` tag)
- OR I think it makes sense if the kwargs stay displayed in the listing view. It might make sense to pretty-print them and put them in a `<pre>` tag.
- It might make sense to only display the name of the task function and not the module names, optionally keep the initials (`p.d.d.t.index_book`)
- Would it make sense to include a brief summary of the last event too ? ("`deferred` 5 minutes ago" with a `title` mentioning the exact date time)

I tend to find proper use of bold, italics, underline etc, and the occasional emoji very nice. Anything that helps.
Colors can be nice though with some caveats: there are colorblind people, so color should duplicate some info, not be the only way to get some info. Can be text color or background color. Django Admin now supports both light and dark mode, so we need to check contrast with both.

Regarding the details view, I like the addition of events as inline. I think the only thing worth bikeshedding is the order of fields. I'd say: id, task name, args, status, queue, lock, queuing lock, priority, scheduled at, attempts

Thanks a lot :) If you'd rather not spend longer on this, please let us know, we can merge thing and follow up for other changes.

ticosax:
Thanks for the feedback. I'll to see how far I can get to incorporate all the suggestions.
Stay tuned

ticosax:
>     Would it make sense to include a brief summary of the last event too ? ("deferred 5 minutes ago" with a title mentioning the exact date time)

It would mean introducing another dependency similar to `humanize` to pretty print the timedelta.
I don't think it's worth it.

I will push my interpretation of your feedback. They improve even more the readability. I hope it matches your expectations :)

ewjoachim:
I could be wrong but [timesince](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#std-templatefilter-timesince) doesn't require humanize, does it ? (There's a python callable version in `django.utils.timesince`, I believe) (note: maybe I missed a deprecation notice or something?)

ticosax:
> I could be wrong but [timesince](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#std-templatefilter-timesince) doesn't require humanize, does it ? (There's a python callable version in `django.utils.timesince`, I believe) (note: maybe I missed a deprecation notice or something?)

you didn't missed anything. I did. Thanks for the pointer. I will look into it.

ewjoachim:
I've added a few commit for additional ideas. Here's the result:

https://github.com/user-attachments/assets/06034308-8554-42da-af76-5602d1d5d158

I'm ready to merge. If you're ready too (in case you want to read/comment my commits), then let's merge.

ewjoachim (approved):
Stellar job 🤩, inspiring and fun. Love it (love it so much  I wanted to play too 😬)

procrastinate/contrib/django/admin.py
```diff
@@ -74,21 +80,38 @@ def has_delete_permission(self, request, obj=None):
...
     @admin.display(description="Args")
     def pretty_args(self, instance: models.ProcrastinateJob) -> str:
-        return mark_safe(f"<pre>{instance.args}</pre>")
+        indent = 2 if len(instance.args) > 1 or len(str(instance.args)) > 30 else None
+        pretty_json = json.dumps(instance.args, indent=indent)
+        if len(pretty_json) > 2000:
+            pretty_json = pretty_json[:2000] + "..."
+        return format_html(
+            '<pre style="margin: 0">{pretty_json}</pre>', pretty_json=pretty_json
+        )
```
ewjoachim, line 102: Just so you know: it's dangerous to mark_safe elements that can be controlled by external sources. `format_html` is exactly made for this: the html will be marked safe, but the dynamic elements within will be escaped.

Imagine if someone submitted a task with `{"tag": "</pre>"}` 😱

procrastinate/contrib/django/admin.py
```diff
@@ -74,21 +80,38 @@ def has_delete_permission(self, request, obj=None):
...
 
     @admin.display(description="Summary")
     def summary(self, instance: models.ProcrastinateJob) -> str:
         if last_event := instance.procrastinateevent_set.latest():  # type: ignore[attr-defined]
             return mark_safe(
```
ewjoachim, line 107: Here, `mark_safe` is safe because external values got escaped when the template was rendered, so the result of the temple is actually safe

procrastinate/contrib/django/admin.py
```diff
@@ -26,11 +77,41 @@ def has_add_permission(self, request, obj=None):
...
-    [
-        models.ProcrastinateJob,
-        models.ProcrastinateEvent,
-    ],
-    ProcrastinateAdmin,
-)
+    @admin.display(description="Summary")
+    def summary(self, instance: models.ProcrastinateJob) -> str:
+        if last_event := instance.procrastinateevent_set.latest():  # type: ignore[attr-defined]
+            return mark_safe(
```
ewjoachim, line 107: Here, `mark_safe` is safe because external values got escaped when the template was rendered, so the result of the temple is actually safe

procrastinate/contrib/django/admin.py
ticosax: TIL

ticosax:
Very nice Thanks for the additions. You can merge it.

PR #1156, opened 2024-08-12
Run sync task in its own thread

### LINKED ISSUE #1124: Creating an asyncio loop in a sync task
> One of my colleagues had a [...] problem. He didn't know that the worker runs under asyncio and used `asyncio.run` in a synchronous task to create a new event loop, leading to a similar error.
The worker is [calling `sync_to_async`](https://github.com/procrastinate-org/procrastinate/blob/e94b4d0427b75a479f20ceed57a93cb5477c831c/procrastinate/worker.py#L278) with the default `thread_sensitive` set to `True`. I made a quick and dirty attempt, set `thread_sensitive` to `False`, and our error was gone (in the end, we rewrote the code of the task to be async). I wonder what the (negative) implications would be if we change the worker code to use `thread_sensitive=False`.

_Originally posted by @medihack in https://github.com/procrastinate-org/procrastinate/issues/609#issuecomment-2245449737_

medihack:
Closes #1124

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/procrastinate/utils.py
+++ b/procrastinate/utils.py
@@ -104,7 +104,7 @@ async def sync_to_async(func: Callable[..., T], *args, **kwargs) -> T:
     Given a callable, return a callable that will call the original one in an
     async context.
     """
-    return await sync.sync_to_async(func)(*args, **kwargs)
+    return await sync.sync_to_async(func, thread_sensitive=False)(*args, **kwargs)
 
 
 def causes(exc: BaseException | None):
--- a/tests/acceptance/test_sync.py
+++ b/tests/acceptance/test_sync.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from time import sleep
 
 import pytest
 from asgiref.sync import sync_to_async
@@ -78,6 +79,30 @@ async def test_nested_sync_to_async(sync_app, async_app):
     await async_app.run_worker_async(queues=["default"], wait=False)
 
     assert sum_results == [3]
+
+
+async def test_sync_task_runs_in_parallel(sync_app, async_app):
+    results = []
+
+    @sync_app.task(queue="default", name="sync_task_1")
+    def sync_task_1():
+        for i in range(3):
+            sleep(0.1)
+            results.append(i)
+
+    @sync_app.task(queue="default", name="sync_task_2")
+    def sync_task_2():
+        for i in range(3):
+            sleep(0.1)
+            results.append(i)
+
+    sync_task_1.defer()
+    sync_task_2.defer()
+
+    async_app.tasks = sync_app.tasks
+    await async_app.run_worker_async(queues=["default"], concurrency=2, wait=False)
+
+    assert results == [0, 0, 1, 1, 2, 2]
 
 
 async def test_cancel(sync_app, async_app):
--- a/tests/acceptance/test_sync.py
+++ b/tests/acceptance/test_sync.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import asyncio
+
 import pytest
+from asgiref.sync import sync_to_async
 
 import procrastinate
 from procrastinate.contrib import psycopg2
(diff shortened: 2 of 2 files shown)


medihack:
Running the sync task in its own thread ensures that it is more isolated. The test looks a bit constructed but fails (even without any exception) when the sync task runs in the main thread (`thread_sensitive=True`).
I tried to find any negative consequences (especially for Django), but as db connections in Django are created per thread, I don't see any. However, I am not absolutely sure.

ewjoachim:
The discussions in Discord are prompting me to wonder: if we're considering changing this, should we consider running sync tasks in subprocesses ? It would avoid threading issues and would make it much easier to kill a task or to send a signal, the same way we do with async. But I don't want this decision to be made lightly and later regretted.

medihack:
When doing IO-heavy stuff, an async task is preferred over a threaded sync task. (If the task depends on a non-async third-party library, then the chances are high that there is an async alternative.) On the other hand, if the task is compute-heavy, then running it in its own process is the way to go.

As mentioned, we also could make this configurable. As an option for the worker, for example as [huey does it](https://huey.readthedocs.io/en/latest/consumer.html#worker-types) (but they chose threaded as the default), or as an option on the task itself (something like `@app.task(sync_threaded=True`). But even then, the question remains what the default should be (in my opinion, for the above-mentioned reason, a separate process).

ewjoachim:
If we did:
`async def` -> async (good for I/O & high level of parallelism, bad for CPU)
`def` -> subprocess (low level of parallelism, not specifically bad for I/O, good for CPU)

That would be not that far from what we have today, would it ?

onlyann:
It sounds reasonable to remove the thread sensitivity when wrapping sync tasks into a Thread.

Maybe this change could be merged while the conversation of introducing sub-processes for CPU bound tasks carries on?

medihack:
> If we did:
> `async def` -> async (good for I/O & high level of parallelism, bad for CPU)
> `def` -> subprocess (low level of parallelism, not specifically bad for I/O, good for CPU)

Yes, this is what I had in mind (when not making it configurable). We would then run the sync task in a `ProcessPoolExecutor`. I also looked a bit around how other distributed (Python) task queues handle this:

- Celery -> [process](https://docs.celeryq.dev/en/stable/userguide/workers.html#concurrency)
- RQ -> [process](https://python-rq.org/docs/workers/)
- Huey -> [thread by default, but separate process optional](https://huey.readthedocs.io/en/latest/consumer.html#worker-types)
- Django-q2 -> [process, but using a worker pool](https://django-q2.readthedocs.io/en/master/cluster.html)
- django-task -> [inside same thread and process](https://github.com/RealOrangeOne/django-tasks/blob/00a9bf6a0cbd21d537305a687be0a2d484ba341d/django_tasks/backends/database/management/commands/db_worker.py#L129)

So, the majority seem to use a separate process.

As you already mentioned, another advantage of using a separate process is that it can be killed. That way, we could even introduce some final shutdown timeout that forces the worker to be entirely stopped (see discussion on Discord). I don't know if this is even possible when only using threads. Another feature that would be possible by using a separate process for a sync task is to specify the maximum time a task can take.

> Maybe this change could be merged while the conversation of introducing sub-processes for CPU bound tasks carries on?

Either way, we said it is a breaking change (is it?!) targeted at v3. So, it won't be in a near-term 2.x release. Or should we merge this one into the main branch and work on a "sub-process" variant in the v3 branch?

onlyann:
> Either way, we said it is a breaking change (is it?!) targeted at v3. So, it won't be in a near-term 2.x release. Or should we merge this one into the main branch and work on a "sub-process" variant in the v3 branch?

A merge into `main` is fine if someone needs it prior to v3 release.
It could be argued it is not a breaking given there is no mention in the docs that sync tasks block each other, and I can hardly imagine a use of this library using long running sync tasks because of it.

That said, my comment was just about merging it into `v3`. Maybe it won't ever make it to the release if everything is a sub-process but we are better off iterating without thread sensitivity until then.

medihack:
> A merge into `main` is fine if someone needs it prior to v3 release. It could be argued it is not a breaking change given there is no mention in the docs that sync tasks block each other, and I can hardly imagine a use of this library using long running sync tasks because of it.

Yes, I think so, too. One can even argue that the current implementation is even a bug as concurrent sync tasks are never processed in parallel (but consecutively in the main thread).
@ewjoachim Would it be ok if we merge this into the main branch, make a minor release, and open another issue / PR with a subprocess implementation?

medihack:
> Yep. Will try to review asap !

Sure, no hurry. Just wanted to know if it's ok when I switch the target branch. I will change it to main.

PR #1172, opened 2024-08-25
Add support for abort notification

onlyann:
### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

## Context

So far, the only use of `LISTEN/NOTIFY` has been to let the worker know that a new job is ready to be processed. 

This extends `listen_notify` to accept different types of notification through using the `payload`:
- *job_inserted*: the existing notification
- *abort_job_requested*: when a running job is requested to be aborted.

Instead of making explicit calls to the database every time a job calls `should_abort`, the worker uses the same mechanism than for detecting new jobs, a combination of Postgres notification and polling.

## Other changes

The edge case of retrying a job to be aborted is handled at the database level instead of the worker.

## Next steps

Once this change lands in v3, it will help with cancellation during the shutdown event and #1084

### DIFF
--- a/docs/howto/advanced/cancellation.md
+++ b/docs/howto/advanced/cancellation.md
@@ -24,11 +24,10 @@ app.job_manager.cancel_job_by_id(33, delete_job=True)
 await app.job_manager.cancel_job_by_id_async(33, delete_job=True)
 ```
 
-## Mark a currently being processed job for abortion
+## Mark a running job for abortion
 
 If a worker has not picked up the job yet, the below command behaves like the
-command without the `abort` option. But if a job is already in the middle of
-being processed, the `abort` option marks this job for abortion (see below
+command without the `abort` option. But if a job is already running, the `abort` option marks this job for abortion (see below
 how to handle this request).
 
 ```python
@@ -38,10 +37,10 @@ app.job_manager.cancel_job_by_id(33, abort=True)
 await app.job_manager.cancel_job_by_id_async(33, abort=True)
 ```
 
-## Handle a abortion request inside the task
+## Handle an abortion request inside the task
 
 In our task, we can check (for example, periodically) if the task should be
-aborted. If we want to respect that request (we don't have to), we raise a
+aborted. If we want to respect that abortion request (we don't have to), we raise a
 `JobAborted` error. Any message passed to `JobAborted` (e.g.
 `raise JobAborted("custom message")`) will end up in the logs.
 
@@ -54,24 +53,10 @@ def my_task(context):
     do_something_expensive()
 ```
 
-There is also an async API
+Behind the scenes, the worker receives a Postgres notification every time a job is requested to abort, (unless `listen_notify=False`).
 
-```python
-@app.task(pass_context=True)
-async def my_task(context):
-  for i in range(100):
-    if await context.should_abort_async():
-      raise exceptions.JobAborted
-    do_something_expensive()
-```
-
-:::{warning}
-`context.should_abort()` and `context.should_abort_async()` does poll the
-database and might flood the database. Ensure you do it only sometimes and
-not from too many parallel tasks.
-:::
+The worker also polls (respecting `polling_interval`) the database for abortion requests, as long as the worker is running at least one job (in the absence of running job, there is nothing to abort).
 
 :::{note}
-When a task of a job that was requested to be aborted raises an error, the job
-is marked as failed (regardless of the retry strategy).
+When a job is requested to abort and that job fails, it will not be retried (regardless of the retry strategy).
 :::
--- a/procrastinate/job_context.py
+++ b/procrastinate/job_context.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import time
-from typing import Any, Iterable
+from typing import Any, Callable, Iterable
 
 import attr
 
@@ -54,6 +54,8 @@ class JobContext:
     additional_context: dict = attr.ib(factory=dict)
     task_result: Any = None
 
+    should_abort: Callable[[], bool]
+
     def evolve(self, **update: Any) -> JobContext:
         return attr.evolve(self, **update)
 
@@ -68,13 +70,3 @@ class JobContext:
             message += f" (started {duration:.3f} s ago)"
 
         return message
-
-    def should_abort(self) -> bool:
-        assert self.job.id
-        job_id = self.job.id
-        return self.app.job_manager.get_job_abort_requested(job_id)
-
-    async def should_abort_async(self) -> bool:
-        assert self.job.id
-        job_id = self.job.id
-        return await self.app.job_manager.get_job_abort_requested_async(job_id)
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -4,9 +4,10 @@ import datetime
 import functools
 import logging
 from enum import Enum
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypedDict, Union
 
 import attr
+from typing_extensions import Literal
 
 from procrastinate import types
 
@@ -20,6 +21,19 @@ DEFAULT_QUEUE = "default"
 DEFAULT_PRIORITY = 0
 
 cached_property = getattr(functools, "cached_property", property)
+
+
+class JobInserted(TypedDict):
+    type: Literal["job_inserted"]
+    job_id: int
+
+
+class AbortJobRequested(TypedDict):
+    type: Literal["abort_job_requested"]
+    job_id: int
+
+
+Notification = Union[JobInserted, AbortJobRequested]
 
 
 def check_aware(
@@ -82,6 +96,9 @@ class Job:
     #: Number of times the job has been tried.
     attempts: int = 0
 
+    # True if the job is requested to abort
+    abort_requested: bool = False
+
     @classmethod
     def from_row(cls, row: dict[str, Any]) -> Job:
         return cls(
@@ -95,6 +112,7 @@ class Job:
             scheduled_at=row["scheduled_at"],
             queue=row["queue_name"],
             attempts=row["attempts"],
+            abort_requested=row.get("abort_requested", False),
         )
 
     def asdict(self) -> dict[str, Any]:
--- a/procrastinate/manager.py
+++ b/procrastinate/manager.py
@@ -1,15 +1,21 @@
 from __future__ import annotations
 
-import asyncio
 import datetime
+import json
 import logging
-from typing import Any, Iterable, NoReturn
+from typing import Any, Awaitable, Iterable, NoReturn, Protocol
 
 from procrastinate import connector, exceptions, jobs, sql, utils
 
 logger = logging.getLogger(__name__)
 
 QUEUEING_LOCK_CONSTRAINT = "procrastinate_jobs_queueing_lock_idx"
+
+
+class NotificationCallback(Protocol):
+    def __call__(
+        self, *, channel: str, notification: jobs.Notification
+    ) -> Awaitable[None]: ...
 
 
 def get_channel_for_queues(queues: Iterable[str] | None = None) -> Iterable[str]:
@@ -360,42 +366,6 @@ class JobManager:
         )
         return jobs.Status(result["status"])
 
-    def get_job_abort_requested(self, job_id: int) -> bool:
-        """
-        Check if a job is requested for abortion.
-
-        Parameters
-        ----------
-        job_id : ``int``
-            The id of the job to get the abortion request of
-
-        Returns
-        -------
-        ``bool``
-        """
-        result = self.connector.get_sync_connector().execute_query_one(
-            query=sql.queries["get_job_abort_requested"], job_id=job_id
-        )
-        return bool(result["abort_requested"])
-
-    async def get_job_abort_requested_async(self, job_id: int) -> bool:
-        """
-        Check if a job is requested for abortion.
-
-        Parameters
-        ----------
-        job_id : ``int``
-            The id of the job to get the abortion request of
-
-        Returns
-        -------
-        ``bool``
-        """
-        result = await self.connector.execute_query_one_async(
-            query=sql.queries["get_job_abort_requested"], job_id=job_id
-        )
-        return bool(result["abort_requested"])
-
     async def retry_job(
         self,
         job: jobs.Job,
@@ -491,26 +461,39 @@ class JobManager:
         )
 
     async def listen_for_jobs(
-        self, *, event: asyncio.Event, queues: Iterable[str] | None = None
+        self,
+        *,
+        on_notification: NotificationCallback,
+        queues: Iterable[str] | None = None,
     ) -> None:
         """
-        Listens to defer operation in the database, and raises the event each time an
-        defer operation is seen.
+        Listens to job notifications from the database, and invokes the callback each time an
+        notification is received.
 
         This coroutine either returns ``None`` upon calling if it cannot start
         listening or does not return and needs to be cancelled to end.
 
         Parameters
         ----------
-        event:
-            This event will be set each time a defer operation occurs
-        queues:
-            If ``None``, all defer operations will be considered. If an iterable of
+        on_notification : ``connector.Notify``
+            A coroutine that will be called and awaited every time a notification is received
+        queues : ``Optional[Iterable[str]]``
+            If ``None``, all notification will be considered. If an iterable of
             queue names is passed, only defer operations on those queues will be
             considered. Defaults to ``None``
         """
+
+        async def handle_notification(channel: str, payload: str):
+            notification: jobs.Notification = json.loads(payload)
+            logger.debug(
+                f"Received {notification['type']} notification from channel",
+                extra={channel: channel, payload: payload},
+            )
+            await on_notification(channel=channel, notification=notification)
+
         await self.connector.listen_notify(
-            event=event, channels=get_channel_for_queues(queues=queues)
+            on_notification=handle_notification,
+            channels=get_channel_for_queues(queues=queues),
         )
 
     async def check_connection_async(self) -> bool:
@@ -836,3 +819,12 @@ class JobManager:
                 }
             )
         return result
+
+    async def list_jobs_to_abort_async(self, queue: str | None = None) -> Iterable[int]:
+        """
+        List ids of running jobs to abort
+        """
+        rows = await self.connector.execute_query_all_async(
+            query=sql.queries["list_jobs_to_abort"], queue_name=queue
+        )
+        return [row["id"] for row in rows]
--- a/procrastinate/sql/schema.sql
+++ b/procrastinate/sql/schema.sql
@@ -269,12 +269,30 @@ DECLARE
     _job_id bigint;
 BEGIN
     UPDATE procrastinate_jobs
-    SET status = 'todo',
-        attempts = attempts + 1,
-        scheduled_at = retry_at,
-        priority = COALESCE(new_priority, priority),
-        queue_name = COALESCE(new_queue_name, queue_name),
-        lock = COALESCE(new_lock, lock)
+    SET status = CASE
+            WHEN NOT abort_requested THEN 'todo'::procrastinate_job_status
+            ELSE 'failed'::procrastinate_job_status
+        END,
+        attempts = CASE
+            WHEN NOT abort_requested THEN attempts + 1
+            ELSE attempts
+        END,
+        scheduled_at = CASE
+            WHEN NOT abort_requested THEN retry_at
+            ELSE scheduled_at
+        END,
+        priority = CASE
+            WHEN NOT abort_requested THEN COALESCE(new_priority, priority)
+            ELSE priority
+        END,
+        queue_name = CASE
+            WHEN NOT abort_requested THEN COALESCE(new_queue_name, queue_name)
+            ELSE queue_name
+        END,
+        lock = CASE
+            WHEN NOT abort_requested THEN COALESCE(new_lock, lock)
+            ELSE lock
+        END
     WHERE id = job_id AND status = 'doing'
     RETURNING id INTO _job_id;
     IF _job_id IS NULL THEN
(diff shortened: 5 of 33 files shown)


medihack:
Excellent. I will have a look at it tomorrow 🙂. I think it's a good idea to handle more stuff at the database level (as long as we can still test it).

onlyann:
> Excellent. I will have a look at it tomorrow 🙂. I think it's a good idea to handle more stuff at the database level (as long as we can still test it).

It also removes some race condition with the previous approach.

Glad you had an acceptance test written against it already or I will have surely missed it.

procrastinate/sql/schema.sql
```diff
@@ -283,13 +301,30 @@ BEGIN
 END;
 $$;
 
-CREATE FUNCTION procrastinate_notify_queue()
+CREATE OR REPLACE FUNCTION procrastinate_notify_queue_job_inserted()
```
medihack, line 304: Do we really need `CREATE OR REPLACE` in the schema.sql? It's no harm, but I think just `CREATE` should be enough. (I think it's also not necessary for `procrastinate_finish_job`).

tests/acceptance/test_async.py
```diff
@@ -11,7 +11,7 @@
 from procrastinate.jobs import Status
 
 
-@pytest.fixture(params=["psycopg_connector", "aiopg_connector"])
+@pytest.fixture(params=["aiopg_connector"])
```
medihack, line 14: What about the `psycopg_connector`?

medihack:
> > Excellent. I will have a look at it tomorrow 🙂. I think it's a good idea to handle more stuff at the database level (as long as we can still test it).
> 
> It also removes some race condition with the previous approach.
> 
> Glad you had an acceptance test written against it already or I will have surely missed it.

Cool, let me know when it's ready for review (still a draft currently and some connector excluded from [tests/acceptance/test_async.py](https://github.com/procrastinate-org/procrastinate/pull/1172/files#diff-bc75006cef45393b9e2aa448f8acc90415da2e48c4cba263a17d97388de2d576)). But looks pretty good at first glance.

procrastinate/sql/schema.sql
onlyann: I guess not. I may as well replace it there too to avoid ambiguity in the future

docs/howto/advanced/cancellation.md
```diff
@@ -54,22 +54,8 @@ def my_task(context):
...
-      raise exceptions.JobAborted
-    do_something_expensive()
-```
-
-:::{warning}
-`context.should_abort()` and `context.should_abort_async()` does poll the
-database and might flood the database. Ensure you do it only sometimes and
-not from too many parallel tasks.
-:::
+The worker receives a Postgres notification every time a running job is requested to abort, unless `listen_notify=False`.
```
medihack, line 57: How about adding something like "Internally, the worker receives ..."? It would make it clearer that this is nothing the user has to handle but something that works behind the curtain.

We should also update the docstrings of [`listen_notify`](https://github.com/onlyann/procrastinate/blob/9e4e0dcb140fcdc56be3817e4a963aa222fde77a/procrastinate/app.py#L271) and [`polling_interval`](https://github.com/onlyann/procrastinate/blob/9e4e0dcb140fcdc56be3817e4a963aa222fde77a/procrastinate/app.py#L261-L262) that this is not just for requesting jobs.

So, `should_abort()` is now used for async and sync tasks. Correct? Haven't you originally planned to cancel the async tasks automatically on abort? (The thing I was a bit ambivalent about). Just wondering.

procrastinate/sql/schema.sql
```diff
@@ -269,12 +269,30 @@ DECLARE
...
     UPDATE procrastinate_jobs
-    SET status = 'todo',
-        attempts = attempts + 1,
-        scheduled_at = retry_at,
-        priority = COALESCE(new_priority, priority),
-        queue_name = COALESCE(new_queue_name, queue_name),
-        lock = COALESCE(new_lock, lock)
+    SET status = CASE
+            WHEN abort_requested = false THEN 'todo'::procrastinate_job_status
+            ELSE 'failed'::procrastinate_job_status
```
medihack, line 274: How about  just using NOT:
```sql
WHEN NOT abort_requested THEN 'todo'::procrastinate_job_status
ELSE 'failed'::procrastinate_job_status
```
or maybe even just the other way around
```sql
WHEN abort_requested THEN 'failed'::procrastinate_job_status
ELSE 'todo'::procrastinate_job_status
```
Same for other cases and migration.

procrastinate/worker.py
```diff
@@ -338,6 +335,39 @@ async def run(self):
...
+    async def _handle_notification(
+        self, *, channel: str, notification: jobs.Notification
+    ):
+        if (
+            notification["type"] == "job_inserted"
+            and notification["status"] == jobs.Status.TODO.value
+        ):
+            self._new_job_event.set()
+        elif notification["type"] == "abort_job_requested":
+            self._job_ids_to_abort.add(notification["job_id"])
```
medihack, line 347: It seems that aborted IDs only accumulate in `_job_ids_to_abort`. Should we remove aborted job IDs from this set or clean it up somehow differently?

docs/howto/advanced/cancellation.md
onlyann: > So, should_abort() is now used for async and sync tasks. Correct? 

Correct

> Haven't you originally planned to cancel the async tasks automatically on abort?

Yes. I aim to keep this PR focused or notification support and open another about issuing an asyncio cancel on async tasks.

procrastinate/sql/schema.sql
onlyann: We can invert the condition.
The reason I went with this order is that the first when represents the normal case on retry.

procrastinate/worker.py
onlyann: I had the same thought. 

At the time, I brushed it away deciding that it will probably not be a real problem in practice.

That said, it should be straightforward to attempt to remove any job id that completed from that list. 

That should only leave ids in the set in certain circumstances (such as multiple worker processes sharing the same queue(s).)

docs/howto/advanced/cancellation.md
onlyann: I amended the documentation.
I also noticed I need to rewrite some of the discussion section around the worker. I will however leave that part for another PR as it is not related to abort notification.

procrastinate/worker.py
```diff
@@ -338,6 +337,49 @@ async def run(self):
...
+            if not self._running_jobs:
+                logger.debug("Not querying jobs to abort because no job is running")
+                continue
+            try:
+                job_ids = await self.app.job_manager.list_jobs_to_abort_async()
+                running_job_ids = {
+                    c.job.id for c in self._running_jobs.values() if c.job.id
+                }
+
+                self._job_ids_to_abort.update(running_job_ids.intersection(job_ids))
```
ewjoachim, line 372: I'm very slightly worried that there are 2 independent places where we react to changes in the abort job list. I think I'd feel slightly better if there was a single function (well, coroutine) handling this.

That said, while the work we do on `self._job_ids_to_abort` is seemingly identical (replace a set vs update a set), it will be different when we actually abort the tasks, because we'll need to identify what jobs should be aborted which we haven't aborted already.

So all in all... Maybe we'll see how to refactor when we add the abortion itself. But I'd prefer if we don't have 2 very related pieces of code that need to do roughly the same thing sitting hundreds of lines appart.

Another solution could be having a queue and a coroutine awaiting on that queue, but that marginally doesn't solve the issue that we're going to do slightly different things on the notification vs the poll.

Trying to put a brain cell on that. It's not blocking for merging this PR.

procrastinate/worker.py
```diff
@@ -365,10 +407,13 @@ async def _shutdown(self, side_tasks: list[asyncio.Task]):
 
     def _start_side_tasks(self) -> list[asyncio.Task]:
         """Start side tasks such as periodic deferrer and notification listener"""
-        side_tasks = [asyncio.create_task(self.periodic_deferrer(), name="deferrer")]
-        if self.wait and self.listen_notify:
+        side_tasks = [
+            asyncio.create_task(self.periodic_deferrer(), name="deferrer"),
```
ewjoachim, line 411: should `periodic_deferrer` start with a `_` too ? Is there a rationale on what gets a `_` vs on on a class that's itself internal ?

procrastinate/worker.py
```diff
@@ -365,10 +407,13 @@ async def _shutdown(self, side_tasks: list[asyncio.Task]):
...
+            asyncio.create_task(self._poll_jobs_to_abort(), name="poll_jobs_to_abort"),
```
ewjoachim, line 412: Given the 2 polling operations are independent, should the intervals be controlled by the same variable ? (can be handled in a different PR)

tests/acceptance/test_async.py
```diff
@@ -108,47 +108,51 @@ def example_task():
...
-    @async_app.task(queue="default", name="task2", pass_context=True)
-    def task2(context):
-        while True:
-            time.sleep(0.02)
-            if context.should_abort():
-                raise JobAborted
+@pytest.mark.parametrize("listen_notify", [False, True])
+@pytest.mark.parametrize("sync_task", [False, True])
+async def test_abort(async_app: app_module.App, listen_notify, sync_task):
+    if sync_task:
```
ewjoachim, line 114: I'm not a huge fan of `if` statements in the tests. When what varies cannot be expressed as simple values in the parametrize, I prefer duplicating the test.

(I like the test code to read as simply as possible. Procrastinate tests need to be somewhat smart because it's hard to test complex things simply, but let's not add to that by trying to be too DRY. Being DRY is usually nice for normal code, but it's much less of a requirement for test code)

tests/acceptance/test_async.py
```diff
@@ -108,47 +108,51 @@ def example_task():
...
     assert result is True
 
-    await asyncio.sleep(0.1)
-    result = await async_app.job_manager.cancel_job_by_id_async(job2_id, abort=True)
-    assert result is True
-
-    await worker_task
+    # when listening for notifications, job should cancel within ms
+    # if notifications are disabled, job will only cancel after polling_interval
+    await asyncio.wait_for(
```
ewjoachim, line 150: If I read this correctly, it means the test may take up to a second in normal conditions ?

I'd rather we keed tests shorter than this, we're already experiencing tests that are growing in duration, and if we can avoid adding tests that take a second when possible, that's better.

ewjoachim (commented):
Well, that was awesome :) Still plenty of comments, but definitly going the right way. Thank you!

docs/howto/advanced/cancellation.md
medihack: Sure, but maybe we can have updated docstrings for [`listen_notify`](https://github.com/onlyann/procrastinate/blob/9e4e0dcb140fcdc56be3817e4a963aa222fde77a/procrastinate/app.py#L271) and [`polling_interval`](https://github.com/onlyann/procrastinate/blob/9e4e0dcb140fcdc56be3817e4a963aa222fde77a/procrastinate/app.py#L261-L262).

Currently, it says for `polling_interval`:
```markdown
polling_interval : ``float``
    Indicates the maximum duration (in seconds) the worker waits between
    each database job poll. Raising this parameter can lower the rate at which
    the worker makes queries to the database for requesting jobs.
    (defaults to 5.0)
```
How about:
```markdown
polling_interval : ``float``
    Indicates the maximum duration (in seconds) the worker waits between
    each database job poll. Raising this parameter can lower the rate at which
    the worker makes queries to the database for requesting jobs or checking
    for job abortion requests.
    (defaults to 5.0)
```

And for `listen_notify` it says:
```markdown
listen_notify : ``bool``
    If ``True``, the worker will dedicate a connection from the pool to
    listening to database events, notifying of newly available jobs.
    If ``False``, the worker will just poll the database periodically
    (see ``polling_interval``). (defaults to ``True``)
```
How about:
```markdown
listen_notify : ``bool``
    If ``True``, the worker will dedicate a connection from the pool to
    listening to database events, notifying of newly available jobs, or
    job abortion requests.
    If ``False``, the worker will just poll the database periodically
    (see ``polling_interval``). (defaults to ``True``)
```

ewjoachim: (while we're at it, 
```
If ``False``, the worker will just poll the database periodically
```
might be misleading in implying that this doesn't happen with ``True``
Maybe:
> If `False`, the worker won't listen to database events, so it will only
> know about new jobs & abortion requests through the polling mechanism
```
?

**Comment by medihack** [comment]
So, after updating those docstrings, I have nothing to add. You contributed a great feature here. Thanks a lot 🙂.
@ewjoachim Do you have something to add?

File: procrastinate/job_context.py [hunk 11]
```diff
@@ -64,6 +64,8 @@ class JobContext:
     additional_context: dict = attr.ib(factory=dict)
     task_result: Any = None
 
+    jobs_to_abort: set[int]
```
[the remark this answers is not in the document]
medihack: If we ever use subprocesses for sync tasks, we should also consider the most straightforward way to provide this information to the subprocess. Or we can decide when the time comes.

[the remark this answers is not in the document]
ewjoachim: (That mainly highlights the importance of making this a "private" attribute (leading `_`). That said, I'm in favor of not trying to guess what we'll end up doing for subprocesses _yet_ :D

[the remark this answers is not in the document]
medihack: Me, too 😞. At least not for a v3 release.

docs/howto/advanced/cancellation.md
onlyann: Pushing some update on this from your feedbacks.
Let me know what you think

procrastinate/job_context.py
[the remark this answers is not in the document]
onlyann: > I wonder if we should just remove frozen=True because, well, the context is effectively not immutable, rather than cheating. I think I'd prefer that (even if I love immutable objects)
Or instead of providing a jobs_to_abort set, we'd give the context a should_abort callable...
Any opinion ?

I went through a similar thought process. 
I also prefer dropping the `frozen` flag and turning it into a boolean

procrastinate/job_context.py
```diff
@@ -80,11 +82,4 @@ def job_description(self, current_timestamp: float) -> str:
...
 
     def should_abort(self) -> bool:
-        assert self.job.id
-        job_id = self.job.id
-        return self.app.job_manager.get_job_abort_requested(job_id)
-
-    async def should_abort_async(self) -> bool:
-        assert self.job.id
-        job_id = self.job.id
-        return await self.app.job_manager.get_job_abort_requested_async(job_id)
```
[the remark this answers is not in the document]
onlyann: What about extending the `Job` class instead to have `abort_requested`. This way, the existing `list_job` operation covers it and there is no need to have distinct operations for every piece of state?

procrastinate/jobs.py
```diff
@@ -22,6 +23,20 @@
 cached_property = getattr(functools, "cached_property", property)
 
 
+class JobInserted(TypedDict):
+    type: Literal["job_inserted"]
+    task_name: str
+    status: str
```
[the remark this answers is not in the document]
onlyann: > If a job is inserted, isn't its status necessarily todo ?

You're right.

I designed this notification to cover not only inserted job rows with `todo` but any other status. 
In retrospect, I can't think of any other use case and if that happens in the future, we'll deal with it when it comes.

I added the `task_name` for logging purposes at the time but didn't end up using it. 

I will change it to contain the `job_id` since this in itself should be enough for fetching additional job information.
The reason I am mentioning it is that I think it would be a useful (future) feature to expose notifications to end users too.

procrastinate/manager.py
```diff
@@ -12,6 +12,12 @@
 QUEUEING_LOCK_CONSTRAINT = "procrastinate_jobs_queueing_lock_idx"
 
 
+class Notify(Protocol):
```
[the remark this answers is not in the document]
onlyann: Not intended to keep this one around, good catch.

I'll change the name too.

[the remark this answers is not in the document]
onlyann: I'll change the name to `NotificationCallback`. 
FYI, `Notify` is the name used by `psycopg`

procrastinate/worker.py
onlyann: Not sure about Python. Coming from a different background, I see `_` being used when it is not meant to be used outside the class, whether the class is internal or not. 
Changing to `_periodic_deferrer` but happy to go with whatever convention you prefer here.

procrastinate/job_context.py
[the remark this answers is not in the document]
onlyann: Thinking more about it, I am concerned that dropping frozen opens the door to all kinds of issues if the end user code starts changing that object. I will give the Callable option a try.

procrastinate/sql/schema.sql
```diff
@@ -269,12 +269,30 @@ DECLARE
     _job_id bigint;
 BEGIN
     UPDATE procrastinate_jobs
-    SET status = 'todo',
-        attempts = attempts + 1,
-        scheduled_at = retry_at,
-        priority = COALESCE(new_priority, priority),
-        queue_name = COALESCE(new_queue_name, queue_name),
-        lock = COALESCE(new_lock, lock)
+    SET status = CASE
```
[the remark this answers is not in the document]
onlyann: I am not following the example. 
Where does `abort_requested` come from when it is outside the `UPDATE` statement?

procrastinate/sql/schema.sql
```diff
@@ -382,10 +417,15 @@ $$;
 
 -- Triggers
 
-CREATE TRIGGER procrastinate_jobs_notify_queue
+CREATE TRIGGER procrastinate_jobs_notify_queue_job_inserted
     AFTER INSERT ON procrastinate_jobs
```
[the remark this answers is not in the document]
onlyann: Good point. 
On a similar note, I think there is also opposite scenario where a job is inserted with a future `scheduled_at`, which will trigger a notification when the job is not yet available.

procrastinate/worker.py
```diff
@@ -224,12 +225,7 @@ async def ensure_async() -> Callable[..., Awaitable]:
...
             exc_info = e
 
-            assert job.id
-            abort_requested = await self.app.job_manager.get_job_abort_requested_async(
-                job_id=job.id
-            )
-
-            if not isinstance(e, exceptions.JobAborted) and not abort_requested:
+            if not isinstance(e, exceptions.JobAborted):
                 job_retry = (
```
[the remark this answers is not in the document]
onlyann: Correct

procrastinate/worker.py
onlyann: I thought about it too. We can expose 2 different settings.  Happy for it to be a separate PR.

procrastinate/job_context.py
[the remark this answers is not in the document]
ewjoachim: Seems to make sense. I agree!

procrastinate/sql/schema.sql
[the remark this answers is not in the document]
ewjoachim: Hm you're right. Though maybe doing a quick SELECT to get the `abort_requested` might be enough. It's not very important but the current code is quite hard to read and I believe will be harder to maintain.

procrastinate/job_context.py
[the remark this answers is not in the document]
ewjoachim: Ah, I might not have understood your proposal.

I think it's worth keeping a `should_abort_async` on the context just because we advertised it before because it's really not going to get in the way of maintenance, independently from how we can make the info available on the job. It's literally:

```python
async def should_abort_async(self):
    # As it stands, `should_abort` doesn't do I/O as of now.
    return self.should_abort()
```

I don't see a case where this get in our way.

ewjoachim (commented):
We're almost there. This `should_abort_async` question and unless I forgot something, we're good !

procrastinate/job_context.py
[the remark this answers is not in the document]
onlyann: Given how short lived that function is, how simple it is to migrate to the non async version and the opportunity to have breaking changes as part of v3, it seems reasonable to remove it.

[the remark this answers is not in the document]
ewjoachim: Ok, won't fight you on this.

procrastinate/sql/schema.sql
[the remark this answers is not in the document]
onlyann: I don't know if this is much better to have something like:

```sql
CREATE OR REPLACE FUNCTION procrastinate_retry_job(
    job_id bigint,
    retry_at timestamp with time zone,
    new_priority integer,
    new_queue_name varchar,
    new_lock varchar
)
RETURNS void
LANGUAGE plpgsql
AS $$
DECLARE
    _job_id bigint;
    _abort_requested boolean;
BEGIN
    -- First, check if the job exists and get its abort_requested status
    SELECT id, abort_requested 
    INTO _job_id, _abort_requested
    FROM procrastinate_jobs
    WHERE id = job_id AND status = 'doing';

    IF _job_id IS NULL THEN
        RAISE EXCEPTION 'Job was not found or not in "doing" status (job id: %)', job_id;
    END IF;

    -- Update the job based on abort_requested status
    UPDATE procrastinate_jobs
    SET
        (status, attempts, scheduled_at, priority, queue_name, lock) =
        CASE
            WHEN NOT _abort_requested THEN
                ('todo'::procrastinate_job_status,
                 attempts + 1,
                 retry_at,
                 COALESCE(new_priority, priority),
                 COALESCE(new_queue_name, queue_name),
                 COALESCE(new_lock, lock))
            ELSE
                ('failed'::procrastinate_job_status,
                 attempts,
                 scheduled_at,
                 priority,
                 queue_name,
                 lock)
        END
    WHERE id = job_id;
END;
$$;
```

procrastinate/worker.py
onlyann: I refactored it to consolidate the logic.

procrastinate/worker.py
```diff
@@ -388,6 +370,10 @@ async def _poll_jobs_to_abort(self):
                 )
                 # recover from errors and continue polling
 
+    def _handle_abort_jobs_requested(self, job_ids: Iterable[int]):
+        running_job_ids = {c.job.id for c in self._running_jobs.values() if c.job.id}
+        self._job_ids_to_abort.update(running_job_ids.intersection(job_ids))
```
ewjoachim, line 375: `running_job_ids.intersection(job_ids)` can be written as:
```python
self._job_ids_to_abort |= running_job_ids & job_ids
```

Also, I believe it's safer to do:
```python
self._job_ids_to_abort |= job_ids
self._job_ids_to_abort &= running_job_ids
```
(this would effectively remove from _job_ids_to_abort any id that's not a currently running job)

ewjoachim (commented):
One last suggestion on your last commit, and I'm good.

docs/howto/advanced/cancellation.md
medihack: What does the bullet point `Fetching updates for running jobs` mean? Does the polling more than checking for new jobs and abortion requests?

ewjoachim (approved):

ewjoachim:
https://github.com/procrastinate-org/procrastinate/pull/1179
PR should be merged quickly. Sorry for the mess (it's all GitHub's fault :D )

onlyann:
> #1179 PR should be merged quickly. Sorry for the mess (it's all GitHub's fault :D )

No worries. I can rebase once it lands to v3 branch

ewjoachim:
v3 is up to date :)

onlyann:
Rebased and squashed 
All 🟢  🥳

ewjoachim:
And Merged !

PR #1211, opened 2024-10-05
v3: fix the migration mess

### LINKED ISSUE #1189: [v3] Migrations backwards compatibility
According to [our own doc](https://procrastinate.readthedocs.io/en/stable/howto/production/migrations.html), migrations named with version X are the migrations you can apply once you run on version X.
According to our doc, people on v2.13 wishing to upgrade without disruption would:
- Change the code to v3.0
- Apply the v3.0 migrations
- Change the code to v3.1
- Apply the v3.1 migrations
All the while their instance is running and processing jobs.

Compatibility matrix:

| Code ➡️<br>⬇️ Schema | v2.13 | v3.0 | v3.1 |
|--------|--------|--------|--------|
| v2.13 | ✅ | ✅ | ❌ |
| v3.0 | ❌ | ✅ | ✅ |
| v3.1 | ❌ | ❌ | ✅ |

❌ means we don't have to make it compatible, but if it happens to be, it's not a problem, of course

We need to decide what we put in v3.0 and v3.1

Note: the change of `status = aborting -> abort_requested` might need a couple of trigger just for the time of the dual compatibility (adding the `abort_requested column and update its value based on the status, then add a trigger that synchronizes the 2 columns, and in the next release, delete the triggers and the column)

ewjoachim:
Closes #1189

> [!WARNING]
> the word "mess" is is solely directed at the mess **I** created trhough unclear rules and overly complex processes, the work done so far by everyone has been amazing

Introduce a new migration process: pre-migrations and post-migrations.
- They will be tied to the corresponding release
- Reworked v3 migrations with this new template in mind
- They allow us to provide a much saner path for blue-green deployment
- The process is currently not tested (but nor was the previous system)
- We will need to be careful when we do a release to rename the migrations in ongoing PRs (as today)
- We (hopefully) won't have to do multiple releases just for the sake of blue-green compatibility.
- I'm adding versioning in the name of all internal objects (functions, triggers, etc) (as mentionned [here](https://github.com/procrastinate-org/procrastinate/issues/1189#issuecomment-2376450912))

So the idea is that:
- Procrastinate v2 with the pre-migrations & without the post-migrations should work
- Procrastinate v3 with the pre-migrations & without the post-migrations should work
- Procrastinate v3 with all migrations should be what we currently have in v3

(I'm 80% sure that I have missed a few corner cases)

Is it going in the right direction according to you @medihack @onlyann  ?

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 


onlyann:
Let's see if I understand. 

pre-script version N can be executed while the code version N-1 is running.
post-script version N is not compatible with version N-1 and must be executed once code version N is running.

Is the idea that procrastinate semantic versioning is at the programmatic API level and that any new version can possibly introduce breaking changes as long as there is the ability to split them in pre-scripts and post-scripts to maintain blue-green deployments of the codebase?

What are your thoughts about testing out that pre-migration script is backward compatible?

I like it.

ewjoachim:
> Is the idea that procrastinate semantic versioning is at the programmatic API level and that any new version can possibly introduce database breaking changes as long as there is the ability to split them in pre-scripts and post-scripts to maintain blue-green deployments of the codebase?

Yep. I think this change removes some hurdles that make keeping blue-green possible (and sane) while the current state of affairs made it more confusing than it needed to be and harder to achieve. We can iterate on the tooling to make this actually easy in the future.

> What are your thoughts about testing out that pre-migration script is backward compatible?

My ideas:
1/ we could run the tests on `n-1 & pre`. (The n-1 tests supposedly should work, though it will be very annoying to fix if they don't for some reason)
2/ for `n & pre`, i'm tempted to run the normal tests with an escape hatch allowing to easily skip the tests that can't run yet because they don't have the DB ready
3/ an alternative could be to redo our end to end tests to be a bit more versatile: test a large part of the external facing API and not necessarily handle the setup of the database.
This way, we may just run the e2e tests on `n-1&pre` and `n&pre` I think this would cover what we actually want to check and would be unlikely to cause issues. I'm not entirely sure if the end to end tests that would run on `n-1&pre` would be the end to end tests as defined on the `n` codebase or on the `n-1` codebase. Both have merits.

procrastinate/sql/migrations/03.00.00_50_post_cancel_notification.sql
```diff
@@ -1,66 +1,103 @@
...
--- Set abort requested flag on all jobs with 'aborting' status
+-- Remove all traces of the "aborting" status
+-- Last sanity update in case the trigger didn't work 100% of the time
 UPDATE procrastinate_jobs SET abort_requested = true WHERE status = 'aborting';
 
+DROP TRIGGER IF EXISTS procrastinate_trigger_sync_abort_requested_with_status_v1 ON procrastinate_jobs;
+DROP FUNCTION IF EXISTS procrastinate_sync_abort_requested_with_status_v1;
+
+-- Create a new enum type without 'aborting'
+CREATE TYPE procrastinate_job_status_v1 AS ENUM (
```
medihack, line 46: Must this not be created in `pre`? Some functions already use it there.

medihack:
I think I get the idea now and also like it 🙂. The separation between `pre` and `post` makes things clearer, as does the versioning of those functions. Should we also give the remaining functions and types a version to make it consistent (`procrastinate_job_event_type`, `procrastinate_defer_job`, `procrastinate_defer_periodic_job`, `procrastinate_unlink_periodic_defers`, `procrastinate_trigger_delete_jobs`?

ewjoachim:
I'm delighted by your answers. You're most probably right that we'll need to continue adjusting a few things.

Given that we need v3+pre to work, we need v3's queries to work with pre, which means the new enum needs to be supported in pre.

Post can only be used to clean the mess.

ewjoachim:
I probably won't have time to work on it Saturday, there might be a chance Sunday and then probably not for a few days, so I'd say: feel free to try something as long as we communicate where we're at.
Alternatively if you want, we can try to do it in pair programming on Sunday or on an evening next week :)

medihack:
> I probably won't have time to work on it Saturday, there might be a chance Sunday and then probably not for a few days, so I'd say: feel free to try something as long as we communicate where we're at. Alternatively if you want, we can try to do it in pair programming on Sunday or on an evening next week :)

Alright, let's see how far I get. I am probably not yet fully aware of the extent of this PR. And yes, pair programming sounds fun (I have never done anything like this before), but it sounds more like you will be tutoring me then ;-)

procrastinate/sql/migrations/03.00.00_50_post_cancel_notification.sql
medihack: I still think this is true. I will move it over to `procrastinate/sql/migrations/03.00.00_01_pre_cancel_notification.sql`

procrastinate/sql/migrations/03.00.00_50_post_cancel_notification.sql
```diff
@@ -1,66 +1,103 @@
...
 CREATE UNIQUE INDEX procrastinate_jobs_lock_idx ON procrastinate_jobs (lock) WHERE status = 'doing';
-CREATE INDEX procrastinate_jobs_id_lock_idx ON procrastinate_jobs (id, lock) WHERE status = ANY (ARRAY['todo'::procrastinate_job_status, 'doing'::procrastinate_job_status]);
+CREATE INDEX procrastinate_jobs_id_lock_idx ON procrastinate_jobs (id, lock) WHERE status = ANY (ARRAY['todo'::procrastinate_job_status_v1, 'doing'::procrastinate_job_status_v1]);
+
+-- Drop the old type
+DROP TYPE procrastinate_job_status;
 
--- Recreate and update the functions
-CREATE OR REPLACE FUNCTION procrastinate_fetch_job(
+-- Recreate and update the functions to use the new column
```
medihack, line 99: I don't think this is necessary when just changing the column type. And we already created those functions in `procrastinate/sql/migrations/03.00.00_01_pre_cancel_notification.sql`.

medihack:
Based on this one, I will work on a new PR (#1227). So we can throw mine away if I go in the wrong direction.

procrastinate/sql/migrations/03.00.00_50_post_cancel_notification.sql
```diff
@@ -160,34 +197,44 @@ BEGIN
...
+
+ALTER FUNCTION procrastinate_trigger_status_events_procedure_insert
+    RENAME TO procrastinate_trigger_function_status_events_insert_v1;
+
+CREATE TRIGGER procrastinate_trigger_status_events_insert_v1
+    AFTER INSERT ON procrastinate_jobs
+    FOR EACH ROW WHEN ((new.status = 'todo'::procrastinate_job_status_v1))
+    EXECUTE PROCEDURE procrastinate_trigger_function_status_events_insert_v1();
+
+CREATE FUNCTION procrastinate_trigger_function_status_events_update_v2()
```
medihack, line 210: Was it intentional that you renamed `procrastinate_trigger_status_events_procedure_update` to `procrastinate_trigger_function_status_events_update`?

ewjoachim:
(closed in favor of https://github.com/procrastinate-org/procrastinate/pull/1227 )

PR #1215, opened 2024-10-07
Shutdown timeout - abort sync tasks

### LINKED ISSUE #1185: Reconsider shutdown_timeout of the worker
In the v3 branch, the worker has a `shutdown_time` . The docs v3 currently say, `If a shutdown_timeout option is specified, the worker will attempt to abort all jobs that have not been completed by that time.` This has not been implemented, but the timeout cancels the asyncio task that runs all side coroutines and fetches/processes the jobs. It does not abort the running jobs.

There were several things [discussed on Discord](https://discord.com/channels/1197292025725329549/1197304897989787770/1271966442153050143):
- Adapt the documentation
- Send an abort request to running jobs
- Introduce another timeout that sends the abort request
- Provide an additional reason for the abortion

onlyann:
Closes #1185

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

This adds support for aborting both sync and async tasks when a worker shuts down and jobs exceed the specified timeout.

Prior to this change, only async tasks were aborted.

`shutdown_timeout `is renamed to `shutdown_graceful_timeout` for clarity.

### DIFF
--- a/docs/howto/advanced/shutdown.md
+++ b/docs/howto/advanced/shutdown.md
@@ -6,13 +6,23 @@ A worker will keep running until:
 - [task.cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) is called on the task created from `app.run_worker_async`
 
 When a worker is requested to stop, it will attempt to gracefully shut down by waiting for all running jobs to complete.
-If a `shutdown_timeout` option is specified, the worker will attempt to abort all jobs that have not completed by that time. Cancelling the `run_worker_async` task a second time also results in the worker aborting running jobs.
+If a `shutdown_graceful_timeout` option is specified, the worker will attempt to abort all jobs that have not completed by that time. Cancelling the `run_worker_async` task a second time also results in the worker aborting running jobs.
+
+The worker will then wait for all jobs to complete.
+
 
 :::{note}
-The worker aborts its remaining jobs by calling [task.cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) on the underlying asyncio task that runs the job.
+The worker aborts its remaining jobs by:
+- setting the context so that `JobContext.should_abort` returns `AbortReason.SHUTDOWN`
+- calling [task.cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) on the underlying asyncio task that runs the job when the job is asynchronous
 
-It is possible for that task to handle `asyncio.CancelledError` and even suppress the cancellation.
+Jobs that do not respect the request to abort will prevent the worker from shutting down until they complete. In a way, it will remain a graceful shutdown for those jobs even after `shutdown_graceful_timeout`.
+
+For more information, see {doc}`./cancellation`.
+
+Currently, Procrastinate does not provide a built-in method to forcefully terminate a worker. This is something you would want to do with your process manager (e.g. systemd, Docker, Kubernetes), which typically offers options to control process termination. In that case, your jobs will be considered stale, see {doc}`../production/retry_stalled_jobs`.
 :::
+
 ## Examples
 
 ### Run a worker until no job is left
@@ -29,15 +39,15 @@ async with app.open_async():
 async with app.open_async():
     # give jobs up to 10 seconds to complete when a stop signal is received
     # all jobs still running after 10 seconds are aborted
-    # In the absence of shutdown_timeout, the task will complete when all jobs have completed.
-    await app.run_worker_async(shutdown_timeout=10)
+    # In the absence of shutdown_graceful_timeout, the task will complete when all jobs have completed.
+    await app.run_worker_async(shutdown_graceful_timeout=10)
 ```
 
 ### Run a worker until its Task is cancelled
 
 ```python
 async with app.open_async():
-    worker = asyncio.create_task(app    run_worker_async())
+    worker = asyncio.create_task(app.run_worker_async())
     # eventually
     worker.cancel()
     try:
@@ -51,7 +61,7 @@ async with app.open_async():
 
 ```python
 async with app.open_async():
-    worker = asyncio.create_task(app.run_worker_async(shutdown_timeout=10))
+    worker = asyncio.create_task(app.run_worker_async(shutdown_graceful_timeout=10))
     # eventually
     worker.cancel()
     try:
@@ -66,7 +76,7 @@ async with app.open_async():
 
 ```python
 async with app.open_async():
-    # Notice that shutdown_timeout is not specified
+    # Notice that shutdown_graceful_timeout is not specified
     worker = asyncio.create_task(app.run_worker_async())
 
     # eventually
--- a/procrastinate/worker.py
+++ b/procrastinate/worker.py
@@ -39,7 +39,7 @@ class Worker:
         wait: bool = True,
         fetch_job_polling_interval: float = FETCH_JOB_POLLING_INTERVAL,
         abort_job_polling_interval: float = ABORT_JOB_POLLING_INTERVAL,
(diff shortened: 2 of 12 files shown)


ewjoachim:
Seems very good on first reading, I'll need a second pass on a reasonably sized screen (i.e. not my phone's) to give a reasonable opinion.

onlyann:
> Seems very good on first reading, I'll need a second pass on a reasonably sized screen (i.e. not my phone's) to give a reasonable opinion.

Thanks. No rush.

docs/howto/advanced/shutdown.md
```diff
@@ -6,13 +6,18 @@ A worker will keep running until:
...
+The worker will then wait for all jobs to complete.
+
 
 :::{note}
-The worker aborts its remaining jobs by calling [task.cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) on the underlying asyncio task that runs the job.
+The worker aborts its remaining jobs by:
+- setting `JobContext.should_abort` to `True`
+- calling [task.cancel](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) on the underlying asyncio task that runs the job when the job is asynchronous.
+Jobs that do not respect the request to abort will prevent the worker from shutting down.
+For more information, see {doc}`howto/cancellation`.
```
medihack, line 19: It's `howto/advanced/cancellation`. Then, building the docs should run through without a warning, and all checks pass.

medihack:
Very easy to understand now 🙂

Can we also fix this typo here: https://github.com/procrastinate-org/procrastinate/blob/183d133847b0e0b1094c50f263b5cc6dee335dc4/docs/howto/advanced/shutdown.md?plain=1#L40

procrastinate/worker.py
```diff
@@ -338,19 +338,11 @@ async def run(self):
         try:
             # shield the loop task from cancellation
             # instead, a stop event is set to enable graceful shutdown
-            await utils.wait_any(asyncio.shield(loop_task), self._stop_event.wait())
-            if self._stop_event.is_set():
-                try:
-                    await asyncio.wait_for(loop_task, timeout=self.shutdown_timeout)
-                except asyncio.TimeoutError:
-                    pass
+            await asyncio.shield(loop_task)
```
medihack, line 341: Is the comment `instead, a stop event is set to enable graceful shutdown` still true? This line doesn't do anything with the `_stop_event` anymore.

onlyann: The stop event is set by calling `stop()` which happens L344.
It explains why the `loop_task` is being shielded. 

Do you recommend a different wording?

medihack: Never mind, waiting for the event is not necessary here anymore. I didn't quite understand it at the first attempt, but now it seems more clear to me 🙂.

ewjoachim:
There's something bugging me.

If I launch an async task, then stop the worker, the task gets cancelled. When worker is up again, the task won't be retried. Isn't that probably never what the user wants ?
Should be make it a special case that if a task is aborted as part of the shutdown and it would have retried, then we retry it ?

The rationale would be:
- If you, the user, abort a specific task, you don't want it to run
- If you stop a worker, and in order to stop, it aborts its tasks, then you probably still want the task to run, no ?

procrastinate/worker.py
```diff
@@ -424,16 +416,33 @@ async def _shutdown(self, side_tasks: list[asyncio.Task]):
...
 
-        # wait for any in progress job to complete processing
-        # use return_exceptions to not cancel other job tasks if one was to fail
-        await asyncio.gather(*self._running_jobs, return_exceptions=True)
+        if self._running_jobs:
+            await asyncio.wait(
+                self._running_jobs, timeout=self.shutdown_graceful_timeout
+            )
+
+        if self._running_jobs:
```
ewjoachim, line 424:
```suggestion
        # As a reminder, tasks have a done callback that
        # removes them from the self._running_jobs dict,
        # so as the tasks stop, this dict will shrink.
        if self._running_jobs:
```

ewjoachim (commented):
I don't understand yet how it works 😅 
In the meantime, here are my first comments (and sorry for the delay)

EDIT: ok, I've understood my confusion.

procrastinate/worker.py
```diff
@@ -424,16 +416,33 @@ async def _shutdown(self, side_tasks: list[asyncio.Task]):
                 extra=self._log_extra(context=None, action="ending_job"),
             )
 
-        # wait for any in progress job to complete processing
-        # use return_exceptions to not cancel other job tasks if one was to fail
-        await asyncio.gather(*self._running_jobs, return_exceptions=True)
+        if self._running_jobs:
+            await asyncio.wait(
+                self._running_jobs, timeout=self.shutdown_graceful_timeout
+            )
```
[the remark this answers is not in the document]
ewjoachim: Ah no, getting mixed with `asyncio.wait_for` sorry

onlyann:
> There's something bugging me.
> 
> If I launch an async task, then stop the worker, the task gets cancelled. When worker is up again, the task won't be retried. Isn't that probably never what the user wants ? Should be make it a special case that if a task is aborted as part of the shutdown and it would have retried, then we retry it ?
> 
> The rationale would be:
> 
> * If you, the user, abort a specific task, you don't want it to run
> * If you stop a worker, and in order to stop, it aborts its tasks, then you probably still want the tasks to run, no ?

I agree

ewjoachim:
Hm, but then we're breaking one more thing (`should_abort`) Do you think we could rather add `get_abort_reason`, so that existing code using `should_abort` continues to work? I think that it's going to be extremely rare that users will actually want to do something with this information, so maybe it's worth trying to make it easier?

If we have:
```python
def get_abort_reason(self) -> AbortReason | None:
	...
```
then we can have
```python
def should_abort(self):
	return bool(self.get_abort_reason())
```
It will work the same internally, no code gets broken, and users who want the info can get it.

onlyann:
Yes, we can do that.
What I didn't like about the separate method is that is allows states such as `should_abort` being `False` but `get_abort_reason` returns a reason.

medihack:
What should sync jobs do when they get a `should_abort` with `AbortReason.SHUTDOWN`? Should they still raise a `JobAborted` exception? It seems the job will not be retried then (not sure about that).

ewjoachim:
> What I don't like about the separate method is that is allows states such as 
> - `should_abort` is `False` but `get_abort_reason` returns a reason.
> - `should_abort` is `True` but `get_abort_reason` is `None`
> 

People only need to call either method, so they probably won't find themselves in this situation, _I guess_.

> It is a breaking change but that would only impact code such as `if should_abort() is True`. Code written in the more standard way `if should_abort()` will continue to work.

Yes, I agree. But it still means we have to document it, so people will still need to check all of their code.

There can be a variety of situation where they did normal things and it's somehow broken:
- if they store `should_abort` in a variable or use it in a function call, pyright will probably complain
- if for some reason they don't need to compare `should_abort` to the literal True or False but to a variable that may contain True or False, then they may write `if should_abort() is my_variable` which will break too.

I think having a function named `should_x` that doesn't return a boolean is a code smell.

onlyann:
> What should sync jobs do when they get a `should_abort` with `AbortReason.SHUTDOWN`? Should they still raise a `JobAborted` exception? It seems the job will not be retried then (not sure about that).

That would retry (but only when caused by a shutdown). 
I have written a test for that specific scenario.

onlyann:
> I think having a function named should_x that doesn't return a boolean is a code smell.

This is what I don't quite like about it.

onlyann:
I'll refactor it to be two separate methods.

PR #1329, opened 2025-02-16
Thread safe In Memory Connector

### LINKED ISSUE #1320: Make InMemoryConnector Thread-Safe for Multi-Threaded FastAPI Use Cases
## Description

When using procrastinate.testing.InMemoryConnector with FastAPI and synchronous routes, an issue arises due to thread-safety violations in asyncio operations. The in-memory connector, while useful for testing, assumes that all operations occur on the same event loop. However, when used in a FastAPI sync route, which executes in a separate worker thread, the connector’s notification system (_notify) tries to interact with an asyncio event loop running in a different thread, leading to a RuntimeError.

## Context: Using Procrastinate with FastAPI Sync Routes
- FastAPI sync routes are executed in a thread pool by default.
- procrastinate.App and its worker are typically running on the main event loop.
- When a sync route defers a task (task.defer()), procrastinate processes it asynchronously.
- The InMemoryConnector calls _notify, which interacts with an asyncio.Event from the wrong thread.
- Since uvloop (default event loop for FastAPI) is not thread-safe, this leads to a non-thread-safe operation error.

## Error Message

When deferring a task from a synchronous route, the following error occurs:
`RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one`

Full traceback:

```
File "/path/to/procrastinate/testing.py", line 195, in _notify
    await self.on_notification(
File "/path/to/procrastinate/worker.py", line 379, in _handle_notification
    self._new_job_event.set()
File "uvloop/loop.pyx", line 1279, in uvloop.loop.Loop.call_soon
File "uvloop/loop.pyx", line 715, in uvloop.loop.Loop._check_thread
RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one
```

## Cause
1. The sync route runs in a separate thread from the main event loop.
1. When task.defer() is called, it eventually triggers _notify in InMemoryConnector.
1. _notify calls an async function from a different thread than the main event loop.
1. uvloop (default FastAPI event loop) does not allow thread-unsafe operations.
1. The RuntimeError occurs when _notify tries to await an event notification from the wrong thread.

## Proposed Fix

I propose a thread-safe version of InMemoryConnector that:
- Saves the event loop and thread ID at startup.
- Ensures all notifications run on the correct event loop.
- Uses asyncio.run_coroutine_threadsafe when necessary to safely execute _notify.

```python
import asyncio
import json
import threading
import procrastinate
from procrastinate import testing


class InMemoryConnectorThreadSafe(testing.InMemoryConnector):
    async def open_async(
        self, pool: procrastinate.connector.Pool | None = None
    ) -> None:
        """
        Save the current event loop and its thread id so that later notifications
        can be scheduled on this loop.
        """
        self._loop = asyncio.get_running_loop()
        self._loop_thread_id = threading.get_ident()
        self.states.append("open_async")

    async def _notify(
        self, queue_name: str, notification: procrastinate.jobs.Notification
    ) -> None:
        """
        Instead of directly awaiting on_notification, we check the current thread.
        If we’re not on the same thread as the one where the loop was saved,
        we schedule the notification on the correct loop.
        """
        if not self.on_notification:
            return

        destination_channels = {
            "procrastinate_any_queue_v1",
            f"procrastinate_queue_v1#{queue_name}",
        }
        for channel in set(self.notify_channels).intersection(destination_channels):
            coro = self.on_notification(
                channel=channel, payload=json.dumps(notification)
            )
            if threading.get_ident() == self._loop_thread_id:
                # Already on the right thread: just await.
                await coro
            else:
                # Not on the correct thread: schedule the coroutine on the saved loop.
                future = asyncio.run_coroutine_threadsafe(coro, self._loop)
                # Wrap the concurrent.futures.Future so we can await it.
                await asyncio.wrap_future(future)
```

Proposed Integration

This could be merged into InMemoryConnector to improve its usability with multi-threaded FastAPI applications.
Suggested changes:
	•	Modify InMemoryConnector to include the _notify fix.
	•	Ensure proper testing for multi-threaded environments.

## Why This Fix?

✅ Prevents non-thread-safe calls to the event loop.
✅ Makes InMemoryConnector compatible with FastAPI sync routes.
✅ Keeps the fix lightweight, only modifying the _notify behavior.
✅ Does not affect the existing usage of InMemoryConnector in single-threaded tests.

## Final Thoughts

Thanks to the Procrastinate team for the awesome work on the library! This fix would make InMemoryConnector much more robust for testing with FastAPI, particularly for teams that mix sync and async routes.
Would love to hear your thoughts on integrating this fix! 🚀

romainrey:
Closes #1320

### Successful PR Checklist:
- [ ] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - More info on issue #1320

#### PR label(s): 

### DIFF
--- a/procrastinate/testing.py
+++ b/procrastinate/testing.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import asyncio
 import datetime
 import json
+import threading
 from collections import Counter
 from collections.abc import Iterable
 from itertools import count
-import threading
 from typing import Any
+
 from procrastinate import connector, exceptions, jobs, schema, sql, types, utils
 
 JobRow = dict[str, Any]
@@ -83,9 +84,7 @@ class InMemoryConnector(connector.BaseAsyncConnector):
     def open(self, pool: connector.Pool | None = None) -> None:
         self.states.append("open")
 
-    async def open_async(
-        self, pool: connector.Pool | None = None
-    ) -> None:
+    async def open_async(self, pool: connector.Pool | None = None) -> None:
         """
         Save the current event loop and its thread id so that later notifications
         can be scheduled on this loop.
@@ -191,12 +190,10 @@ class InMemoryConnector(connector.BaseAsyncConnector):
             if job["status"] in {"failed", "succeeded"}
         ]
 
-    async def _notify(
-        self, queue_name: str, notification: jobs.Notification
-    ) -> None:
+    async def _notify(self, queue_name: str, notification: jobs.Notification) -> None:
         """
         Instead of directly awaiting on_notification, we check the current thread.
-        If we’re not on the same thread as the one where the loop was saved,
+        If we are not on the same thread as the one where the loop was saved,
         we schedule the notification on the correct loop.
         """
         if not self.on_notification:
--- a/procrastinate/testing.py
+++ b/procrastinate/testing.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
+import asyncio
 import datetime
 import json
 from collections import Counter
 from collections.abc import Iterable
 from itertools import count
+import threading
 from typing import Any
-
 from procrastinate import connector, exceptions, jobs, schema, sql, types, utils
 
 JobRow = dict[str, Any]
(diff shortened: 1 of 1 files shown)


ewjoachim (dismissed):
Thank you for this :) It probably would have take quite a while to find and fix it if it wasn't for dedicated contributors like yourself :)

ewjoachim:
You have set your PR so as not to enable maintainers to contribute it, so here's the patch that, I think, would fix the CI:
```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index f361ea28..2f41d32e 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -258,8 +258,8 @@ def not_opened_app(connector, reset_builtin_task_names) -> app_module.App:


 @pytest.fixture
-def app(not_opened_app: app_module.App):
-    with not_opened_app.open() as app:
+async def app(not_opened_app: app_module.App):
+    async with not_opened_app.open_async() as app:
         yield app
```

I'll let you apply it :)

PR #1343, opened 2025-03-10
Only log Error level if all retry attempts fail( #1342)

### LINKED ISSUE #1342: Tasks that have retries left still fail logging Error instead of Info
Hi All, 

I noticed that we are now logging an Error every time we fail the task even when retries are left. 
Before the upgrade to v3, it used to log Info level first and only logged Error level if all retry attempts failed. 
It is creating a lot of Error logs. 

I saw there was a similar pull request made before the upgrade to V3:
https://github.com/procrastinate-org/procrastinate/issues/708

I think we can simply change the one line in worker.py line 197:
`        log_level = logging.ERROR if status == jobs.Status.FAILED **and not job_retry** else logging.INFO
`

Is it something we can consider to change? 

Many thanks!

minjicho-support:
Closes #1342

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [x] (not applicable?)

#### PR label(s): 

### DIFF
--- a/tests/unit/test_worker.py
+++ b/tests/unit/test_worker.py
@@ -645,7 +645,10 @@ async def test_run_job_retry_failed_job(
     recover_on_attempt_number,
     expected_status,
     expected_attempts,
+    caplog,
 ):
+    caplog.set_level("INFO")
+
     worker.wait = False
 
     attempt = 0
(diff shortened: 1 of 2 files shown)


tests/unit/test_worker.py
```diff
@@ -668,6 +671,28 @@ def task_func():
...
+    assert (
+        len(
+            [
+                record
+                for record in caplog.records
+                if record.levelname == "INFO" and "to retry" in record.message
+            ]
+        )
+        == 1
+    )
```
ewjoachim, line 683: I'm a believe that the `==1` line produced by the formatter is an indication that the code would be more readable with an intermediary variable=
```python
info_records = [...]
assert len(info_records) == 1
```

Usually in those case, the code ends up being more readable AND less lines :)
(same below)

tests/unit/test_worker.py
```diff
@@ -668,6 +671,28 @@ def task_func():
...
+
+    if job_row["status"] == "failed":
```
ewjoachim, line 685: I _tend_ to dislike `if` statements in tests.
Rule of thumb would be:
- Either there's a really good reason for the test to have multiple branches (I don't think it's the case here)
- Or you're trying to do too much with a single parametrized test that should be 2 distinct tests (that might be a solution, but I like the next solution better)
- Or you should add one more parametrize parameter (such as `count_error_logs`)(I believe it's the case here).

ewjoachim (commented):
Good idea :) 2 small comments

tests/unit/test_worker.py
minjicho-support: Hi @ewjoachim Thank you very much for the comment. I have addressed your comments in my new commit.

ewjoachim (approved):

PR #1349, opened 2025-03-21
Fix (probably) incorrect test matrix.

TkTech:
The way the test matrix is currently setup doesn't test against all supported combinations of python versions and postgres versions, but rather just in sets of {py3.9, pg13}, {py3.10, pg14}, {py3.11, pg15}, etc, which is most likely not intentional. This change ensure CI/CD tests against the full range.

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,17 +15,8 @@ jobs:
   tests:
     strategy:
       matrix:
-        include:
-          - python-version: "3.9"
-            postgres-version: "13"
-          - python-version: "3.10"
-            postgres-version: "14"
-          - python-version: "3.11"
-            postgres-version: "15"
-          - python-version: "3.12"
-            postgres-version: "16"
-          - python-version: "3.13"
-            postgres-version: "17"
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        postgres-version: ["13", "14", "15", "16", "17"]
 
     name: "py${{ matrix.python-version }}"
     runs-on: ubuntu-latest


ewjoachim:
This is completely intentional, the idea is that all supported python versions and pg versions should be tested at least once, but unless we have specific reasons to believe a given combination will fail, we avoid the quadratic explosion of the test matrix for various reasons:

- It's extremely unlikely that one combination will fail. As much as I can remember, it's always been 1 or more rows or one or more columns of the matrix that fail.
- This would be 25 tests instead of 5, so 5 times more costly. Even if this seems free, the cost exists:
	- Microsoft pays for this and the more we use it, the more incentive they will have to eventually rescind the free offer
	- These are real machine using real energy and producing real CO2 on the planet somewhere
	- GitHub limits to 20 concurrent jobs on free projects, so if we have 25 jobs, this will double the perceived CI time (as it will need 2 rounds).
- Some of the cells from the matrix may be unsupported and it will be equally painful to maintain the special cases

> [!NOTE]
> it's the same with Django: we don't test all Django versions with all Python versions with all postgres version, otherwise we would have even more tests.

> [!NOTE]
> The kind of issues we may have from mixing very old and very new versions of Python and PG are very likely to be issues from Django or Psycopg, and even if people encountered bugs, there's a lot of cases where we can only direct them to the appropriate tracker.

Given that, there are multiple possible strategies:
- test old version of Python and PG and new versions of Python and PG together (test a diagonal of the matrix). Pro: most likely to get the best compatibility. Con: less likely to find deprecation issues due to support being suboptimal for extreme differences in versions.
- test newest versions of Python with oldest versions of PG and vice versa. Pro: more likely to find bugs. Con: Django 5.2 only supports PG 14+, so it will be harder to setup (and in general, we'll keep running against these kind of problems).
- make random pairings. Whether it's randomly generated pairing different at every run or statically random, I'm not a fan.

With all that said, I'm curious if you have opinions going the other way :)

ewjoachim:
One this that's true is that this could deserve a couple lines (and maybe a link to this issue) on the CONTRIBUTING doc :)

ewjoachim:
I'm ok with discussing about how to improve our practice.

I'm not sure testing at release time is the best, as this can make releasing a bit more complex (in the event the test fails). On the other hand, as long as we test on the oldest python+oldest PG and newest python + newest PG on the PRs, I'm pretty confident that any other combination will fail extremely rarely, so there won't be a lot of cases where release is actually made harder.

That said, what should be tested ? I feel we should test any version of the software big enough to have official support policies, which should in most cases overlap pretty well with anything present on https://endoflife.date, in other words we should probably test Django too, which multiply tests by 3.

PR #1382, opened 2025-05-04
Allow to batch defer multiple jobs at once

### LINKED ISSUE #1284: Batch queue tasks?
Could there be a way to queue a set of tasks at once rather than sequentially calling `defer` or `defer_async` (which perform `INSERT`s one at a time - ends up being quite slow)

It could look like
```python
taskname.defer_batch_async([{"arg1":"val1"},{"arg2":"val2"}])
```

medihack:
Closes #1284 

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -158,11 +158,6 @@ class JobDeferrer:
 
         return self.job.evolve(task_kwargs=final_kwargs)
 
-    def make_multiple_new_jobs(
-        self, task_kwargs_list: list[dict[str, types.JSONValue]]
-    ) -> list[Job]:
-        return [self.make_new_job(**task_kwargs) for task_kwargs in task_kwargs_list]
-
     def _log_before_defer_job(self, job: Job) -> None:
         logger.debug(
             f"About to defer job {job.call_string}",
@@ -209,7 +204,7 @@ class JobDeferrer:
         Defer multiple jobs at once. This is more efficient than deferring them one by
         one, as it only requires one database query.
         """
-        jobs = self.make_multiple_new_jobs(task_kwargs_list)
+        jobs = [self.make_new_job(**task_kwargs) for task_kwargs in task_kwargs_list]
         self._log_before_batch_defer_jobs(jobs=jobs)
         jobs = await self.job_manager.batch_defer_jobs_async(jobs=jobs)
         self._lob_after_batch_defer_jobs(jobs=jobs)
@@ -236,7 +231,7 @@ class JobDeferrer:
         Defer multiple jobs at once. This is more efficient than deferring them one by
         one, as it only requires one database query.
         """
-        jobs = self.make_multiple_new_jobs(task_kwargs_list)
+        jobs = [self.make_new_job(**task_kwargs) for task_kwargs in task_kwargs_list]
         self._log_before_batch_defer_jobs(jobs=jobs)
         jobs = self.job_manager.batch_defer_jobs(jobs=jobs)
         self._lob_after_batch_defer_jobs(jobs=jobs)
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -171,8 +171,9 @@ class JobDeferrer:
         )
 
     def _log_before_batch_defer_jobs(self, jobs: list[Job]) -> None:
+        job_count = len(jobs)
         logger.debug(
-            f"About to defer {len(jobs)} jobs",
+            f"About to defer {job_count} {'job' if job_count == 1 else 'jobs'}",
             extra={
                 "action": "about_to_defer_jobs",
                 "jobs": [job.log_context() for job in jobs],
@@ -180,8 +181,9 @@ class JobDeferrer:
         )
 
     def _lob_after_batch_defer_jobs(self, jobs: list[Job]) -> None:
+        job_count = len(jobs)
         logger.info(
-            f"Deferred {len(jobs)} jobs",
+            f"Deferred {job_count} {'job' if job_count == 1 else 'jobs'}",
             extra={"action": "jobs_defer", "jobs": [job.log_context() for job in jobs]},
         )
 
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -189,7 +189,7 @@ class JobDeferrer:
 
     async def defer_async(self, **task_kwargs: types.JSONValue) -> int:
         """
-        See `Task.defer` for details.
+        See `Task.defer_async` for details.
         """
         # Make sure this code stays synchronized with .defer()
         job = self.make_new_job(**task_kwargs)
@@ -199,14 +199,11 @@ class JobDeferrer:
         assert job.id  # for mypy
         return job.id
 
-    async def batch_defer_async(
-        self, task_kwargs_list: list[dict[str, types.JSONValue]]
-    ) -> list[int]:
+    async def batch_defer_async(self, *task_kwargs: types.JSONDict) -> list[int]:
         """
-        Defer multiple jobs at once. This is more efficient than deferring them one by
-        one, as it only requires one database query.
+        See `Task.batch_defer_async` for details.
         """
-        jobs = [self.make_new_job(**task_kwargs) for task_kwargs in task_kwargs_list]
+        jobs = [self.make_new_job(**kwargs) for kwargs in task_kwargs]
         self._log_before_batch_defer_jobs(jobs=jobs)
         jobs = await self.job_manager.batch_defer_jobs_async(jobs=jobs)
         self._lob_after_batch_defer_jobs(jobs=jobs)
@@ -218,6 +215,9 @@ class JobDeferrer:
         return job_ids
 
     def defer(self, **task_kwargs: types.JSONValue) -> int:
+        """
+        See `Task.defer` for details.
+        """
         # Make sure this code stays synchronized with .defer_async()
         job = self.make_new_job(**task_kwargs)
         self._log_before_defer_job(job=job)
@@ -226,14 +226,11 @@ class JobDeferrer:
         assert job.id  # for mypy
         return job.id
 
-    def batch_defer(
-        self, task_kwargs_list: list[dict[str, types.JSONValue]]
-    ) -> list[int]:
+    def batch_defer(self, *task_kwargs: types.JSONDict) -> list[int]:
         """
-        Defer multiple jobs at once. This is more efficient than deferring them one by
-        one, as it only requires one database query.
+        See `Task.batch_defer` for details.
         """
-        jobs = [self.make_new_job(**task_kwargs) for task_kwargs in task_kwargs_list]
+        jobs = [self.make_new_job(**kwargs) for kwargs in task_kwargs]
         self._log_before_batch_defer_jobs(jobs=jobs)
         jobs = self.job_manager.batch_defer_jobs(jobs=jobs)
         self._lob_after_batch_defer_jobs(jobs=jobs)
--- a/procrastinate/jobs.py
+++ b/procrastinate/jobs.py
@@ -158,6 +158,11 @@ class JobDeferrer:
 
         return self.job.evolve(task_kwargs=final_kwargs)
 
+    def make_multiple_new_jobs(
+        self, task_kwargs_list: list[dict[str, types.JSONValue]]
+    ) -> list[Job]:
+        return [self.make_new_job(**task_kwargs) for task_kwargs in task_kwargs_list]
+
     def _log_before_defer_job(self, job: Job) -> None:
         logger.debug(
             f"About to defer job {job.call_string}",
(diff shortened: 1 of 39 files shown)


procrastinate/sql/schema.sql
```diff
@@ -86,30 +86,52 @@ CREATE INDEX procrastinate_periodic_defers_job_id_fkey_v1 ON procrastinate_perio
...
-    scheduled_at timestamp with time zone
+CREATE FUNCTION procrastinate_defer_jobs_v1(
+    queue_names character varying[],
+    task_names character varying[],
+    priorities integer[],
+    locks text[],
+    queueing_locks text[],
+    args_list jsonb[],
+    scheduled_ats timestamp with time zone[]
 )
```
ewjoachim, line 97: I've never playd with that, but I believe you can create your own PG composite types, so instead of arrays of values (where you wish all arrays have the same length), you could define a single array of composite values

https://www.postgresql.org/docs/current/rowtypes.html

procrastinate/jobs.py
```diff
@@ -190,3 +228,21 @@ def defer(self, **task_kwargs: types.JSONValue) -> int:
         self._log_after_defer_job(job=job)
         assert job.id  # for mypy
         return job.id
+
+    def batch_defer(
```
medihack: @ewjoachim As discussed on Discord, it would be possible to reuse `defer` resp. `defer_async` in `jobs.py` to also defer multiple jobs with something like this:
```python
@overload
def defer(self, *task_kwargs: types.JSONDict) -> list[int]: ...
@overload
def defer(self, **single_task_kwargs: types.JSONValue) -> int: ...
def defer(
	self, *task_kwargs: types.JSONDict, **single_task_kwargs: types.JSONValue
) -> int | list[int]:
	# Make sure this code stays synchronized with .defer_async()

	if task_kwargs and single_task_kwargs:
		raise ValueError(
			"Cannot set both kwargs for a single job and multiple jobs"
		)

	if task_kwargs:
		jobs = [self.make_new_job(**kwargs) for kwargs in task_kwargs]
		self._log_before_batch_defer_jobs(jobs=jobs)
		jobs = self.job_manager.batch_defer_jobs(jobs=jobs)
		self._lob_after_batch_defer_jobs(jobs=jobs)

		job_ids: list[int] = []
		for job in jobs:
			assert job.id
			job_ids.append(job.id)
		return job_ids

	job = self.make_new_job(**single_task_kwargs)
	self._log_before_defer_job(job=job)
	job = self.job_manager.defer_job(job=job)
	self._log_after_defer_job(job=job)
	assert job.id  # for mypy
	return job.id
```
Unfortunately, this would break the type hints of `defer` and `defer_async` in `tasks.py`, which use `ParamSpec`. And from my investigations, it doesn't seem to be possible to extract the type information of kwargs from `ParamSpec` or overload those methods there.
So, to avoid losing the type hints when calling defer, I would keep a separate `batch_defer` method.

procrastinate/sql/schema.sql
medihack: Yes, good idea. Unfortunately, it is not so easy to realise as the composite types would have to be registered in the connectors, see [psycopg doc](https://www.psycopg.org/psycopg3/docs/basic/pgtypes.html). I am a bit hesitant to create such a dependency between the connectors and the schema (especially as the current implementation works just fine). What do you think?

medihack: I was wrong ... registering component types only seems to be relevant when returning a result. I'll try again using component types.

medihack: Ok, finally got it working 😁.

procrastinate/manager.py
```diff
@@ -123,13 +123,18 @@ def batch_defer_jobs(self, jobs: list[jobs_module.Job]) -> list[jobs_module.Job]
...
+            "jobs": [
+                (
+                    job.queue,
+                    job.task_name,
+                    job.priority,
+                    job.lock,
+                    job.queueing_lock,
+                    job.task_kwargs,
+                    job.scheduled_at,
+                )
```
ewjoachim, line 135: An namedtuple is a tuple, so you could even use a namedtuple and kwargs and it should still work :)

procrastinate/testing.py
```diff
@@ -5,12 +5,21 @@
...
 from collections.abc import Iterable
 from itertools import count
-from typing import Any
+from typing import Any, Union
 
 from procrastinate import connector, exceptions, jobs, schema, sql, types, utils
 
 JobRow = dict[str, Any]
 EventRow = dict[str, Any]
+JobToDefer = tuple[
```
ewjoachim, line 14: https://docs.python.org/3/library/typing.html#typing.NamedTuple

procrastinate/manager.py
medihack: Oh yes, thanks. Makes it much clearer.

procrastinate/testing.py
```diff
@@ -94,55 +94,61 @@ async def close_async(self) -> None:
...
+        for job in jobs:
+            queueing_lock = job.queueing_lock
+            if queueing_lock is not None and (
+                any(
+                    job["queueing_lock"] == queueing_lock and job["status"] == "todo"
+                    for job in self.jobs.values()
+                )
+                or queueing_lock in seen_queueing_locks
+            ):
+                from . import manager
```
medihack, line 111: Why is the dynamic import here necessary?

procrastinate/manager.py
```diff
@@ -51,37 +52,89 @@ async def defer_job_async(self, job: jobs.Job) -> jobs.Job:
...
+
+    def batch_defer_jobs(self, jobs: list[jobs_module.Job]) -> list[jobs_module.Job]:
+        """
+        Sync version of `batch_defer_jobs_async`.
+        """
+        try:
+            results = self.connector.get_sync_connector().execute_query_all(
+                **self._defer_jobs_query_kwargs(jobs=jobs)
+            )
+        except exceptions.UniqueViolation as exc:
```
medihack, line 115: Coverage tells me that L115-116 are not hit. I wonder why that is, because it should be tested by https://github.com/procrastinate-org/procrastinate/blob/95741d55563736ab02b3164066c5a65c7091b326/tests/integration/test_manager.py#L657

procrastinate/testing.py
medihack: Yes, doesn't seem to be necessary anymore. Removed it.

medihack:
I also added some benchmarks; of course, those batched ones are much faster. Strangely, even the other non-batch benchmarks are faster now (if I understood the results correctly) 🚀. (I will link the benchmarks in our FAQ as a separate PR after this one is merged).

procrastinate/testing.py
```diff
@@ -94,55 +103,59 @@ async def close_async(self) -> None:
...
+                    job["queueing_lock"] == queueing_lock and job["status"] == "todo"
+                    for job in self.jobs.values()
+                )
+                or queueing_lock in seen_queueing_locks
+            ):
+                raise exceptions.UniqueViolation(
+                    constraint_name=manager.QUEUEING_LOCK_CONSTRAINT
+                )
+
+            seen_queueing_locks.add(queueing_lock)
```
ewjoachim, line 124:
```suggestion
        current_queuing_locks = {
            job["queueing_lock"]
            for job in self.jobs.values()
            if job["status"] == "todo"
        } - {"None}
        
        new_queueing_locks = [job.queueing_lock for job in jobs]
        new_queueing_locks_set = set(new_queueing_locks)
        if (
            len(new_queueing_locks) != len(new_queueing_locks_set)
            or current_queuing_locks & current_queuing_locks
        ):
                raise exceptions.UniqueViolation(
                    constraint_name=manager.QUEUEING_LOCK_CONSTRAINT
                )
```

Not sure it's more readable, what do you think ?

tests/integration/test_manager.py
```diff
@@ -624,6 +676,43 @@ async def test_defer_job_violate_queueing_lock(pg_job_manager, job_factory):
...
+                    queueing_lock="queueing_lock",
+                    task_kwargs={"c": "d"},
+                ),
+            ]
+        )
+    cause = excinfo.value.__cause__
+    assert isinstance(cause, exceptions.UniqueViolation)
+
+    # TODO: When QUEUEING_LOCK_CONSTRAINT_LEGACY in manager.py is removed, we can
+    # also remove the check for the old constraint name "procrastinate_jobs_queueing_lock_idx"
```
ewjoachim, line 707: Good thinking :o

ewjoachim (approved):
Impressive work 👏 👏 👏 

I wish I had more smart and insightful remarks, but I read from top to bottom, and it all looked great. You had some serious patience fixing all the tests the good way. I feel it must have been quite annoying, do you think there are changes we should make so that future PRs like this are easier ? Or it's just that sometimes one must take the long way through ?

procrastinate/manager.py
```diff
@@ -35,7 +36,7 @@ class JobManager:
     def __init__(self, connector: connector.BaseConnector):
         self.connector = connector
 
-    async def defer_job_async(self, job: jobs.Job) -> jobs.Job:
+    async def defer_job_async(self, job: jobs_module.Job) -> jobs_module.Job:
```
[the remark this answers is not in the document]
medihack: Do you mean having only `batch_defer_jobs` and `batch_defer_jobs_async` and getting rid of `defer_job` and `defer_jobs_async`? To me, `defer_job` and `defer_jobs_async` seem to be public APIs. I would also keep it that way because we differentiate between `batch_defer_async` and `defer_async` in jobs.py and tasks.py. But I may have understood your comment wrong.

tests/integration/test_manager.py
medihack: Just copied it from somewhere else 😉.

medihack:
> Impressive work 👏 👏 👏
> 
> I wish I had more smart and insightful remarks, but I read from top to bottom, and it all looked great. You had some serious patience fixing all the tests the good way. I feel it must have been quite annoying, do you think there are changes we should make so that future PRs like this are easier ? Or it's just that sometimes one must take the long way through ?

Not that annoying. I just changed one case and let Copilot figure out the rest of them 😁. If one has half a clue about the subject matter, then this AI stuff is really amazing and a game changer. So I would say let's keep it that way. The tests, as they are, have proven their worth (we have so few bugs).

procrastinate/manager.py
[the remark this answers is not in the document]
ewjoachim: Ah sorry, it was probably unclear.

Today, `defer_job_async` calls `self._defer_jobs_query_kwargs` under the hood. Would it be simpler if it called `batch_defer_jobs_async` under the hood ?

I'm not completely sure it makes sense, but it _feels_ that way.

procrastinate/testing.py
```diff
@@ -107,21 +107,23 @@ async def defer_jobs_all(self, jobs: list[types.JobToDefer]) -> list[JobRow]:
...
+        } - {None}
 
-            seen_queueing_locks.add(queueing_lock)
+        new_queueing_locks = [
+            job.queueing_lock for job in jobs if job.queueing_lock is not None
+        ]
+        new_queueing_locks_set = set(new_queueing_locks)
+        if (
+            len(new_queueing_locks) != len(new_queueing_locks_set)
+            or current_queuing_locks & new_queueing_locks_set
```
ewjoachim, line 122: (ah good, you spotted and fixed the typo in my suggestion that I just saw :D 👏  Sorry about that)

ewjoachim (approved):

procrastinate/manager.py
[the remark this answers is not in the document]
medihack: Ah, yes, it absolutely makes sense now. And it was even extremely helpful in another way. I added a `TODO:` myself and forgot about it. Because I have the problem that when I defer multiple jobs, I can't say for sure anymore which queuing lock is the source of the `UniqueViolation`. This information is easy to provide by the InMemoryConnector, but I am not so sure about the PostgreSQL connectors. Currently, `UniqueViolation` only provides the `constraint_name` (the idx), but not the name of the queueing lock that led to it.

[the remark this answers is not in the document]
medihack: Yes yes, I am already on it 😁. It seems we only handle `psycopg2.errors.UniqueViolation` and `psycopg.errors.UniqueViolation` (aiopg also uses `psycopg2.errors.UniqueViolation` and DjangoConnector is client only).

procrastinate/testing.py
medihack: > (ah good, you spotted and fixed the typo in my suggestion that I just saw :D 👏 Sorry about that)

No, prob. It's good that occasionally I also have to switch my brain on 😉. Unfortunately, I had to rewrite it because of how the queueing lock name is now handled.

procrastinate/manager.py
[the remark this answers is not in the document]
medihack: All right, I hope this works now in all scenarios.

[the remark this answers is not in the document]
medihack: Just saw that the coverage went down a bit because of the connectors 😞. I will try to improve that.

procrastinate/testing.py
```diff
@@ -107,22 +107,31 @@ async def defer_jobs_all(self, jobs: list[types.JobToDefer]) -> list[JobRow]:
...
         new_queueing_locks = [
             job.queueing_lock for job in jobs if job.queueing_lock is not None
         ]
-        new_queueing_locks_set = set(new_queueing_locks)
-        if (
-            len(new_queueing_locks) != len(new_queueing_locks_set)
-            or current_queuing_locks & new_queueing_locks_set
-        ):
+
+        counts = Counter(new_queueing_locks)
```
ewjoachim, line 114: 👍

procrastinate/manager.py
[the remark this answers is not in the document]
medihack: Done

ewjoachim (approved):
🎉

ewjoachim:
> with your help

Nice of you, but you really did all the work !

PR #1425, opened 2025-06-30
fix: try to reconnect on exception

woolfred:
Hey 👋 

We stumbled over problems, when loosing the connection to the database.
After connection loss (stopping and restarting our db), procrastinate continued to work, but only via polling.

This is just the simplest approach trying to reestablish the listen for new jobs.
Let's discuss how we can finish/polish this, if this should be part of the library :) 

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/procrastinate/contrib/aiopg/aiopg_connector.py
+++ b/procrastinate/contrib/aiopg/aiopg_connector.py
@@ -37,7 +37,8 @@ async def wrap_exceptions() -> AsyncGenerator[None, None]:
         queueing_lock = None
         if constraint_name == manager.QUEUEING_LOCK_CONSTRAINT:
             assert exc.diag.message_detail
-            match = re.search(r"Key \((.*?)\)=\((.*?)\)", exc.diag.message_detail)
+            match = re.search(r"Key \((.*?)\)=\((.*?)\)",
+                              exc.diag.message_detail)
             assert match
             column, queueing_lock = match.groups()
             assert column == "queueing_lock"
@@ -180,7 +181,8 @@ class AiopgConnector(connector.BaseAsyncConnector):
             if base_on_connect:
                 await base_on_connect(connection)
             if json_loads:
-                psycopg2.extras.register_default_jsonb(connection.raw, loads=json_loads)
+                psycopg2.extras.register_default_jsonb(
+                    connection.raw, loads=json_loads)
 
         final_args = {
             "dsn": "",
@@ -307,7 +309,7 @@ class AiopgConnector(connector.BaseAsyncConnector):
         on_notification: connector.Notify,
         channels: Iterable[str],
         *,
-        listen_notify_reconnect_interval: float = 2.0,
+        reconnect_interval: float = 2.0,
     ) -> None:
         # We need to acquire a dedicated connection, and use the listen
         # query
--- a/procrastinate/contrib/aiopg/aiopg_connector.py
+++ b/procrastinate/contrib/aiopg/aiopg_connector.py
@@ -326,7 +326,8 @@ class AiopgConnector(connector.BaseAsyncConnector):
                         await self._execute_query_connection(
                             connection=connection,
                             query=self._make_dynamic_query(
-                                query=sql.queries["listen_queue"], channel_name=channel_name
+                                query=sql.queries["listen_queue"],
+                                channel_name=channel_name,
                             ),
                         )
                     await self._loop_notify(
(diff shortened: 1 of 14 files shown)


ewjoachim (commented):
Excellent idea. Seems like a great start :)

ewjoachim:
This is the kind of change that might nod need extensive test, but if you manage to get a test that covers the new lines, feel free. You may want to make the sleep time a parameter of the function so that in the tests we don't need to actually sleep for 2s (this is the kind of things that really slows down the test suite in the long run)

Maybe something else to do would be to ensure that if a worker does die, it doesn't die silently, but that's ok to do in another PR

ewjoachim:
Yep, my change was just suggested for making better tests, but good tests are worth adjustements in code. That said, yes, raising this param all the way up makes sense, I'm perfectly fine doing it :)

ewjoachim (commented):
Hm, do you think it would make sense to define it as an argument of `listen_notify` rather than an argument of the connector, and then passed in from the worker, which would receive it from the app like [fetch_job_polling_interval](https://github.com/procrastinate-org/procrastinate/blob/main/procrastinate/app.py#L275) ?

woolfred:
> Hm, do you think it would make sense to define it as an argument of `listen_notify` rather than an argument of the connector, and then passed in from the worker, which would receive it from the app like [fetch_job_polling_interval](https://github.com/procrastinate-org/procrastinate/blob/main/procrastinate/app.py#L275) ?

Oh, sure. Looks like a much better place with all the other configs.

procrastinate/contrib/aiopg/aiopg_connector.py
```diff
@@ -303,7 +303,11 @@ def _make_dynamic_query(self, query: str, **identifiers: str) -> Any:
 
     @wrap_exceptions()
     async def listen_notify(
-        self, on_notification: connector.Notify, channels: Iterable[str]
+        self,
+        on_notification: connector.Notify,
+        channels: Iterable[str],
+        *,
+        reconnect_interval: float = 2.0,
```
ewjoachim, line 310: It looks like it should be acceptably complex to implement the use of `reconnect_interval` in the AioPG Connector, do you want to do it ?

ewjoachim (commented):
One last detail I saw, but we're getting there :)
(I'll do a full review at some point)

woolfred:
Hey, just wanted to check in, what the usual release cycle is and when you expect this to land :)

ewjoachim:
You're right to check in, this had left my mind entirely, sorry.

ewjoachim:
Released as part of 3.5.0

woolfred:
@ticosax there are no changes for the django_connector, as it does not support listen/notify.
Do you have a setup where another connector is used like psycopg?

woolfred:
@ticosax no we use it in a FastAPI project and configure a connector (PsycopgConnector). 
@ewjoachim the DjangoConnector does look to me, like it does not support listen/notify and therefor didn't had a relevant code change. Do you have a first guess, how any of the changes might have led to en exhaustion of the connection pool? 

We will move to version 3.5.0 soon and then will watch out for similar problems and should be able to investigate more.

woolfred:
@ewjoachim thanks! 
@ticosax can you tell which one are you using for the worker? Do you have `psycopg3` or `aiopg` installed for the worker?

ticosax:
Psycopg3

-------- Message d'origine --------

woolfred:
Ok. So we would also run into this problem. Looks like this needs some more work then.
@ewjoachim Don't know what is the best way till then. A fix will need some time, we don't have right now, so yanking or rollback is probably the best way forward.

ewjoachim:
I'll revert. Feel free to repoen a PR with the fix someday (and I'll try to merge it faster!)(especially if there's a regression test for this issue 😅)

ticosax:
Hi there, unfortunately I don't have access to the logs any longer (90 days retention) and I haven't tried to run our project with this version again. I don't think there is anything specific in our project.
We are using django 5.2 + django con pool + psycopg + 1 single postgres instance db (16.8), and we are constantly running as close as possible the max limit of connections the database can offer, to maximize the throughput. (We experience burst of http traffic sometimes)

That said, I wrote "I suspect ..." as one should  interpret as "maybe", it was not some kind of disguised politeness. I reacted in a rush, reverted the version bump and went on with my life. I did not investigated in depth. maybe it was circumstantial.

I'm sorry I can not do better than this. Maybe some other djangonista wants to give it a try, and share their experience ?
If we are the only one who experienced this behavior, maybe it was a false positive after all. Might worth to take that risk and let other projects give it a try. I'll promise I won't upgrade immediately and wait if someone else reports something. Don't let this testimony be a road block. My confidence is not very high.

PR #1474, opened 2025-11-05
added timezone support for cron schedules in periodic tasks

### LINKED ISSUE #1472: Setting timezone for periodic task
I have a periodic task that's supposed to run every 10 seconds of every weekday from 8 am to 5pm. Its cron expression is thus : `* 8-17 * * 1-5 */10`. I did'nt realize until today that this will run in UTC while I am in Africa/Blantyre (which is 2 hours ahead). To run it in the time I expect, I am simply specifying the hour segment as UTC, i.e., `6-15`.

With the native crontab, I could specify which timezone a particular entry should run in with **CRON_TZ** environment variable. How do I do something similar in this case?

dreackm:
**Closes #1472**

This PR adds native timezone support to Procrastinate’s periodic task scheduler.

The `Blueprint.periodic` decorator now accepts a `tzinfo` argument, which can be either a string or a `zoneinfo.ZoneInfo` object. When provided, croniter will evaluate the cron expression in the specified timezone rather than defaulting to UTC.

Previously, users had to manually convert local times to UTC to ensure tasks ran at the intended hours. This often led to confusion and misaligned schedules. With this enhancement, users can define periodic tasks directly in their local timezone, improving predictability and reducing the risk of scheduling errors.

### Key Changes
- The `tzinfo` attribute is now set on the `croniter` instance within `PeriodicTask`.

### Example Usage
```python
from zoneinfo import ZoneInfo

tzinfo = "Africa/Blantyre" or ZoneInfo("Africa/Blantyre")  # UTC+2

@app.periodic(cron="* 8-17 * * *", tzinfo=tzinfo, ...)
@app.task(...)
def between_5_and_7(timestamp): ...
```

Without `tzinfo`: runs from 10 AM to 7 PM UTC  
With `tzinfo`: runs from 8 AM to 5 PM local time

### Successful PR Checklist:

- [ ] Tests
  - [ ] (not applicable?)
- [ ] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

### DIFF
--- a/procrastinate/blueprints.py
+++ b/procrastinate/blueprints.py
@@ -9,6 +9,7 @@ from typing_extensions import Concatenate, ParamSpec, TypeVar, Unpack
 
 from procrastinate import exceptions, jobs, periodic, retry, utils
 from procrastinate.job_context import JobContext
+from zoneinfo import ZoneInfo
 
 if TYPE_CHECKING:
     from procrastinate.tasks import ConfigureTaskOptions, Task
@@ -195,6 +196,7 @@ class Blueprint:
                 task=periodic_task.task,
                 cron=periodic_task.cron,
                 periodic_id=periodic_task.periodic_id,
+                tzinfo=periodic_task.tzinfo,
                 configure_kwargs=periodic_task.configure_kwargs,
             )
 
@@ -359,6 +361,7 @@ class Blueprint:
         *,
         cron: str,
         periodic_id: str = "",
+        tzinfo: str | None | ZoneInfo = None,
         **configure_kwargs: Unpack[ConfigureTaskOptions],
     ):
         """
@@ -371,11 +374,17 @@ class Blueprint:
             Cron-like string. Optionally add a 6th column for seconds.
         periodic_id :
             Task name suffix. Used to distinguish periodic tasks with different kwargs.
-        **kwargs :
-            Additional parameters are passed to `Task.configure`.
+        tzinfo :
+            Timezone in which the cron expression should be interpreted. Accepts a
+            timezone name string (e.g., "Africa/Blantyre"), a `zoneinfo.ZoneInfo` instance,
+            or `None`. When `None` (the default), the underlying `croniter` library
+            will interpret the schedule in UTC (the current behaviour).
+        **configure_kwargs :
+            Additional parameters are passed to ``Task.configure``.
         """
+
         return self.periodic_registry.periodic_decorator(
-            cron=cron, periodic_id=periodic_id, **configure_kwargs
+            cron=cron, periodic_id=periodic_id, tzinfo=tzinfo, **configure_kwargs
         )
 
     def will_configure_task(self) -> None:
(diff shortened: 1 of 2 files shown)


procrastinate/periodic.py
```diff
@@ -35,10 +36,26 @@ class PeriodicTask(Generic[P, R, Args]):
...
+        return croniter_instance
+
+    def get_tzinfo(self):
+        tzinfo = self.tzinfo
+        if isinstance(tzinfo, str):
+            try:
+                return ZoneInfo(tzinfo)
+            except (ZoneInfoNotFoundError, ValueError):
+                logger.error(f"{tzinfo} is not a valid timezone.")
+                return None
```
ewjoachim: A failure could be impactful, I think we need to raise.

procrastinate/periodic.py
```diff
@@ -79,6 +98,7 @@ def register_task(
         cron: str,
         periodic_id: str,
         configure_kwargs: tasks.ConfigureTaskOptions,
+        tzinfo: str | None | ZoneInfo = None,
```
ewjoachim: (Please put None last in the types)

procrastinate/blueprints.py
```diff
@@ -371,11 +374,17 @@ def periodic(
             Cron-like string. Optionally add a 6th column for seconds.
         periodic_id :
             Task name suffix. Used to distinguish periodic tasks with different kwargs.
-        **kwargs :
-            Additional parameters are passed to `Task.configure`.
+        tzinfo :
+            Timezone in which the cron expression should be interpreted. Accepts a
+            timezone name string (e.g., "Africa/Blantyre"), a `zoneinfo.ZoneInfo` instance,
+            or `None`. When `None` (the default), the underlying `croniter` library
+            will interpret the schedule in UTC (the current behaviour).
```
ewjoachim: The mention `(the current behaviour)` makes sense as a PR comment because we're talking code evolution, but doesn't make sense in the docstring. In 2 years time, what will "current" refer to ?

ewjoachim (commented):
Nice work so far, we're missing tests. Also, I'd prefer we reach a consensus on the issue before merging the PR, as we may decide on a slightly different API to mitigate the risks.

procrastinate/periodic.py
```diff
@@ -35,10 +36,26 @@ class PeriodicTask(Generic[P, R, Args]):
     cron: str
     periodic_id: str
     configure_kwargs: tasks.ConfigureTaskOptions
+    tzinfo: str | None | ZoneInfo = None
```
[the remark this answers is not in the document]
ewjoachim: Ah no, I meant: whether we recieve a str or a ZoneInfo, what we store is a ZoneInfo.

procrastinate/periodic.py
ewjoachim: > The only issue I can see is that it might confuse users when their chosen timezone isn’t applied. Besides that, does it have any other impact?

If I have scheduled batch processing at 12AM while there's no activity, and it happens at 2PM during peak hours, it could easily cause disruption.

leogregianin:
nice work. i'd like to see this in production

Issue #1134, opened 2024-07-30
Django OperationalError: the connection is closed


nicokant:
I have a django application using procrastinate with the following dependencies:
```
python 3.11

psycopg[binary]==3.1.19
django==5.0.6
procrastinate[django]==2.9.1
```
When I start the worker it works fine, but after some time I start getting this error
 
Logs:
```

INFO 2024-07-30 08:42:38,237 worker 21 139709328389952 Job import_dataset[25](id='35') ended with status: Error, to retry, lasted 0.001 s
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 294, in _cursor
    return self._prepare_cursor(self.create_cursor(name))
                                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/postgresql/base.py", line 332, in create_cursor
    cursor = self.connection.cursor()
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/psycopg/connection.py", line 852, in cursor
    self._check_connection_ok()
  File "/usr/local/lib/python3.11/site-packages/psycopg/connection.py", line 485, in _check_connection_ok
    raise e.OperationalError("the connection is closed")
psycopg.OperationalError: the connection is closed

The above exception was the direct cause of the following exception:
Traceback (most recent call last):
 File "/usr/local/lib/python3.11/site-packages/procrastinate/worker.py", line 281, in run_job
    task_result = await await_func(*job_args, **job.task_kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/procrastinate/utils.py", line 108, in sync_to_async
    return await sync.sync_to_async(func)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/asgiref/sync.py", line 468, in __call__
    ret = await asyncio.shield(exec_coro)
[...]

PR #1317, opened 2025-02-03
Worker middleware feature

medihack:
Closes #1292

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20breaking%20%F0%9F%92%A5
  - [x] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20feature%20%E2%AD%90%EF%B8%8F
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20bugfix%20%F0%9F%95%B5%EF%B8%8F
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20miscellaneous%20%F0%9F%91%BE
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20dependencies%20%F0%9F%A4%96
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20documentation%20%F0%9F%93%9A

PR #1555, opened 2026-06-02
contrib/django: auto-close Django DB connections after task execution

### LINKED ISSUE #1106: Error on teardown of test database in Django integration test
When running an integration (sync) test using Pytest in Django I end up with a warning that another session is using the database:
```
adit/core/tests/test_tasks.py::test_process_dicom_task_that_succeeds
  /app:0: PytestWarning: Error when trying to teardown test databases: OperationalError('database "test_postgres" is being accessed by other users\nDETAIL:  There is 1 other session using the database.')

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

This happens as soon as I run the worker, as mentioned in the documentation:
```python
my_app = app.with_connector(app.connector.get_worker_connector())  # type: ignore
my_app.run_worker(wait=False, install_signal_handlers=False, listen_notify=False)
```

It seems the worker leaves some connections open. I tried calling `app.close()` explicitly at the end of the test, but this also doesn't help.

### LINKED ISSUE #1316: Django memory leak?
Hi,

It appears that Django's ORM is leaking quite a lot into RAM, and I'm not exactly sure why I can't find anything about this problem.

Let's start with the fix. Wrapping all my tasks with this (largely inspired by [Django Channels](https://channels.readthedocs.io/en/latest/topics/databases.html#database-sync-to-async)) resolved the issue:

```python
from django.db import close_old_connections, reset_queries

def plug_psycopg_leak(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        close_old_connections()
        reset_queries()

        try:
            return func(*args, **kwargs)
        finally:
            close_old_connections()
            reset_queries()

    return wrapper
```

This quickly became an issue on my app because it stores big blobs in DB and they basically were never freed from memory.

> _Note_ &mdash; Those blobs make sense in my app because they're only read/written from Procrastinate so I don't really care how much slower the operation is in comparison to storing that in a S3 bucket, and as a bonus I don't need a S3 bucket

This is how my `memray flamegraph --leaks` diagnostic looked like before the fix:

Image

Now all the cursors get freed and I don't see a spot of psycopg in my flamegraph anymore.

The weird thing is, it looks like this leak only happens on my production server. When running on my local machine, the leak is nowhere to be seen in my flamegraph. Which I guess is consistent with the fact that no one reported this issue?

Before digging any deeper, I'd like to get the point of view from the Procrastinate team on this. Like, what is the intended behavior in regards to clearing the Django connections? How come Channels needs it and Procrastinate doesn't mention it?

Thanks!

### LINKED ISSUE #1551: contrib/django: auto-close Django per-thread DB connections after task execution
### TL;DR

When a Procrastinate sync task uses Django ORM, the per-thread Django DB connection opened inside the asgiref `sync_to_async` worker thread is never closed. In long-running workers this accumulates connections over time; in tests it surfaces as `database "test_postgres" is being accessed by other users` at session teardown.

Every other major Django-integrated job system (Celery, RQ, django-q, Channels) closes Django connections automatically as part of its unit-of-work lifecycle. Procrastinate is the outlier. I'd like to propose this be handled in `procrastinate.contrib.django` so consumers don't have to remember per-task cleanup.

### Background

This was previously discussed in #1106 (closed by me in 2024 with the conclusion "I have the feeling there's nothing Procrastinate can do about it"). After ≥18 months of recurring pain across two of my Django projects — and an independent confirmation from @mschfh in November 2025 — I think that earlier conclusion was premature. The fix is small, scoped to the Django contrib, and would eliminate a real footgun.

### Symptom

A test that exercises a Procrastinate worker (e.g. via the `run_worker_once()` pattern documented at <https://procrastinate.readthedocs.io/en/stable/howto/django/tests.html#integration-tests>):

```python
@pytest.mark.django_db(transaction=True)
def test_something():
    # Setup that defers a job using a sync task containing ORM calls
    ...
    run_worker_once()
    ...
```

ends with:

```
PytestWarning: Error when trying to teardown test databases:
OperationalError('database "test_postgres" is being accessed by other users
DETAIL:  There is 1 other session using the database.')
```

The same leak happens to periodic tasks. Anything wired with `@app.periodic` and using ORM will leak one connection per worker startup. In `adit_radis_shared.common.tasks.retry_stalled_jobs` (runs every 10 min), this leaks on every test session that touches `run_worker_once()`.

### Why it's the Django contrib's responsibility

Django's connection cache is **per-thread**, and the contract is that the caller must close per-thread connections in non-request contexts. Procrastinate sync tasks run inside an asgiref `sync_to_async` worker thread. The application code doesn't own that thread — Procrastinate does.

That means:

- A test-side fixture that calls `connections.close_all()` cannot help, because the leaked connection is in a different thread the test can't reach. (Verified: this is what led to my original "nothing to do" comment.)
- An application-side per-task `db.close_old_connections()` works but is **opt-in, undiscoverable, easy to forget**, and has to be added to every new task in every consumer codebase. New users hit the warning, get confused, find #1106, and end up implementing the workaround themselves.
- A Procrastinate-side wrapper has direct access to the worker-thread context. One implementation point covers all consumers.

This is the standard pattern in other Django-integrated job systems:

| Framework | Closes Django connections after work? | How |
|---|---|---|
| Celery | Yes | `task_postrun` signal triggers `close_old_connections` |
| Channels | Yes | `database_sync_to_async` decorator wraps cleanup around each call |
| django-q2 / django-rq | Yes | Close after each job |
| Procrastinate | **No** | Caller's responsibility |

### Proposed fix

In the Django contrib's task-execution path (in `procrastinate/contrib/django/`), wrap the user's task with a finalizer that calls Django's `close_old_connections()` in the same worker thread context the task ran in.

Sketch (the exact insertion point depends on the current internals — I'm happy to do the implementation work):

```python
# In procrastinate.contrib.django, wherever sync tasks are invoked via
# sync_to_async (or equivalent in the async path):

from asgiref.sync import sync_to_async
from django import db

async def _run_django_task_with_cleanup(task_fn, *args, **kwargs):
    try:
        if asyncio.iscoroutinefunction(task_fn):
            return await task_fn(*args, **kwargs)
        else:
            return await sync_to_async(task_fn)(*args, **kwargs)
    finally:
        # Runs in the worker thread (for sync tasks) or the worker event-loop
        # thread (for async tasks). Closes the per-thread connection cache.
        await sync_to_async(db.close_old_connections)()
```

A cleaner alternative, matching how Celery does it, is firing Django's `request_finished` signal at task end — `close_old_connections` is already wired to that signal in Django:

```python
from django.core.signals import request_finished

# in the same finally branch:
request_finished.send(sender=None)
```

Either approach is roughly 5–10 lines, lives entirely in `contrib/django/`, and has no effect on non-Django Procrastinate usage.

### Testing

The integration test I never managed to land in 2024 is:

```python
@pytest.mark.django_db(transaction=True)
@pytest.mark.filterwarnings("error::pytest.PytestUnraisableExceptionWarning")
def test_django_task_does_not_leak_db_connections(django_db_blocker):
    @app.task
    def task_that_uses_orm():
        list(MyModel.objects.all())

    task_that_uses_orm.defer()
    run_worker_once()

    with django_db_blocker.unblock():
        with connection.cursor() as c:
            c.execute(
                "SELECT count(*) FROM pg_stat_activity "
                "WHERE datname = current_database() AND pid <> pg_backend_pid()"
            )
            assert c.fetchone()[0] == 0
```

This is what the consumer projects (RADIS, ADIT) have had to work around at the test-fixture level. If `contrib/django` handles connection cleanup, this passes out of the box.

### Migration impact

The change is purely additive. Consumers who currently call `db.close_old_connections()` inside their tasks — there are many in our codebases linked back to #1106 — would keep working unchanged; the explicit cleanup just becomes a no-op (no connection is older than the freshly-closed one).

The change is also opt-in conceptually: it only affects users of `procrastinate.contrib.django`. No effect on the core async API or non-Django consumers.

### What I'd like

If you're open to the direction, I'm happy to do the implementation and write the integration test. I want to make sure we agree on the approach (wrapper vs. signal vs. something else) before I spend the time. References:

- #1106 (the original 2024 thread)
- Downstream issue we just filed: openradx/adit-radis-shared#197
- Linked downstream issue: openradx/adit#352

medihack:
## What

When a Procrastinate task uses the Django ORM, the per-thread Django DB connection opened to run it was never closed. In long-running workers this accumulates connections over time; in tests it surfaces at teardown as `database "..." is being accessed by other users`.

The Django contrib now manages Django's per-thread database connections around each task the same way Django does around an HTTP request — `close_old_connections()` **before and after** every task (mirroring Django's `request_started`/`request_finished` wiring, and what Celery and Channels do), plus `reset_queries()` after each task to clear Django's per-connection query log. This covers **both sync and async** tasks:

- **sync tasks** clean up in the asgiref `sync_to_async` worker-pool thread (where the leaked connection lives);
- **async tasks** clean up via `sync_to_async(thread_sensitive=True)`, since Django's `connection.close()` is `@async_unsafe` and must not run on the event loop.

`CONN_MAX_AGE` is respected — healthy persistent connections are kept, only obsolete/unusable ones are dropped. `reset_queries()` is a no-op unless `DEBUG=True`, but closing the connection does not clear the query log (it lives on the persistent connection wrapper), so a long-running worker with `DEBUG=True` would otherwise accumulate it.

## How

- New `procrastinate/contrib/django/db_cleanup.py`:
  - `wrap_task(task)` — idempotent, async/sync-aware wrapper that runs the cleanup around the task. Preserves `inspect.iscoroutinefunction(task.func)` so the worker's sync/async branching keeps working.
  - `DjangoApp(procrastinate.App)` — overrides `_register_task` and `add_tasks_from`, the two registration chokepoints, so every task (decorator, blueprints, `import_paths`/autodiscover, builtin, periodic) is wrapped.
- `apps.py` `create_app` instantiates `DjangoApp` instead of `procrastinate.App`.
- `DjangoApp` is re-exported as `procrastinate.contrib.django.DjangoApp` (public) so a test can build a throwaway app that gets the same cleanup; the Django test howto's broken `ProcrastinateApp(...)` snippet is fixed to use it.
- Docs: a "Database connections" section in `howto/django/basic_usage.md`, including how a very long task can release its connection mid-run.

The cleanup runs only at task boundaries (not mid-task); the docs note how a long-running task can call `close_old_connections()` itself if it idles on the DB for a long stretch.

## Tests

- **Unit** (`tests/unit/contrib/django/test_db_cleanup.py`): cleanup before+after and on error, `reset_queries` runs once after-only, async path, sync/async-ness preserved, idempotency, both-registration-paths coverage.
- **Integration** (`tests/integration/contrib/django/test_db_cleanup.py`): defer a sync/async ORM task, run the worker once, assert no net `pg_stat_activity` session increase. Uses `transaction=True` so the worker's separate connection actually sees the committed job. Verified red-first — both tests fail (leaked sessions, incl. the exact teardown error) without the wrapper.

## Related issues

- **Closes #1551** — the test-teardown `database "..." is being accessed by other users` leak (supersedes the closed #1106, which was entirely about this warning in Django integration tests). The Django test howto's per-test "throwaway app" pattern is also fixed and made leak-free via the public `DjangoApp`.
- **Likely fixes #1316** ("Django memory leak?") — this is the in-contrib version of the Channels-style `close_old_connections()` wrapper that resolved it downstream, plus `reset_queries()` for the `DEBUG=True` query-log growth. Left as "Closes"-free since memory-leak confirmation needs the reporter's environment.
- **Partially addresses #1134** ("OperationalError: the connection is closed") — the before-task close drops a stale connection left by a previous task, so the next task reconnects fresh; a connection that dies *mid-way through a single long task* is out of scope (see the long-task note in the docs).

## Summary by CodeRabbit

* **Documentation**
  * Added a "Database connections" how-to and updated test guidance and API docs for the Django integration.

* **New Features**
  * Django integration now auto-wraps registered tasks to manage DB connections around task execution and exposes a Django-specific app class.

* **Bug Fixes**
  * Ensures DB connections are closed/reset for both sync and async tasks to prevent leaks.

* **Tests**
  * Added unit and integration tests verifying no leaked DB sessions for sync and async tasks.

### DIFF
--- a/docs/howto/django/basic_usage.md
+++ b/docs/howto/django/basic_usage.md
@@ -53,6 +53,32 @@ in Django's settings (see {doc}`logs`).
 See {doc}`../basics/command_line` for more details on the CLI.
 If you prefer writing your own scripts, see {doc}`scripts`.
 
+## Database connections
+
+The worker manages Django's database connections around each task the same way
+Django does around an HTTP request: per-thread connections are closed before and
+after every task (sync or async) via `close_old_connections()`. You don't need to
+manage connections in your tasks, and `CONN_MAX_AGE` is respected, so persistent
+connections are reused between tasks rather than force-closed.
+
+Because the connection is only closed at these task boundaries (not in between), a
+connection opened early in a long-running task stays open until the task finishes,
+even across a long stretch of non-database work. If that matters — to free a
+connection slot, or to avoid reusing a connection that may have dropped while
+idle — close it from within the task once you're done with the database for a
+while:
+
+```python
+from django.db import close_old_connections
+
+@app.task
+def long_task():
+    do_early_db_work()
+    close_old_connections()  # release the connection before the long idle stretch
+    do_hours_of_non_db_work()
+    do_late_db_work()  # reconnects fresh
+```
+
 ## Deferring jobs
 
 Defer jobs from your views works as you would expect:
--- a/docs/howto/django/basic_usage.md
+++ b/docs/howto/django/basic_usage.md
@@ -57,9 +57,11 @@ If you prefer writing your own scripts, see {doc}`scripts`.
 
 The worker manages Django's database connections around each task the same way
 Django does around an HTTP request: per-thread connections are closed before and
-after every task (sync or async) via `close_old_connections()`. You don't need to
-manage connections in your tasks, and `CONN_MAX_AGE` is respected, so persistent
-connections are reused between tasks rather than force-closed.
+after every task (sync or async) via `close_old_connections()`, and Django's
+query log (only populated under `DEBUG=True`) is cleared with `reset_queries()`.
+You don't need to manage connections in your tasks, and `CONN_MAX_AGE` is
+respected, so persistent connections are reused between tasks rather than
+force-closed.
 
 Because the connection is only closed at these task boundaries (not in between), a
 connection opened early in a long-running task stays open until the task finishes,
--- a/docs/howto/django/basic_usage.md
+++ b/docs/howto/django/basic_usage.md
@@ -59,16 +59,16 @@ The worker manages Django's database connections around each task the same way
 Django does around an HTTP request: per-thread connections are closed before and
 after every task (sync or async) via `close_old_connections()`, and Django's
 query log (only populated under `DEBUG=True`) is cleared with `reset_queries()`.
-You don't need to manage connections in your tasks, and `CONN_MAX_AGE` is
-respected, so persistent connections are reused between tasks rather than
-force-closed.
+So for an ordinary task you don't have to do anything — the connection it opens is
+cleaned up at the task boundaries, and `CONN_MAX_AGE` is respected, so persistent
+connections are reused between tasks rather than force-closed.
 
-Because the connection is only closed at these task boundaries (not in between), a
-connection opened early in a long-running task stays open until the task finishes,
-even across a long stretch of non-database work. If that matters — to free a
-connection slot, or to avoid reusing a connection that may have dropped while
-idle — close it from within the task once you're done with the database for a
-while:
+The one case this doesn't cover is *within* a single long-running task: the worker
+only closes at task boundaries, so a connection opened early stays open until the
+task finishes, even across a long stretch of non-database work. That holds a
+connection slot for the whole task and risks the idle connection being dropped
+before a later query. If that matters, close it from within the task once you're
+done with the database for a while:
 
 ```python
 from django.db import close_old_connections
--- a/docs/howto/django/tests.md
+++ b/docs/howto/django/tests.md
@@ -78,13 +78,17 @@ def test_my_task():
 :::{note}
 Registering a new task on the django procrastinate app within a test will
 make this task available even after the test has run. You probably want to
-avoid this, for example by creating a new app within each of those tests:
+avoid this, for example by creating a new app within each of those tests. Use
+{py:class}`~procrastinate.contrib.django.DjangoApp` (the same app class the
+contrib uses) so tasks run by a worker still get the automatic connection
+cleanup (see {doc}`basic_usage`) and don't leak connections at test database
+teardown:
 
 ```python
-from procrastinate.contrib.django import app
+from procrastinate.contrib.django import DjangoApp, app
 
 def test_my_task():
-    new_app = procrastinate_app.ProcrastinateApp(app.connector)
+    new_app = DjangoApp(connector=app.connector)
 
     @new_app.task
     def my_task(a, b):
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -100,6 +100,8 @@ Django
 
 .. automodule:: procrastinate.contrib.django
 
+.. autoclass:: procrastinate.contrib.django.DjangoApp
+
 
 SQLAlchemy
 ----------
--- a/procrastinate/contrib/django/__init__.py
+++ b/procrastinate/contrib/django/__init__.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+from .db_cleanup import DjangoApp
 from .procrastinate_app import app
 from .utils import connector_params
 
 __all__ = [
+    "DjangoApp",
     "app",
     "connector_params",
 ]
--- a/procrastinate/contrib/django/apps.py
+++ b/procrastinate/contrib/django/apps.py
@@ -7,7 +7,7 @@ from django.utils import module_loading
 
 import procrastinate
 
-from . import django_connector, procrastinate_app, settings
+from . import db_cleanup, django_connector, procrastinate_app, settings
 
 
 class ProcrastinateConfig(apps.AppConfig):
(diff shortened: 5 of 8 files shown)


medihack:
Converting this to a **draft** for now.

The connection-cleanup logic here is correct, but it relies on a `DjangoApp` subclass that intercepts task registration, and to make test apps leak-free it has to **expose that class publicly**. A worker middleware (#1317) would be a nicer fit: the cleanup becomes a single execution-time hook the worker applies around every task, so we'd need **no `App` subclass and no new public class** — a plain `procrastinate.App` run by the worker would get the cleanup too.

Note the cleanup body would be essentially the same either way (sync tasks must close in their own `sync_to_async` thread, async via `sync_to_async(thread_sensitive=True)`); the middleware only changes *how it's attached*, and only works if the middleware runs in the task's execution context rather than on the event loop.

**Plan:**
1. Try to land the middleware feature (#1317) first and implement the Django connection cleanup on top of it — preferred, since it avoids exposing an extra class.
2. If the middleware approach doesn't work out, fall back to merging this PR as-is.

Keeping this branch ready as the fallback in the meantime.

medihack:
Closing in favor of #1577, which has been merged. #1577 delivers the same automatic Django DB connection cleanup — `close_old_connections()` before/after each task and `reset_queries()` after, for both sync and async tasks — but builds on the task middleware feature instead of wrapping `task.func`, while keeping the same public `DjangoApp` surface. The exploration here directly shaped that final approach.

PR #1508, opened 2026-02-07
feat: add external connection support for atomic job deferral

psontag:
Allow users to pass their own database connection when deferring jobs via task.configure(connection=conn).defer(), enabling atomic deferral within user-managed transactions.

Adds separate execute_query_all_*_with_connection methods rather than adding an optional connection param to the existing execute_query_all_* methods, to avoid breaking changes to the connector interface.

Supported by SyncPsycopgConnector, PsycopgConnector, and SQLAlchemyPsycopg2Connector. Unsupported connectors raise ConnectorException with a clear message.

Closes #1500

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20breaking%20%F0%9F%92%A5
  - [x] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20feature%20%E2%AD%90%EF%B8%8F
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20bugfix%20%F0%9F%95%B5%EF%B8%8F
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20miscellaneous%20%F0%9F%91%BE
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20dependencies%20%F0%9F%A4%96
  - [ ] https://github.com/procrastinate-org/procrastinate/labels/PR%20type%3A%20documentation%20%F0%9F%93%9A

## Summary by CodeRabbit

* **New Features**
  * Defer jobs using an externally managed, transaction-scoped database connection (sync & async); job persistence now follows the caller’s commit/rollback.

* **Documentation**
  * Added a how-to guide with usage patterns and examples for using external connections with supported connectors.

* **Tests**
  * Expanded integration and unit tests validating external-connection deferral behavior and connector handling.

PR #1560, opened 2026-06-10
feat: add external connection support for job cancellation and status

### LINKED ISSUE #1550: Extend external connection support to cancel and status queries
Hi, following up on https://github.com/procrastinate-org/procrastinate/issues/1500 / https://github.com/procrastinate-org/procrastinate/pull/1508. The new `connection=` on defer has been really useful for us, but we hit a related case it doesn't cover.

Our setup is a fairly typical one: a webserver that talks to the queue, and a separate pool of workers that actually run the jobs. The webserver is the only thing that enqueues jobs, cancels them, and reads their status. So `defer`, `cancel_job_by_id`, and `get_job_status` all get called from the webserver, which already maintains its own connection pool for the rest of its database work.

With https://github.com/procrastinate-org/procrastinate/pull/1508 in place, defer happily uses our pool:

```python
task.configure(connection=conn).defer_async(...)
```

But `cancel_job_by_id` and `get_job_status` don't accept a connection, so they go through procrastinate's own pool. In our process that means we have to either:

1. Open a separate connection (outside our pool) just for the cancel/status call, which is wasteful and partly defeats the point of having a pool, or
2. Skip the public API and call the underlying SQL functions directly through our own connection, which couples us to procrastinate's internal SQL surface and breaks if it ever changes.

Neither feels great. What we'd like is the same shape https://github.com/procrastinate-org/procrastinate/pull/1508 added for defer:

```python
await app.job_manager.cancel_job_by_id_async(job_id, connection=conn)
await app.job_manager.get_job_status_async(job_id, connection=conn)
```

Additionally, we update other data on cancel as well, and it would be great for that to be transactional too.

Looks like the implementation mirrors #1508 closely: a new `execute_query_one_*_with_connection` on the connector interface, alongside the existing `execute_query_all_*_with_connection`, and an optional `connection=` on the two `JobManager` methods. Non-breaking; when `connection` isn't passed the existing pool path is unchanged.

I'd be happy to open a PR if you think this is worth pursuing.

atteahma:
Follow-up to #1508. `cancel_job_by_id` / `cancel_job_by_id_async` and `get_job_status` / `get_job_status_async` now take the same optional `connection=` argument as defer, so they can run on a user-managed connection inside the caller's own transaction. If you update your own tables when cancelling a job, both changes commit (or roll back) together.

The implementation follows the same pattern: new `execute_query_one_*_with_connection` methods on the connector interface next to the existing `execute_query_all_*` ones, implemented for `SyncPsycopgConnector`, `PsycopgConnector` and `SQLAlchemyPsycopg2Connector`. Other connectors keep raising `ConnectorException`. When `connection` isn't passed, nothing changes.

One note on the approach: this adds two more `*_with_connection` methods, and I know from the review on that PR (https://github.com/procrastinate-org/procrastinate/pull/1508#discussion_r3020389290) that multiplying these isn't the preferred long-term direction. I kept the existing pattern for consistency, but happy to rework it to fold `connection=` into the regular `execute_query_*` methods if you'd rather go that way now.

Closes #1550

### Successful PR Checklist:

- [x] Tests
  - [ ] (not applicable?)
- [x] Documentation
  - [ ] (not applicable?)

#### PR label(s): 

## Summary by CodeRabbit

* **New Features**
  * Job cancellation and job-status retrieval now support externally-managed database connections, enabling transactional workflows within caller-controlled commits and rollbacks.

* **Documentation**
  * Production guide expanded with a dedicated section and examples for cancelling jobs and reading status using the same external connection, including updated timing notes.

* **Tests**
  * Added unit and integration coverage for connection-scoped cancellation/status across supported connectors, including commit/rollback behavior.

### DIFF
--- a/docs/howto/production/external_connection.md
+++ b/docs/howto/production/external_connection.md
@@ -80,6 +80,8 @@ your own tables when cancelling a job and want both changes to commit (or roll b
 together:
 
 ```python
+job_id = 33
+
 async with await psycopg.AsyncConnection.connect("...") as conn:
     await conn.execute("UPDATE orders SET state = 'cancelled' WHERE id = %s", [42])
     await app.job_manager.cancel_job_by_id_async(job_id, connection=conn)
--- a/docs/howto/production/external_connection.md
+++ b/docs/howto/production/external_connection.md
@@ -2,7 +2,8 @@
 
 Procrastinate supports deferring jobs on an externally-managed database connection.
 This lets you insert the job within the same transaction as your own database
-operations, enabling patterns like the **transactional outbox**.
+operations, enabling patterns like the **transactional outbox**. Cancelling a job
+and reading its status can run on an external connection too.
 
 ## When to use this
 
@@ -71,6 +72,20 @@ with connector.engine.connect() as conn:
     conn.commit()
 ```
 
+## Cancelling jobs and reading status
+
+`cancel_job_by_id` / `cancel_job_by_id_async` and `get_job_status` /
+`get_job_status_async` accept the same `connection` argument. Useful if you update
+your own tables when cancelling a job and want both changes to commit (or roll back)
+together:
+
+```python
+async with await psycopg.AsyncConnection.connect("...") as conn:
+    await conn.execute("UPDATE orders SET state = 'cancelled' WHERE id = %s", [42])
+    await app.job_manager.cancel_job_by_id_async(job_id, connection=conn)
+    await conn.commit()  # the order update and the cancellation commit together
+```
+
 ## Supported connectors
 
 - `SyncPsycopgConnector` — pass a `psycopg.Connection`
@@ -82,7 +97,9 @@ Other connectors will raise a `ConnectorException` if `connection` is provided.
 ## Important notes
 
 - **NOTIFY is delayed**: The database NOTIFY that wakes workers fires when you
-  commit. This is expected and acceptable.
+  commit. This is expected and acceptable. The same applies when cancelling a
+  running job with `abort=True`: the worker only sees the abort request once you
+  commit.
 - **Error handling**: Errors from the job INSERT (e.g. queueing lock violations)
   are wrapped as usual via procrastinate's exception hierarchy, but they may abort
   your transaction. Use savepoints if you need to isolate the defer from the rest
--- a/procrastinate/manager.py
+++ b/procrastinate/manager.py
@@ -348,7 +348,11 @@ class JobManager:
         )
 
     def cancel_job_by_id(
-        self, job_id: int, abort: bool = False, delete_job: bool = False
+        self,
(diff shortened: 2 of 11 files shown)


docs/howto/production/external_connection.md
```diff
@@ -71,6 +72,20 @@ with connector.engine.connect() as conn:
...
+`cancel_job_by_id` / `cancel_job_by_id_async` and `get_job_status` /
+`get_job_status_async` accept the same `connection` argument. Useful if you update
+your own tables when cancelling a job and want both changes to commit (or roll back)
+together:
+
+```python
+async with await psycopg.AsyncConnection.connect("...") as conn:
+    await conn.execute("UPDATE orders SET state = 'cancelled' WHERE id = %s", [42])
+    await app.job_manager.cancel_job_by_id_async(job_id, connection=conn)
+    await conn.commit()  # the order update and the cancellation commit together
```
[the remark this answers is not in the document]
atteahma: Fixed in 3779304d. Defined job_id above the connection block and used the variable. Didn't take the suggested diff as-is though: 42 in the UPDATE is the order's id, not the job's, so reusing one value for both would be misleading.

procrastinate/manager.py
```diff
@@ -410,53 +433,82 @@ async def cancel_job_by_id_async(
...
         """
-        result = self.connector.get_sync_connector().execute_query_one(
-            query=sql.queries["get_job_status"], job_id=job_id
-        )
+        if connection is not None:
+            result = (
+                self.connector.get_sync_connector().execute_query_one_with_connection(
+                    connection, query=sql.queries["get_job_status"], job_id=job_id
+                )
+            )
```
ewjoachim: I think this code is trying to say it would like it better if it was splito more lines with intermediate variables :) That's how I read the parens cascade :)

ewjoachim (approved):
> I kept the existing pattern for consistency, but happy to rework it to fold connection= into the regular execute_query_* methods if you'd rather go that way now.

I'm in favor of doing it... later. Care to open an issue so we don't forget ?

(Let's kept PRs small and focused)

procrastinate/manager.py
atteahma: Agreed -- Updated in 0662085

atteahma:
> > I kept the existing pattern for consistency, but happy to rework it to fold connection= into the regular execute_query_* methods if you'd rather go that way now.
> 
> I'm in favor of doing it... later. Care to open an issue so we don't forget ?
> 
> (Let's kept PRs small and focused)

Created [here](https://github.com/procrastinate-org/procrastinate/issues/1586)!

ewjoachim (approved):

ewjoachim:
> Created https://github.com/procrastinate-org/procrastinate/issues/1586!

Thank you :)

Issue #1308, opened 2025-01-27
Add hooks / callbacks to worker


medihack:
As discussed in the middleware feature request (#1292), I want to propose an alternative solution that uses hooks/callbacks at specific locations in the worker that a user can plug into. The implementation is easier, and the usage is, in my opinion, more flexible, as we can add additional hooks if needed.

I would start with the following ones (https://github.com/procrastinate-org/procrastinate/tree/worker-hooks):
- `before_fetching_job_hook` (should work fine for rate limiting)
- `job_processing_started_hook` (has access to the JobContext)
- `job_processing_ended_hook` (has access to the JobContext and JobResult)

I would also implement that if any of those hooks (or the job itself) raise a `StopWorker` exception, then `worker.stop()` will be called.

Do you have any opinions about using hooks instead of middleware?

onlyann:
Job started and job complete (whether successful or not) calbacks would be useful.

I am not sure about before fetch. If this is intended to be used as a way to determine if a job should be fetched, I don't know how that would quite work.

ewjoachim:
I think providing a wrapper function for tasks would be a way to provide a hook before and after, and if we propagate the exception, it can also discriminate between success and failure, all with a usual API. That said, if we provide individual hooks, they can be easily implemented as wrappers underneath (and probably vice versa)

About the before-fetch: it could be useful, but this all depends on how it may affect the fetch query. If it doesn't affect the query, then I don't see what it may do except a very basic rate limit (that works only if the worker processes only the rate limited task).

Do we have other ideas for usecases of pre-fetch (or any other hook: pre start, pre stop, abort, anything ...) ?
Should we allow task-related hooks to be bound to specific tasks or should hooks be always worker-wide ?

medihack:
> I think providing a wrapper function for tasks would be a way to provide a hook before and after, and if we propagate the exception, it can also discriminate between success and failure, all with a usual API. That said, if we provide individual hooks, they can be easily implemented as wrappers underneath (and probably vice versa)

Yes, and we can let the user plug more into the whole lifecycle of the worker then with middleware. Of course, we could have different kinds of middleware wrapping different stuff, but I find this cumbersome.

> About the before-fetch: it could be useful, but this all depends on how it may affect the fetch query. If it doesn't affect the query, then I don't see what it may do except a very basic rate limit (that works only if the worker processes only the rate limited task).

[...]

PR #1579, opened 2026-06-18
Add worker middleware

### LINKED ISSUE #1292: Middleware feature
In #1262 and #1289 we mentioned that a middleware feature would be helpful in some scenarios. In the [`middleware`](https://github.com/procrastinate-org/procrastinate/tree/middleware) branch, I have already started to implement it, but I would like to discuss the implementation (@ewjoachim). I use a similar implementation to Django for its middleware. I would like to discuss two API alternatives.

1) The way I implemented it currently:
```python
ProcessTask: TypeAlias = Callable[[tasks.Task, job_context.JobContext], Awaitable[Any]]
Middleware: TypeAlias = Callable[
    [ProcessTask],
    Callable[[tasks.Task, job_context.JobContext], Awaitable[Any]],
]

def simple_middleware(process_task: ProcessTask):
    async def middleware(task: tasks.Task, context: job_context.JobContext):
        # Do something before the task is processed
        return await process_task(task, context)
        # Do something after the task is processed
    return middleware
```

2) An alternative (maybe better) way:
```python
ProcessJob: TypeAlias = Callable[[job_context.JobContext], Awaitable[Any]]
Middleware: TypeAlias = Callable[
    [ProcessJob],
    Callable[[job_context.JobContext], Awaitable[Any]],
]

def simple_middleware(process_job: ProcessTask):
    async def middleware(context: job_context.JobContext):
        # Do something before the job is processed
        return await process_job(context)
        # Do something after the job is processed
    return middleware
```

After some thought, 2) is better because `process_job` also wraps the part where the job status is saved back to the db. This is especially important if we allow the user to raise a `StopWorker` exception (https://github.com/procrastinate-org/procrastinate/issues/1262#issuecomment-2568467649).
One thing that bothers me a bit is that in both scenarios, the `context` was already created, and thereby a `start_timestamp` is already set. So when the user adds some pause in the middleware (e.g. for #1289) before the job is processed, this pause will be included in the overall job duration. I'm not sure if this is what we want or not.

medihack:
Implements the event-loop `worker_middleware` chain reserved as a follow-up in #1556. Builds on the task middleware from #1556 (which closed #1292).

## What

A **worker middleware**: an always-async callable `(call_next, context, worker)` that wraps every job's run on the event loop — both sync and async tasks — registered worker-wide only.

```python
async def otel_mw(call_next, context, worker):
    with tracer.start_as_current_span(f"run/{context.task.name}"):
        return await call_next()

app.run_worker(worker_middleware=[otel_mw])
```

Unlike task middleware, a single worker middleware wraps both sync and async tasks, so there's no sync/async kind-matching. The motivating case is OpenTelemetry-style tracing around every job without monkey-patching (#1027).

## Design

- **Boundary A — wraps the task run, not the bookkeeping.** Composed as the outermost layer around the task-middleware chain at the `ensure_async` call site in `Worker._process_job`, staying inside the existing `try/except/finally`. The task's exception propagates through `call_next`; status classification and the DB status write stay outside the middleware. (The alternative — wrapping `_process_job` incl. the status write — was rejected: it would overlap the future `job_processing_ended` hook and force a non-Pythonic outcome API, since `_process_job` swallows exceptions internally.)
- **Always async, worker-wide only.** Non-async / non-callable entries raise `TypeError` at worker construction. No per-task parameter, no `Task`/`Blueprint` changes.
- **`worker.stop()`**, not a `StopWorker` exception (already thread-safe/cooperative since #1556).
- **Nesting:** worker middleware → worker-wide task middleware → per-task task middleware → task.
- **No-op when unset**; reuses the generic `compose` / `is_async_middleware` from #1556 unchanged.

## Tests

`tests/unit/test_middleware.py`: wraps async task, wraps **sync** task (the headline), nests outside task middleware, result transform, exception→failed, `JobAborted`→aborted, retry-through-middleware, `stop()` from inside, `worker_defaults` injection, rejects sync/non-callable, no-op. Full unit suite 538 passing; ruff + basedpyright clean; docs `-W` clean.

## Docs

Restructured `docs/howto/advanced/middleware.md` to present task and worker middleware as parallel sections with a comparison; `WorkerMiddleware` added to the API reference.

## Follow-up (out of scope)

The worker **hooks** from #1308 (`before_fetching_job` for rate limiting; `job_processing_ended(JobResult)` for resolved-status observation) — complementary and deliberately separate. Keeping the middleware at boundary A keeps the hooks orthogonal.

## Release Notes

* **New Features**
  * Added worker-level middleware support to wrap jobs on the event loop, enabling middleware to run in the event loop context for both sync and async tasks.

* **Documentation**
  * Reorganized and clarified middleware concepts, distinguishing between task middleware and worker middleware with improved explanations of execution contexts and ordering.
  * Updated API reference to include the new worker middleware type.

### DIFF
--- a/docs/howto/advanced/middleware.md
+++ b/docs/howto/advanced/middleware.md
@@ -99,8 +99,9 @@ async def otel_mw(call_next, context, worker):
 app.run_worker(worker_middleware=[otel_mw])
 ```
 
-It can also be set on `worker_defaults`. A non-async or non-callable entry raises
-`TypeError` when the worker is created.
+It can also be set on `worker_defaults`. When the worker is created, a non-callable
+entry raises `TypeError`, and a sync (non-async) entry raises
+{py:class}`~procrastinate.exceptions.MiddlewareKindMismatch`.
 
 A worker middleware is the **outermost** layer. For a given job the full chain is,
 outermost to innermost: worker middleware → worker-wide task middleware → per-task
--- a/docs/howto/advanced/middleware.md
+++ b/docs/howto/advanced/middleware.md
@@ -1,16 +1,30 @@
-# Add task middleware
+# Add middleware
 
-A *task middleware* wraps the execution of a task, letting you run code before and
-after it — for logging, dependency-injection scopes, resource cleanup, and so on.
+*Middleware* wraps the execution of your tasks, letting you run code before and
+after each one — for logging, tracing, dependency-injection scopes, resource
+cleanup, and so on.
 
-A middleware is a callable taking `(call_next, context, worker)`. It must call (or
-`await`) `call_next()` to run the next middleware — or the task itself — and return
-the result. `context` is the {py:class}`~procrastinate.JobContext`; `worker` is the
-running worker (you may call `worker.stop()` from it).
+Procrastinate has two kinds of middleware:
 
-## The sync/async rule
+- **Task middleware** runs *in the task's own execution context* — a sync
+  middleware in the task's worker thread, an async middleware on the event loop —
+  so it must match the task's sync/async nature. Register it per-task or
+  worker-wide.
+- **Worker middleware** is *always async* and wraps every job on the event loop,
+  both sync and async tasks. Register it worker-wide only.
 
-A middleware's nature must match the task it wraps:
+Both are callables taking `(call_next, context, worker)`. A middleware must call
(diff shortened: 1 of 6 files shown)


ewjoachim (approved):
Awesome, thank you !

PR #1595, opened 2026-07-16
Add persistent pause/resume queue

art321ur:
## Why
 When we are using procrastinate, we found that sometimes a queue needs to stop temporarily (to allow other non-procrastinate services do some maintenance or allow them to use limited resources that procrastinate queue would consume). 
Today the only options are to stop the worker (which also stops every other queue) or to let the jobs run and fail. This PR adds a way to pause one queue and leave the  rest of the queues working.

I have tried to work around it (in simplified setup) in our codebase but the only way to solve this was to have some hacky `fetch_job` interception. 
```
class PauseAwareConnector(PsycopgConnector):  
    async def execute_query_one_async(self, query: LiteralString, **arguments: Any) -> dict[str, Any]:
        if query == sql.queries["fetch_job"] and arguments.get("queues") is not None:
            arguments = {**arguments, "queues": await self._unpaused(arguments["queues"])}
        return await super().execute_query_one_async(query, **arguments)

    async def _unpaused(self, queues: Sequence[str]) -> list[str]:
        rows = await super().execute_query_all_async(_SELECT_PAUSED_QUEUES)
        paused = {row["queue_name"] for row in rows}
        return [queue for queue in queues if queue not in paused]

```

## How
- the pause lives in the database, so it is shared by all workers and survives
  a worker restart
- `procrastinate_fetch_job_v3` is a copy of `_v2` with one extra predicate that skips jobs whose queue has a
  row
- jobs already running are not interrupted
- each pause is held under a *pause key* (`"default"` if not given). A queue stays paused until every key is released, so two independent processes (for example a deploy script and a maintenance task) can pause the same queue without one resuming the other by mistake
 - cleanup of dead pauses is assumed to be DIY - but the `paused_at` column could be used for that
 
### Migrations

  - `03.10.00_01_pre_add_queue_pause.sql`
  - `03.10.00_50_post_add_queue_pause.sql`

### Successful PR Checklist:

- [x] Tests
    - Unit, integration and acceptance tests
- [x] Documentation
    - New how-to: `docs/howto/advanced/pause_queue.md`.

#### PR label(s): 

## Summary by CodeRabbit

* **New Features**
  * Queues can now be paused and resumed through synchronous and asynchronous APIs.
  * Paused queues stop receiving new jobs while running jobs finish normally.
  * Supports multiple pause keys, resuming individual keys or clearing all keys.
  * Added APIs to list paused queues with optional filters.
  * Workers automatically resume fetching jobs when a queue is resumed.

* **Documentation**
  * Added comprehensive queue pause and resume guidance.
  * Updated Django model documentation for paused queue records.

### DIFF
--- a/docs/howto/advanced.md
+++ b/docs/howto/advanced.md
@@ -9,6 +9,7 @@ advanced/locks
 advanced/schedule
 advanced/priorities
 advanced/cancellation
+advanced/pause_queue
 advanced/queueing_locks
 advanced/cron
 advanced/retry
--- a/docs/howto/advanced/pause_queue.md
+++ b/docs/howto/advanced/pause_queue.md
@@ -1,0 +1,91 @@
+# Pause a queue
+
+We can pause a queue so that workers stop fetching its jobs, then resume it
+later. This is useful to temporarily hold a queue's work, for example during a
+maintenance window or while an external dependency is unavailable, without
+stopping the worker process.
+
+Pausing only stops the *fetching* of new jobs: jobs already being processed run
+to completion, and pending jobs stay in the `todo` state until the queue is
+resumed. Workers keep fetching jobs from the other queues.
+
+The pause is stored in the database, so it is shared by every worker consuming
+the queue and it survives a worker restart.
+
+## Pause a queue
+
+```python
+# by using the sync method
+app.job_manager.pause_queue("some_queue")
+# or by using the async method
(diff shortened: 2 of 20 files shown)


art321ur:
Let me know if I should change or improve anything - PR description, docs etc- first time trying to contribute to OSS ;)

